Issue #1390, opened 2014-07-03
vsvim adds 0x8D after a autocompleted curly brackets when used together with visual assist


michalbrzozowski:
I noticed that my struct failed compilation due to the following errors:
C2144: syntax error : 'int' should be preceded by ';'
C4430: missing type specifier - int assumed. Note: C++ does not support default-int

Looking at the code i couldt see anything wrong with the code;

struct Foo
{
  char var;
};

Then i wrote the same struct in notepad and copied it in and it compiled perfectly. 

So i saved the file and opened it in a hex editor and noticed a small difference.

The struct that didn't compile had a extra byte(0x8D) added between the opening bracket and the carriage return + line feed. 
E.g:
0x7B, 0x8D, 0x0D, 0x0A
~~~~~~~^ error

While the correct line not using (auto complete bracket) gave the following:
0x7B, 0x0D, 0x0A

This char does not show up in the editor, so its impossible to find it, without opening the file in a hex editor.

Disabling the vmvim plugin makes autocomplete of brackets work properly again.

Im using visual studio 2012, haven't tried to repro on any other versions.

jaredpar:
This should be fixed with the associated PR.

Issue #1008, opened 2012-12-21
Problem between vsVim and VisualAssist auto close brace


darkpotpot:
When using vsVim Plugin, there is a problem with VisualAssist autoclose brace feature.

If in visual assist I activate the option advanced/corrections/Insert () and closing } ) ] and put the parameters to "closing brace is inserted after a blank line", there is something weird which is inserted between the brace.

For example, typing :

namespace Foo
{

put a } right after the { (instead of after a blank line) and, on top of that, something weird is inserted between { and }, producing an incorrect syntax.

jaredpar:
Should be fixed by #1509

PR #1509, opened 2014-11-27
Fix invalid characters with visual assist

fredizzimo:
When using a keyboard layout that requires alt-gr to type "{", VsVim would
convert the newline character to an invalid 0x8D character. This is fixed
by not applying the alt and shift modifiers, if both modifers are down at
the same time.

This fixes bug #1390 and #1008, which are both really the same. The bugs really forced you to turn off many features of visual assist, so it was quite bad. English keyboard layouts were unaffected.

I'm not really sure if the solution is the best, but I guess you could get a better idea of how to fix it from this.


fredizzimo:
Hi Jared,

Sorry for the delay, I was planning to figure out exactly what went wrong and why the code was needed, so I could write accurate comments. However right after that I got way too busy with work, and  after that with Christmas holidays. 

However during that time, I found out that the fix is not enough, and there are still some problems remaining.

It seems like it's somehow eating some linefeed characters, leaving just a carriage return, which is causing line ending differences.

But even worse than that, some times the tab characters are missing, causing the curly brackets to not align properly. 

Those other two problems are also caused by the alt-gr, using an English keyboard layout, that doesn't happen. I will continue investigating this, and push some more changes, when I have found a complete fix.

fredizzimo:
Hi Jared,

I have now added the requested comments, and fixed the remaining problems that I described earlier.

Best Regards,
Fred Sundvik

jaredpar:
@fredizzimo awesome, thank you!

PR #1537, opened 2015-01-08
788 whitespace only not part of paragraph

### LINKED ISSUE #788: "ap" (a paragraph) counts whitespace-only lines as part of the paragraph
Doing a `dap` or `cap` or `yap` etc does not work as expected in that it treats whitespace-only lines as part of the paragraph. Vim does not treat these as part of the paragraph. This is pretty vital when tabs etc are not visible -- there's no way in that case to tell whether dap will delete the entire buffer or just what looks like a single paragraph.

Kantis:
Fixes #788


jaredpar:
Looks great!  Thanks!

Issue #1151, opened 2013-08-07
Support `[ and `] to jump to beginning/end of previously changed or yanked text


angularsen:
``[` and ``]` doesn't seem to do anything in VsVim. It is also not shown when typing `:marks`.
However, the related ``<` and ``>` marks seem to work.

Need this for [Selecting your pasted text](http://vim.wikia.com/wiki/Selecting_your_pasted_text)
"[...] select the text you just pasted. For example, you may want to paste some text, then select it for indenting, or formatting, or performing a substitution."

See special marks section:
http://vim.wikia.com/wiki/Using_marks

squerl:
For anyone else finding this issue and waiting eagerly, I'd like to recommend marks that cover the important use cases of missing **\`] `[**:

**`.** jumps to position last changed in buffer. If you paste a block of text, I used to use **`]** to jump to the end of it, but **`.** does the exact same thing.

**`<** jumps to beginning of last visual selection. When I would highlight a selection and yank and want to jump to the beginning of that, I would use **`[**, but **`<** does the same thing too!

Hope this helps.

anjdreas:
@squerl Thanks for the tip on \`. In my usecase, I just want to format the text I just pasted for indentation etc. So I can now solve this with ```v`.``` or create an alias for it.

Example:
Yank some lines of text
Go somewhere and paste it
Cursor is now at beginning of pasted text
Enter select mode ```v```
Go to end of pasted text ``` `.```
Format text ```=```

anjdreas:
I wound up with these two aliases:

```
" Paste and then format the pasted text
nnoremap fp pv`.= 
" Paste over selection, and then format the pasted text
vnoremap fp "_dPv`.= 
```

ricksladkey:
I have implemented this but I can't submit it until PR #2139 is merged. Here is a screenshot after yanking a visual selection and pasting it:

![image](https://user-images.githubusercontent.com/2680524/41272951-644c1228-6de5-11e8-936f-57013a69f244.png)

Stamo-Gochev:
Nice! When the update will be available for the VS extension?

PR #1538, opened 2015-01-09
Add support for additional special marks [ and ]

Kantis:
As mentioned in #1151 

Creating pull request to indicate WIP. Feedback appreciated!


Src/VimCore/VimTextBuffer.fs
```diff
@@ -23,6 +23,8 @@ type internal VimTextBuffer
     let _switchedModeEvent = StandardEvent<SwitchModeKindEventArgs>()
     let mutable _modeKind = ModeKind.Normal
     let mutable _lastVisualSelection : ITrackingVisualSelection option = None
+    let mutable _lastModifiedSpan : ITrackingVisualSelection option = None
```
jaredpar: This doesn't appear to be used (may just be missing it in the change).  If not can we delete?

Src/VimCore/VimTextBuffer.fs
```diff
@@ -23,6 +23,8 @@ type internal VimTextBuffer
     let _switchedModeEvent = StandardEvent<SwitchModeKindEventArgs>()
     let mutable _modeKind = ModeKind.Normal
     let mutable _lastVisualSelection : ITrackingVisualSelection option = None
+    let mutable _lastModifiedSpan : ITrackingVisualSelection option = None
+    let mutable _lastChangedOrYankedSpan : SnapshotSpan option = None
```
jaredpar: The type of this value should be `ITrackingSpan`.  

**Long explanation for those unfamiliar with the editor model here:**

The `SnapshotSpan` type represents a span of text with in a specific snapshot (represented as `ITextSnapshot`) of the buffer.  A snapshot is immutable and can be safely held onto for long periods of time.  Every edit to the buffer (represented as `ITextBuffer`) produces a new snapshot value.  

Operations such as caret move, edits, etc ...  are all done in terms of the current snapshot of the buffer.  If you attempt to say move the caret using a `SnapshotPoint` from a previous snapshot it will throw an exception.  

Storing the last modified span as a `SnapshotSpan` value means it will become invalid on the very next edit.  Any attempt to use <kbd>`[</kbd> at that point will result in an exception.

Types like `ITrackingSpan` are meant for storing spans which need to adjust as the buffer is edited.  The `TrackingSpanUtil` type makes transitions between `SnapshotSpan` and `ITrackingSpan` pretty easy.

``` fsharp
let trackingSpan = TrackingSpanUtil.Create snapshotSpan SpanTrackingMode.EdgeInclusive
```

Src/VimCore/VimTextBuffer.fs
```diff
@@ -45,6 +47,15 @@ type internal VimTextBuffer
                 | None -> None
                 | Some visualSelection -> Some (_bufferTrackingService.CreateVisualSelection visualSelection)
 
+    member x.LastChangedOrYankedSpan
+        with get() =
+            match _lastChangedOrYankedSpan with
+            | None -> None
+            | Some snapshotSpan -> _lastChangedOrYankedSpan
+        and set value =
+            // Necessary? Unfamiliar with F# praxis. Feels like it should be the default.
```
jaredpar: Yep this is necessary.

Src/VimCore/VimTextBuffer.fs
```diff
@@ -45,6 +47,15 @@ type internal VimTextBuffer
                 | None -> None
                 | Some visualSelection -> Some (_bufferTrackingService.CreateVisualSelection visualSelection)
 
+    member x.LastChangedOrYankedSpan
+        with get() =
+            match _lastChangedOrYankedSpan with
```
jaredpar: You should be able to simplify this method to just

```
with get() = _lastChangedOrYankedSpan
```

jaredpar:
Overall this looks great.  I added some feedback about the storage for the yanked / modified span.  It just needs to be adjusted a small bit to account for user edits.  After that this change should be good to go.

Kantis:
Thanks a lot for the feedback @jaredpar :) However, the marks need to be updated on deletes, shifts, replaces, case changes, etc.. Quite a few commands trigger updates to these marks. I've been considering how to best handle all these events without bloating every single operation in CommandUtil. My current plan is to implement a SpecialMarksUpdater component similar to ChangeTracker in MefComponents. Do you think that's a good idea? Or should I simply update the marks from within CommandUtil?

jaredpar:
@Kantis 

Implementing this feature in a separate component designed like `ChangeTracker` would be hard if not impossible.  It would need to be able to distinguish between edits made by VsVim and edits made by other components (which do happen in many cases).  That's not really possible to do.

I think this has to be done manually :(   

One way I thought of to avoid the code bloat though is to add specific APIs for the edit into `CommandUtil`.  Essentially we add APIs along the line of 

``` fsharp
member x.TextBufferDelete (span : SnapshotSpan) = 
  _textBuffer.Delete(span.Span) |> ignore
  _vimBuffer.LastChangedOrYankedSpan <- span  
``` 

Would also need another one for `Insert` as well and possibly `Replace`.

Edits which should update the `[` and `]` register go through this edit API instead of manually editting via `ITextBuffer`.  That would centralize the updates and keep the code bloat down.  

I don't like the name though.  I was pondering `DeleteAndTrack` but hated that more.

Kantis:
Maybe we could ignore _textBuffer entirely (within CommandUtil) and instead add methods to modify the TextBuffer through the VimTextBuffer? 

E.g.. In CommandUtil:
```
_vimTextBuffer.Delete(span.Span) 
```

In VimTextBuffer
```
member x.Delete(span : SnapshotSpan) =
    _textBuffer.Delete(span.Span) |> ignore
    x.LastChangedOrYankedSpan <- span
```

Src/VimCore/CommandUtil.fs
```diff
@@ -3040,6 +3046,9 @@ type internal CommandUtil
         _textView.Selection.Select(_textBuffer.CurrentSnapshot.GetExtent(), false)
         CommandResult.Completed ModeSwitch.NoSwitch
 
+    member x.UpdateLastChangedOrYankedSpan snapshotSpan =
+        _vimTextBuffer.LastChangedOrYankedSpan <- Some(snapshotSpan)
+
```
jaredpar: I believe this is unused

jaredpar:
Adding it to `IVimTextBuffer` would definitely work as well.  But it would create a bit of an inconsistency with who was responsible for updating the span:

- Yank operations: would explicitly set the `LastYankOrChangeSpan`
- Change operations: would implicitly set the `LastYankOrChangeSpan` by calling into `IVimTextBuffer::Delete`

Usually I opt for trying to keep it the same for all operations if possible.

Kantis:
>Adding it to IVimTextBuffer would definitely work as well. But it would create a bit of an inconsistency with who was responsible for updating the span:

Sadly it got a bit spread out. Modes_Insert_InsertMode sets the marks in OnLeave and OnEnter now. The rest of the mark updates are done explicitly within CommandUtil. Do you want me to refactor it somehow?

jaredpar:
@Kantis why did you move away from tracking it as a span?  That seems to be the more natural way of doing it.  Updating the start without the end or vice versa seems like it's potentially ignoring data.

Nit: please avoid double blank lines.

Kantis:
I had a case where I wanted to update the endpoint separately (InsertMode::OnLeave). 

Will do. I feel a bit handicapped without ReSharper.. :)

Issue #1541, opened 2015-01-11
vw selects too much


Kantis:
Using the sample text:
```
the brown
dog
```

Put cursor at start of 2nd word of first line and use `vwy`. the d character from the following line will be yanked as well. In gVim 7.4 the linebreak is included, but the d character is not.

Kantis:
Seems to be an issue even when not at line end

jaredpar:
I believe what you are seeing here is an effect of the `selection` setting.  

- `inclusive`: this is the default and when set the `'d'` character will be selected.
- `exclusive`: when this is set the `'d'` character will not be selected.  

VsVim mimics the gVim behavior for this scenario.

Kantis:
Thanks @jaredpar. This "issue" came up while I was working on more unit tests for #1538 . I was doing this to attempt setting special marks:
```
Create("the brown dog");
_textView.MoveCaretTo(1);
_vimBuffer.Process("v2wy");
```

I came across this now though, which seems much more appropriate.

```
var visualSelection = VisualSelection.NewCharacter(new CharacterSpan(span), Path.Forward);
_vimTextBuffer.LastVisualSelection = FSharpOption.Create(visualSelection);
_vimBuffer.Process("y");
```

PR #1546, opened 2015-01-12
Visual character selection should have select caret

Kantis:
Fixes issue #1541 and replicates gVim behaviour

### DIFF
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaretController.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaretController.cs
@@ -102,13 +102,15 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
                         }
                     }
                     break;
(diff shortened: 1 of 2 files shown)


Kantis:
Not sure if this should be achieved by setting IsSelectionInclusive somehow. I couldn't find any precedents and the global setting seems to always be Inclusive which gives us the block caret.

jaredpar:
I don't believe this change is correct.  The Visual Character mode should use block selection.  Only when `selection=exclusive` should we be using the select caret which the current code does.

Kantis:
Doh. I didn't know about the selection=inclusive / exclusive setting. Makes sense now, thanks.

jaredpar:
@Kantis writing VsVim has caused me to learn about so many settings and commands I never knew existed :)

Kantis:
Great way to learn :)

Issue #664, opened 2011-10-12
Completion with intellisense not repeatable.


rodimius:
Press 'i' type conso, then press tab to complete to Console, press escape.
. only repeats the conso part of the insert.
This works similarly for multiple completions in one insert.
I end up with lines like:
conso<tab>.R<tab>();

Version 1.1.0.0

fredizzimo:
At least for the Visual Assist case, the undo grouping in the insert mode(https://github.com/jaredpar/VsVim/wiki/Vim-undo-grouping-in-Insert-mode) seems to be the biggest culprit.

Most cases, got fixed when I deleted the  _sessionData <- { _sessionData with CombinedEditCommand = None } line, from the OnCaretPositionChanged function in the Modes_Insert_InsertMode.fs file.

Visual Assist always updates the mouse cursor, when using intellisense, and resetting the session data clears all changes that have been made before that, which leaves only the last change, which in turn usually includes deletion of the first letters in the autocompleted word. Also if you have autocompletion of brackets turned on, the cursor will be moved inside the brackets, and you will get an empty repeat.

Unfortunately this modification breaks strict Vim compability when navigating with the mouse in insert mode, so I haven't made any pull request with that change. No unit tests were failing though, so some tests should probably be added for that case.

I think the best way to fix this would be to add some non-standard option, that will disable insert mode "undo-grouping" totally and threat all changes in the insert mode as one atomic insert. IMO, the grouping doesn't really make sense otherwise either. For example you see an error in some text that you typed, and want to correct it, so you navigate there with the arrow keys, but that means that you cannot repeat it anymore. I can't really think of any case where I would like to have the normal Vim behaviour, but that doesn't mean that there aren't any. You can always exit the insert mode, if you want a new group.

I will start experimenting with this now, and see if I can prepare a changelist that let's you disable the grouping by some option, to let you repeat exacty what was done in insert mode. Do you think that's a good idea? Also what should that option be called?

jaredpar:
@fredizzimo this is an area where I think a non-standard option has to come into play.  Extensions like R# and Visual Assist do operations that just aren't comparable to what happens in gVim.

ricksladkey:
This appears to work correctly now, and without setting `vsatomicinsert`, both in 2.5.0.0 and on master.

ricksladkey:
Actually, I see that several issues are still being reported for this issue when either Visual Assist or ReSharper is present.

ricksladkey:
[...]

PR #1552, opened 2015-01-25
Add vsvimatomicinsert setting to fix Intellisense repeations

fredizzimo:
Hi Jared,

I have added a non-standard vim setting vsvimatomicinsert (ains), which when enabled fixes most of the issues with repeating commands that uses Visual Assist code completion. The primary issue that it fixes is #664, but probably not completely.

When the vsatomicinsert option is enabled, then all movements in insert mode is tracked. All absolute movements (by the mouse or Visual Assist for example), are converted into relative ones. So you can repeat the action in a different location.

I also added a couple of unit test for the normal insert mode, since no test broke, when I changed the code, when it clearly should have broken something.

I have not tested this very extensively yet, since I just finished it. But I'm creating this pull request so that you can look at it and comment. Unfortunately, I'm not going to write much c++ at work during the following weeks, so I won't be able to test it properly during that time either.

However most basic tests seems to work, even multi-line code snippet repetitions, that opens a popup to ask for some information. They can't be repeated into a block with a different indentation though, and I don't have any idea how to fix that, I also think I saw some existing bug about that. 

Another thing that breaks is, if you have the "Insert _ after m and shift" option enabled. It works by highlighting the the _ character, while shift is pressed, so that it will be overwritten if you press _. When you then release shift, it stops the highlight and moves the cursor forward. This could probably be detected and fixed, but I think much better, is to just disable that option. At least for me, I don't see how it's very useful, 

Do you have some comments about the naming of the option? And should we have separate options for tracking direct cursor movements(with mouse or automatic), and movements done by the arrow keys?

I have no previous F# experience, and almost no C# experience either, so please tell me if you see something that could be done in a better way.

Edit:
I forgot to mention, that this doesn't fix macros containing code completion. That has to be fixed in a different way.

Best Regards,
Fred Sundvik


academician:
Hey, Fred and Jared - if this is an actual fix to some of the Visual Assist integration problems I've been having, I'd love to see it merged. Any plans to look at it?

fredizzimo:
Hi academician,

I have been running with this enabled since I made the merge request, and everything seems to be working great for me. I haven't actually encountered any problems at all for my use cases, the dot repetition seems to work, even for multiline completions.

However, it has to be noted that the behavior is different from normal Vim, so therefore I added a new option ("vsvimatomicinsert"). Without that option everything should behave as it did before.

It also doesn't fix all the autocompletion problems with recorded macros. But those should be possible to fix in a similar way.

From my point of view, this change can be merged, if Jared agrees on the naming of the option, and the actual implementation. It would of course be great if someone else could test it first as well.

Src/VimCore/VimSettings.fs
```diff
@@ -149,6 +149,7 @@ type internal GlobalSettings() =
 
     static let GlobalSettingInfoList = 
         [|
+            (AtomicInsertName, "ains", SettingValue.Toggle false)
```
jaredpar: Can we use the pattern `(AtomicInsertName, AtomicInsertName, ...)` here?  In general for non-standard options, like `CaretOpacityName`, I don't provide abbreviations.  The reasoning is to try and keep non-standard options as visible as possible.

jaredpar:
@fredizzimo sorry.  I completely lost track of this PR after you initially submitted it.  I just looked it over and think it's mostly good to go.  Provided a minor comment about some naming issues but otherwise the change looks great!

fredizzimo:
Hi Jared,

Sorry for the delay, I had some problems with the computer at home. But I have now updated the pull request with the short name of the option removed.

jaredpar:
Thanks for the contribution!

PR #1559, opened 2015-02-10
Better build output

madskristensen:
Using this general purpose VSIX build steps for AppVeyor, you get these benefits:

1. The artifacts are pushed with a clean name (`VsVim.vsix` instead of `Src\VsVim\bin\Debug\VsVim.vsix`)
2. The artifacts get a label: `Latest build`
3. The version number is incremented in the .vsixmanifest before build using the AppVeyor {build} property
  - It will then be major.minor.{build}
4. The AppVeyor build version is updated to the same version as your VSIX version
5. You no longer have to make point-point releases like 1.7.1. It will just be 1.7.{build} similar to how VS does it
6. The produced .vsix is automatically uploaded to vsixgallery.com
  - Here you can give your users access to the nightly build
  - You can deep link to the nightly build. AppVeyor doesn't allow deep links to artifacts
     - Example: http://vsixgallery.com/extensions/VsVim.Microsoft.e214908b-0458-4ae2-a583-4310f29687c3/VsVim.vsix

You can see what the AppVeyor build looks like here https://ci.appveyor.com/project/madskristensen/vsvim/build/1.7.5 and the entry in vsixgallery.com here http://vsixgallery.com/extension/VsVim.Microsoft.e214908b-0458-4ae2-a583-4310f29687c3/

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,13 @@
 
 configuration: Debug
 
+install: 
+  - ps: (new-object Net.WebClient).DownloadString("https://raw.github.com/madskristensen/ExtensionScripts/master/AppVeyor/vsix.ps1") | iex
+
+before_build: 
+  # Updates the version number in the .vsixmanifest and updates the AppVeyor build number to match
+  - ps: Vsix-IncrementVsixVersion | Vsix-UpdateBuildVersion
+
 build:
   project: VimAll.sln
   verbosity: minimal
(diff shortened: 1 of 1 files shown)


jaredpar:
This seems really nice.  I've long wanted an easy way to have a nightly dogfood system (for those brave enough to deal with the churn).  

Is there any way to get the increment level past the point release.  Essentially having the nightly build increment from 1.7.1 to 1.7.1.1, then 1.7.1.2?

madskristensen:
yes, you can increment the `build` or the `revision` part of the version. It defaults to `build`. To set it to revision, do this:

`Vsix-IncrementVsixVersion -versionType revision`

That would change the version to this:

1.7.1 => 1.7.1.{build}
1.7    => 1.7.0.{build}

Issue #1574, opened 2015-03-06
Tag parsing does not recognize non alphanumeric characters in tag/attribute names


leafgarland:
The tag parser fails to handle tag names with `:` and `.`, this is a pain when working with namespaces and Xaml.

I had a quick look at making a fix, [see here](https://github.com/leafgarland/VsVim/commit/803dc0a289c256337be622913f6121d3d172a409), I can put that in a PR if you want. It just adds some extra chars to what is considered a valid name.

Looking at vim's behaviour, I think it parses any characters until whitespace as the tag name and then anything until the closing `>` is skipped. That might be a better fix.

jaredpar:
I'd love it if you sent that along as a PR.

jaredpar:
Closing as PR addressed this.

PR #1575, opened 2015-03-07
Tag parser handles some other characters in tag and attributes names

leafgarland:
The tag parser recognises letters, digits and a few other valid characters:
':', '.', and '_' in tag and attribute names.

Fixes issue #1574


jaredpar:
Thanks for contributing this!

PR #1589, opened 2015-04-10
Support ReSharper 9.1

vadimi:
ReSharper 9.1 has updated version of `JetBrains.Platform.VisualStudio.SinceVs10` assembly. New file version is `102`.

### DIFF
--- a/Src/VsVimShared/Implementation/ReSharper/ResharperVersionutility.cs
+++ b/Src/VsVimShared/Implementation/ReSharper/ResharperVersionutility.cs
@@ -56,7 +56,8 @@ namespace Vim.VisualStudio.Implementation.ReSharper
 
                 // Starting with ReSharper 9, the assembly we detect is part of the "ReSharper Platform"
                 // Which for ReSharper 9 is of version 6.0
-                if (version.Major == 6)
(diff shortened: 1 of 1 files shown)


oysteinkrog:
Tested it, confirmed working, thanks!
@jaredpar Would be nice to get this merged as R# 9.1 has been released now and people will be upgrading.

Src/VsVimShared/Implementation/ReSharper/ResharperVersionutility.cs
```diff
@@ -56,7 +56,8 @@ internal static ReSharperVersion DetectFromAssembly(Assembly assembly)
 
                 // Starting with ReSharper 9, the assembly we detect is part of the "ReSharper Platform"
                 // Which for ReSharper 9 is of version 6.0
-                if (version.Major == 6)
+                // and Resharper 9.1 has version 102
+                if (version.Major == 6 || version.Major == 102)
```
jaredpar: Seems like such an odd change on the R# part.

vadimi: I totally agree. I was also surprised when saw this version number.

PR #1676, opened 2015-10-12
Fixed tag block parsing attributes without values

brnls:
Fixed parsing tags when containing attributes without values ("\<button disabled\>Search\</button\>"). Please let me know if anything needs to be changed.

An aside: The test projects use XUnit for unit tests, but the XUnit Runner NuGet package isn't part of the test project so I could not get tests to run without installing it. I installed it and then removed it when I was done testing. Is there some other way to get them to run?


jaredpar:
Thanks for this!

Issue #1679, opened 2015-10-15
Navigating between vertical split windows fails when both windows are viewing the same file.


mattrrichard:
Reproduce Steps:
1. Open a single file
2. <kbd>CTRL</kbd>-<kbd>w</kbd> <kbd>v</kbd> to split file
3. <kbd>CTRL</kbd>-<kbd>w</kbd> <kbd>l</kbd> to move to the second column
4. Attempt <kbd>CTRL</kbd>-<kbd>w</kbd> <kbd>h</kbd> to move back. 

If you instead open two different files in the columns, it works as expected.

mattrrichard:
Using VsVim version 2.0.370 in VS 2013

mattrrichard:
Looks like it's due to [the way `MoveFocusHorizontally` finds the currently active window](https://github.com/jaredpar/VsVim/blob/f711e222b249559aa9b5d00747cdf3f4820f0b66/Src/VsVimShared/VsVimHost.cs#L580):

```
var indexOfActiveDoc = topLevelWindows.FindIndex(win => win.Document == _dte.ActiveDocument);
```

When there are two windows displaying the same document, the index returned from that might be wrong.  I'll try to look into it more deeply later today if I get time.  This is something I haven't done before so it'll be an adventure!

idealist1508:
it's awesome.  I thought it is impossible, because of #1602

jaredpar:
> When there are two windows displaying the same document, the index returned from that might be wrong.

That's likely the case.

PR #1680, opened 2015-10-16
Fix 1679 (Ctrl-w h/l navigation when multiple windows have the same document)

mattrrichard:
Fixes issue #1679 

This turned out to be surprisingly easy to deal with; the window object can be equated against directly to solve two windows with the same document.

While I was in there, I noticed that left and right were swapped because the `indexDelta` was subtracted rather than addded to the current index.  I hope you don't mind that I fixed this here rather than creating a separate issue and pull request just for that.


jaredpar:
Thanks for this!

Issue #1671, opened 2015-09-18
:m0 moves line to second line instead of first


bschwehn:
in vim 

:m0 will move the current line to the top. In vsvim it moves it to the second line, just below the top line. 

Is this a bug or do I have some setting wrong/another add in interfering? 

Thanks, 
Ben

jaredpar:
That just looks like a bug in the `:m` implementation.  I will get that looked at.

mattrrichard:
I dug into this, and I see what happening: When the interpreter executes a MoveTo or CopyTo command, it uses its `GetLine` method to convert to destination line specifier into an `ITextSnapshotLine`.  While in that method, it runs into one of the mismatches between Vim and VS's functionalities, how lines are numbered.  The `Util.VimLineToTssLine` method is used to convert and you lose the information that the user wanted to insert before the first line.

```
/// Vim is fairly odd in that it considers the top line of the file to be both line numbers
    /// 1 and 0.  The next line is 2.  The editor is a zero based index though so we need
    /// to take that into account
    let VimLineToTssLine line = 
        match line with
        | 0 -> 0
        | _ -> line - 1
```

The problem really boils down to losing information while calling GetLine: **a line specifier for line 1 and a line specifier for line 0 have the same result **.  

The only idea I have for how to fix this is to refactor `GetLineCore` into two methods where the first returns the target *Vim* line number and have CopyTo and MoveTo decide whether to insert at the start of the destination line or the end by looking at the target line number.

ricksladkey:
My understanding was that in vim, AFTER line 0 and BEFORE line 1 refer to the same place.  So even though line numbers are one-based, operations that conceptually append can still insert before the first line due to the fictitious line zero.

mattrrichard:
I don't think VisualStudio has a concept of the "fictitious line zero" which is the origin of the problem.

jhamm:
I verified it works by building from the source. Can this issue be closed? This will help get us closer to 2.0.2 :smile:

PR #1681, opened 2015-10-17
Fix 1671 (copy and move commands inserting below first line)

mattrrichard:
I added a new method to the interpreter which I called `GetLineAndVimLineNumber` which returns a tuple of the `ITextSnapshotLine` and the corresponding vim line number specified by the `LineSpecifier` given to the method.  `GetLineCore` was changed to simply call `GetLineAndVimLineNumber` to avoid duplicating the line specifier interpreter logic.

I refactored the `RunCopyTo` and `RunMoveTo` methods to get the destination position injected to them rather than the destination line so that the decision to insert before or after the destination line can happen in a common place.  I modified RunCopyToAndMoveTo to use `GetLineAndVimLineNumber` and decide whether to insert the text before the destination line or after by inspecting the line number (and pass that back to the lambda sent by RunCopyTo or RunMoveTo).

**Overall, my changes feel maybe too invasive to me so I would really appreciate feedback on this!**

As a side note, I noticed that the [line specifier case for the current line in `GetLineCore`](https://github.com/jaredpar/VsVim/blob/master/Src/VimCore/Interpreter_Interpreter.fs#L302-L303) calls out to get the current line even though the method presumably gets the current line as a parameter while the [case for an adjustment from the current line](https://github.com/jaredpar/VsVim/blob/master/Src/VimCore/Interpreter_Interpreter.fs#L335-L336) uses the parameter.
```fsharp`
| LineSpecifier.CurrentLine -> 
    x.CaretLine |> Some

//...

| LineSpecifier.AdjustmentOnCurrent adjustment -> 
    getAdjustment adjustment currentLine
```

In my GetLineAndVimLineNumber method I duplicated this behavior just in case it was intentional, but it looked odd to me so I wanted to bring it to your attention.

Double side note: I hate the name `GetLineAndVimLineNumber`, but I couldn't think of anything better.

### DIFF
--- a/Src/VimCore/Interpreter_Interpreter.fs
+++ b/Src/VimCore/Interpreter_Interpreter.fs
@@ -286,67 +286,27 @@ type VimInterpreter
         None
 
     member x.GetVimLineNumber lineSpecifier currentLine =
+        x.GetLineAndVimLineNumber lineSpecifier currentLine
+        |> Option.map (fun (vimLine, line) -> vimLine)
 
+    member x.GetLineAndVimLineNumber lineSpecifier currentLine = 
         // Get the ITextSnapshotLine specified by lineSpecifier and then apply the
         // given adjustment to the number.  Can fail if the line number adjustment
         // is invalid
         let getAdjustment adjustment (line : ITextSnapshotLine) = 
-            adjustment + line.LineNumber + 1
-
-        match lineSpecifier with 
-        | LineSpecifier.CurrentLine -> 
-            x.CaretLine.LineNumber + 1 |> Some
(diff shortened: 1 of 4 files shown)


mattrrichard:
tag #1671 since I somehow left that out of the above.

mattrrichard:
I discovered some cases that my code does not cover that vim allows.  Vim seems to allow negative line numbers as long as they are adjusted back to a valid address before being acted on.  For example: `:m 0-2 2` moves the current line above the first line.  In VsVim (both before and after my changes) the same command results in an "Invalid address" error.

jaredpar:
Looks great!

Issue #1678, opened 2015-10-15
Implement 'normal' command


ecsousa:
One feature that a really miss in VsVim is the normal command. I find it very useful to record macros like this:

```
jVi{:g/public/normal O[DataMember]
```

With this marco, if I am in the first line of a POCO class, it would put a DataMember attribute to each property of that class.

normal command is a great combo with g command

jaredpar:
That's interesting.  I've never seen this before.

jhamm:
This is a useful feature. I verified it works by building from the source. Should this issue be closed?

jaredpar:
@jhamm yep.  Closing.

PR #1691, opened 2015-11-03
Normal command implementation

ecsousa:
Implementing feature request #1678.

I'm not sure if the way I converted a string into a KeyInput list would be the right one (I used RegisterValue class for doing so).


ecsousa:
I my second commit, I have changed the way to convert string into KeyInput list. Actually, new way it is not even converting the first char array into string, it is going directly yo KeyInput list, which should provide better performance for the solution

ecsousa:
I've just realized that my implementation does not handle line rage specifiers.

I'm closing this PR for now, and I will be working on getting it done. I will reopen this PR afterwards.

PR #1695, opened 2015-11-06
Normal command implementation

ecsousa:
Implementing feature request #1678.

### DIFF
--- a/Src/VimCore/Interpreter_Interpreter.fs
+++ b/Src/VimCore/Interpreter_Interpreter.fs
@@ -4,6 +4,7 @@ open EditorUtils
 open Microsoft.VisualStudio.Text
 open Vim
 open Vim.StringBuilderExtensions
+open Vim.VimCoreExtensions
 open System.Collections.Generic
 open System.ComponentModel.Composition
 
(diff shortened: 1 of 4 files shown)


Src/VimCore/Interpreter_Parser.fs
```diff
@@ -1650,6 +1651,17 @@ type Parser
         x.ParseGlobalCore lineRange (not hasBang)
 
     /// Parse out the :help command
```
jaredpar: Can you move this comment back to the x.ParseHelp command

Src/VimCore/Interpreter_Parser.fs
```diff
@@ -1650,6 +1651,17 @@ type Parser
         x.ParseGlobalCore lineRange (not hasBang)
 
     /// Parse out the :help command
+    member x.ParseNormal lineRange =
+        let inputs = seq {
+            while not _tokenizer.IsAtEndOfLine && _tokenizer.CurrentChar = ' ' do
+                _tokenizer.MoveNextChar()
```
jaredpar: Believe this loop could be simplified to `x.SkipBlanks`

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -807,6 +808,97 @@ type VimInterpreter
             if lineRange.Count > 1 then
                 _foldManager.CreateFold lineRange)
 
+    member x.RunNormal (lineRange: LineRangeSpecifier) input =
+        let transactionMap = System.Collections.Generic.Dictionary<ITextBuffer, ILinkedUndoTransaction>();
+        let modeSwitchMap = System.Collections.Generic.Dictionary<ITextBuffer, IVimBuffer>();
```
jaredpar: These maps should use an `ITextView` as the key instead of `ITextBuffer`.  It is possible, although typically unlikely in VIsualStudio, that two `IVimBuffer` instances will map to the same `ITextBuffer`.  The underlying WPF model allows for that to happen.  The `ITextView` to `IVimBuffer` mapping though is unique.

jaredpar:
This looks great!  I left a small bit of feedback, mostly nit picky.  Once that's addressed though this should be good to go.

Src/VimCore/Interpreter_Interpreter.fs
ecsousa: So, why not use `IVimBuffer`? From your description, it seems to me like a better choice over `ITextBuffer` and even `ITextView`.

One more thing to consider: this map strategy I have copied from RunMacro method (on CommandUtil). It seems that we might need to change that implementation as well.

jaredpar: Yes, `IVimBuffer` would also work.  

> One more thing to consider: this map strategy I have copied from RunMacro method (on CommandUtil). It seems that we might need to change that implementation as well.

Yikes.  Yep, I will update that.

jaredpar: I take back this bit of feedback.  I went back and looked at the RunMacro implementation and it reminded me why I built the maps this way.  My assertion about mapping between `ITextBuffer` and `IVimBuffer` is correct.  But in this case it's being tracked to `ILinkedUndoTransaction`.  That is definitely tied 1:1 with `ITextBuffer`, even if that value is shared with multiple `IVimBuffer`.  So `transactionMap` is correct here.  

I think `modeSwitchMap` could be just a `HashSet<IVimBuffer>` here.  But doesn't seem like a big deal.

ecsousa:
I have just pushed the fixes.

Thank you!

Issue #1708, opened 2015-11-25
Normal command issue


vadimi:
I started to use `normal` command (which is very useful by the way) and noticed a case that doesn't work in VsVim, but works perfectly fine in vim. Let's say I want to remove all code from specific symbol (for example comma) till the end of the line in selected range of lines:

`:'<,'>g/,/normal nD`

Nothing happens in VsVim, but as I mentioned vim handles it correctly.

Adding @ecsousa just in case.

ecsousa:
I have taken a quick, this bug is not in normal command. It is in global command which doesn't copy search pattern into "/" register. This way, n stroke will fail as it has no search to continue.

If you, however, first search for ",", and then execute the command, it will work.

I will try to take a look in this issue, to see if I can provide any help.

PR #1709, opened 2015-11-26
Make global command to fill last search pattern

ecsousa:
This will fix issue #1708.


jaredpar:
Thanks for the quick fix!

PR #1719, opened 2015-12-21
Fix normal command used with double quote character

ecsousa:
I have just found a bug on normal command implementation: it was not working if expression contained double quote character. For instance:

    :normal df"


jaredpar:
Good catch!

Having the `"` character be both a meaningful token and comment indicator based on position is one of the most challenging aspects of vimscript :frowning:

PR #1742, opened 2016-02-10
fix #1653 Replace mode with paste register

### LINKED ISSUE #1653: Ctrl-R keystroke for pasting register content replaces only one character in replace mode
As I checked in vim, Control-R and then register key, while being in replace mode should replace the same number of characters after cursor as in register, but VsVim replaces only one and inserts the rest.

Also looks like it should draw double quote under cursor as in insert mode, but that's minor thing.

jhamm:
add overwrite insertCommand and Replace(string) to support this

Please let me know if there are any changes that I can make or if there is a better way of accomplishing this.


jaredpar:
Looks great.  Thanks!

PR #1743, opened 2016-02-19
fix #1714 <C-r><C-w> current word in status line

### LINKED ISSUE #1714: <C-r><C-w> does not put current word in status line
First, thanks for bringing vim to VS. Here it's the thing: `<C-r><C-w>` does not put the current word in status line. I tried it using VS 2013 and 2015 Pro, with no vsvimrc file.
- Steps:
	- Put cursor in any word
	- Type `:s/`
	- Hit `<C-r><C-w>`
	- When `<C-r>` is typed, a double quote (") is shown in the status line. But the word is not
	paste in status line.

jhamm:
HistoryUtil -
move ProcessCore function into Process
move the _inPasteWait in Process to the | None condition
add ProcessPasteSpecial to handle <C-r><C-w> and <C-r><C-a>
add IVimBuffer argument to CreateHistorySession so the current word can be calculated

Vim behavior - when in Command mode with text ":cat" then <C-w> the text is removed to end up with ":"

### DIFF
--- a/Src/VimCore/HistoryUtil.fs
+++ b/Src/VimCore/HistoryUtil.fs
@@ -19,13 +19,15 @@ type HistoryCommand =
     | Cancel
     | Back
     | Paste
+    | PasteSpecial of WordKind
     | Clear
 
 type internal HistorySession<'TData, 'TResult>
     (
         _historyClient : IHistoryClient<'TData, 'TResult>,
         _initialClientData : 'TData,
-        _command : string
+        _command : string,
+        _buffer : IVimBuffer option
     ) =
 
     let _registerMap = _historyClient.RegisterMap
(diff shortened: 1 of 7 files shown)


Src/VimCore/HistoryUtil.fs
```diff
@@ -49,22 +51,18 @@ type internal HistorySession<'TData, 'TResult>
...
-        else
-            x.ProcessCore keyInput
-
-    member x.ProcessCore keyInput =
         match Map.tryFind keyInput HistoryUtil.KeyInputMap with
         | Some HistoryCommand.Execute ->
             // Enter key completes the action
             let result = _historyClient.Completed _clientData _command
             _historyClient.HistoryList.Add _command
+            _inPasteWait <- false
```
jaredpar: Oops :frowning:

jaredpar:
Looks great!

jhamm:
Thanks :smile:

PR #1745, opened 2016-02-21
Vim Disable, allow all fall back actions including key chords

### LINKED ISSUE #1654: Disabed VsVim still interferes with typing capital S, C or P
I'm not sure that this was discussed before but I think it's slightly disturbing that disabling VsVim with control-shift-f12 results in loss of ability to type capital letters S,C or P because it handles Control,Shift,Alt+SCP by itself, so I have to remap them to Visual Studio in that case.

### LINKED ISSUE #1713: VsVim conflicts with the new C# Interactive Window
Somehow VsVim prevents you adding a space in the C# interactive window. Disabling VsVim resolves the issue.

References
https://www.visualstudio.com/en-us/news/vs2015-update1-vs.aspx#Csharp

jhamm:
Fix #1654
Fix #1713

VsVimShared/FallbackKeyProcessor

Handle Shift+C Shift+P as capital C or P:
* GetKeyBinding does not use the AggregateKeyInput which trims off the
* ModifierKeys, it uses the new struct of Char * VimKeyModifiers.
* TryConvertLettersToKeyInput uses the KeyInputUtil methods CharToKeyInput
* and ChangeKeyModifiersDangerous. ApplyKeyModifiersToChar is removing the
* VimKeyModifiers.Shift which causes issues for the TryProcess.

Allow fallback Key Chords when VsVim is disabled:
* Visual Studio allows Key Chord length up to 2, now the fallback can
* support this.
* FallbackCommand has Scope for sorting purposes
* FallbackCommandList is changed to ILookup<char, FallbackCommand>
* All RemovedBindings are added to the FallbackCommandList
* TryParse supports Key Chords of 1 or 2 length
* TryParse uses keyInput.Char and keyInput.KeyModifiers as match criteria.
* When using the standard == operator, this also compares the VimKey which
* is not necessary.

Manual Tests:
type all upper and lower case letters
Ctrl+A (select all); Ctrl+U (lower case selection); Ctrl+Shift+U; (upper
case selection); Ctrl+Z (undo)
Ctrl+M, Ctrl+A (collapse all outlining); Ctrl+M, Ctrl+O (collapse to
definitions); Ctrl+M, Ctrl+X (expand all outlining)
Ctrl+A (select all); Ctrl+K, Ctrl+C (comment selection); Ctrl+K, Ctrl+U
(uncomment selection)
Ctrl+R, Ctrl+W (show whitespace toggle ON); Ctrl+R, Ctrl+W (show
whitespace toggle OFF)
(from C# Interactive window with VsVim enabled) type a space

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
+++ b/Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
@@ -17,20 +17,46 @@ namespace Vim.VisualStudio.Implementation.Misc
     internal sealed class FallbackKeyProcessor : KeyProcessor
     {
         /// <summary>
+        /// A smaller container that represents the parts of a KeyStroke that
+        /// the FallbackKeyProcessor uses.
+        /// </summary>
+        internal struct KeyCharModifier
+        {
+            private char _char;
+            private VimKeyModifiers _keyModifiers;
+
+            internal KeyCharModifier(char c, VimKeyModifiers keyModifiers)
+            {
+                _char = c;
+                _keyModifiers = keyModifiers;
+            }
+            internal char Char { get { return _char; } }
+            internal VimKeyModifiers KeyModifiers { get { return _keyModifiers; } }
+
+            internal static KeyCharModifier Create(KeyStroke stroke)
+            {
+                return new KeyCharModifier(stroke.Char, stroke.KeyModifiers);
+            }
+        }
+
+        /// <summary>
         /// A fallback command is a tuple of a previously bound key paired
         /// with its corresponding Visual Studio command
         /// </summary>
         internal struct FallbackCommand
         {
-            private KeyInput _keyInput;
+            private ScopeKind _scopeKind;
+            private List<KeyCharModifier> _keyBinding;
             private CommandId _command;
 
-            internal FallbackCommand(KeyInput keyInput, CommandId command)
+            internal FallbackCommand(ScopeKind scopeKind, KeyBinding keyBinding, CommandId command)
             {
-                _keyInput = keyInput;
+                _scopeKind = scopeKind;
+                _keyBinding = keyBinding.KeyStrokes.Select(KeyCharModifier.Create).ToList();
                 _command = command;
             }
-            internal KeyInput KeyInput { get { return _keyInput; } }
+            internal ScopeKind ScopeKind { get { return _scopeKind; } }
+            internal List<KeyCharModifier> KeyBindings { get { return _keyBinding; } }
             internal CommandId Command { get { return _command; } }
         }
 
@@ -41,7 +67,8 @@ namespace Vim.VisualStudio.Implementation.Misc
         private readonly IVimBuffer _vimBuffer;
         private readonly ScopeData _scopeData;
 
-        private List<FallbackCommand> _fallbackCommandList;
+        private KeyInput _firstChord;
+        private ILookup<char, FallbackCommand> _fallbackCommandList;
 
         /// <summary>
         /// In general a key processor applies to a specific IWpfTextView but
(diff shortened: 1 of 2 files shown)


Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
```diff
@@ -17,20 +17,46 @@ namespace Vim.VisualStudio.Implementation.Misc
     internal sealed class FallbackKeyProcessor : KeyProcessor
     {
         /// <summary>
+        /// A smaller container that represents the parts of a KeyStroke that
+        /// the FallbackKeyProcessor uses.
+        /// </summary>
+        internal struct KeyCharModifier
+        {
+            private char _char;
+            private VimKeyModifiers _keyModifiers;
```
jaredpar: Please make fields `readonly`.

Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
```diff
@@ -17,20 +17,46 @@ namespace Vim.VisualStudio.Implementation.Misc
...
+        /// <summary>
         /// A fallback command is a tuple of a previously bound key paired
         /// with its corresponding Visual Studio command
         /// </summary>
         internal struct FallbackCommand
         {
-            private KeyInput _keyInput;
+            private ScopeKind _scopeKind;
+            private List<KeyCharModifier> _keyBinding;
             private CommandId _command;
```
jaredpar: Apparently I don't follow my own advice :frowning:

Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
```diff
@@ -158,7 +177,7 @@ private bool TryConvertLetterToKeyInput(KeyEventArgs keyEventArgs, out KeyInput
             }
 
             var c = (char)('a' + (key - Key.A));
-            keyInput = KeyInputUtil.ApplyKeyModifiersToChar(c, keyModifiers);
+            keyInput = KeyInputUtil.ChangeKeyModifiersDangerous(KeyInputUtil.CharToKeyInput(c), keyModifiers);
```
jaredpar: Why did you need to make this change?  In general `ChangeKeyModifiersDangerous` should never be called because it produces key combinations that don't map to non-English languages.

Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
```diff
@@ -200,19 +221,53 @@ internal bool TryProcess(KeyInput keyInput)
...
+                _firstChord = null;
                 return false;
             }
 
             // Check for any applicable fallback bindings, in order
             VimTrace.TraceInfo("FallbackKeyProcessor::TryProcess {0}", keyInput);
-            foreach (var fallbackCommand in _fallbackCommandList)
+            var findFirstChar = _firstChord != null ? _firstChord.Char : keyInput.Char;
+            var findFirstModifiers = _firstChord != null ? _firstChord.KeyModifiers : keyInput.KeyModifiers;
+            var cmds = _fallbackCommandList[findFirstChar]
```
jaredpar: Won't this cause an exception when `findFirstChar` isn't in the look up structure?

jhamm: The _fallbackCommandList is an ILookup so if nothing is found an empty IEnumerable<FallbackCommand> is returned.

Src/VsVimShared/Implementation/Misc/FallbackKeyProcessor.cs
jhamm: When ApplyKeyModifiersToChar, it alters the char and keyModifier so they are not picked up by the _fallbackCommandList. Here are some examples from the tests above:
````
Ctrl+a -> 1 '\u0001' with no KeyModifiers
Ctrl+m -> 13 '\r' with no KeyModifiers
Ctrl+r -> 18 '\u0012' with no KeyModifiers
````

When the ChangeKeyModifiersDangerous is used, the char and modifier are correctly passed along. Each of the tested keyboard shortcuts work.

Do you have any suggestions on how to get the best of both approaches?

jaredpar: @jhamm it's unfortunately not easy to test unless you're very familiar with non-English keyboard layouts.  In the past I've had to rely on customers to walk me through the issues.  

VS makes this more difficult because it's notion of key bindings is equally odd and not always non-English friendly.  The `ChangeKeyModifiersDangerous` though is more along the lines of how VS thinks so it may be more appropriate here.  

I think we may just have to try this and see if it blocks anyone.

PR #1746, opened 2016-02-21
Text-object delete/change fix edgecase

### LINKED ISSUE #1660: Problem with delete/change inside command (not always)
Hi,
I'm a big fan of the "delete/change inside" command but sometimes it doesn't work. For example, with the following line:
`   boost::wregex l_OutlineFileNameRegex(DuplicateBackslash(L"^OutlineFileName:(.*\\.*\\\\).*$"));`
If the cursor is on DuplicateBackslash word, typing ***ci)*** does nothing. In Vim, it deletes everyting up to the last parenthesis.
It seems like VsVim is confused with the ***\\)*** and doesn't count this one as a closing parenthesis.

jhamm:
fix #1660

### DIFF
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -1074,7 +1074,13 @@ type internal MotionUtil
             | None -> false
             | Some point -> SnapshotPointUtil.GetChar point = '\\'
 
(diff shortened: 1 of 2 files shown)


jaredpar:
Nice catch!

PR #1747, opened 2016-02-24
fix #1628 Mouse Visual Selection causes flicker

### LINKED ISSUE #1628: Visual selection with mouse causes flickering and over-selection
I'm currently using Visual Studio 2013 (12.0.31101.00 Update 4), and sometimes (quite often in fact) when editing larger files, if I select text with the mouse (which causes then displays Visual mode in the Vim status bar), the selection jumps as I move downwards and selects text underneath the mouse cursor. 

This only happens when dragging _down_, not up, and only happens during movement; if I stop moving the mouse the selection remains - but that means you might have a selection that includes text below the mouse cursor (for a downward drag selection). 

I searched the issues for something similar, but couldn't find anything listed. Apologies if this issue already exists!

jhamm:
no tests created because SelectionChangeTracker is not under test.
manual test notes added to #1628.
Reviewed other uses of `EditorUtil.GetTextViewLines`, they either have the correct `(y +_textView.ViewportTop)` or do not call the `GetTextViewLineContainingYCoordinate` method.

### DIFF
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -1501,6 +1501,9 @@ module TextViewUtil =
     let GetCaretPointAndLine textView = (GetCaretPoint textView),(GetCaretLine textView)
 
     /// Get the set of ITextViewLines for the ITextView
+    ///
+    /// Be aware when using GetTextViewLineContainingYCoordinate, may need to add the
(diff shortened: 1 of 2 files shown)


jaredpar:
Sorry I did actually review this when you submitted it.  I apparently forgot to hit the merge button :frowning:

jhamm:
Thanks ;)

PR #1749, opened 2016-02-25
fix #1650 BlockCaret after EOL

### LINKED ISSUE #1650: Cursor goes past end of line when clicking past the end of line in normal mode
In gVim, when clicking somwhere to right of the text on a line, the cursor is positioned on the last character of the line. In VsVim the cursor is positioned to the right of the last character. Placing the cursor past the end of the text on a line normally isn't possible in Vim and in VsVim it can only be placed there by using the mouse.

~~~
Some text.<EOL>
         ||               ^- click here.
         |^- the cursor is placed here.
         ^- it should go here.
~~~

jhamm:
This is one way of accomplishing this issue. This uses the BlockCaret
and move the left position over by 1 width when the cursor is at the end
of the line unless ve=onemore is set.

### DIFF
--- a/Src/VimCore/CommonOperations.fs
+++ b/Src/VimCore/CommonOperations.fs
@@ -177,9 +177,9 @@ type internal CommonOperations
         let allowPastEndOfLine = 
             _vimTextBuffer.ModeKind = ModeKind.Insert ||
             _globalSettings.IsVirtualEditOneMore ||
-            VisualKind.IsAnyVisualOrSelect _vimTextBuffer.ModeKind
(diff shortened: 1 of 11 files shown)


jhamm:
Lets hold off on merging this PR. I think the tests should cover both with and without the ve=onemore setting.

jaredpar:
Okay will hold off on merging.

jhamm:
I think it is ready to merge now. Thanks!

We can talk about the exclusion of any visual mode from `CommonOperations.AdjustCaretForVirtualEdit`. That is the behavior I saw in vim with no rc applied as well as with spf13 for both ve= and ve=onemore

Issue #1659, opened 2015-08-17
VsVim StatusLine loads above horizontal scrollbar -- sometimes


gwojan:
Most of the time after a project is loaded the VsVim statusline is located below the horizontal scrollbar. However, opening multiple VS sessions results in the statusline above the scrollbar. It seems totally inconsistent so I haven't been able to detect a pattern.

What is the correct location for the statusline?

I'm including a couple if images just for clarity.

![vsvim statusline below scrollbar](https://cloud.githubusercontent.com/assets/6855215/9306839/d914c18e-44c8-11e5-91b7-e06cf23ca69a.PNG)

![vsvim statusline above scrollbar](https://cloud.githubusercontent.com/assets/6855215/9306845/e355c2ba-44c8-11e5-83bc-128aa0a09e3a.PNG)

jaredpar:
That's truly odd.  I've never seen that before and don't have a good idea when it would happen.  The placement of the command margin is declarative so it's odd that it would behave unpredictably in this manner.  Going to ask around to see if anyone has an idea of why this could be happening.

jaredpar:
CC @jasonmalinowski in case he has an idea here.

gwojan:
I guess I should ask which is the correct location declared for the statusline -- above or below the scrollbar? Which of the above images is "correct"?

jasonmalinowski:
Margins are ordered the same way other orderable things are, with Name and Order attributes. If you didn't specify it, then you are probably at the mercy of the order MEF happens to import things. There's some description of that [here](https://msdn.microsoft.com/en-us/library/dd885244.aspx).

jaredpar:
@jasonmalinowski do you know the name the scroll bar margin?  Or where I should go looking for it?

jasonmalinowski:
Per the link I gave:

>These are the built-in margins:
>
>   * "Wpf Horizontal Scrollbar"
>   * "Wpf Vertical Scrollbar"
>   * "Wpf Line Number Margin"

Are these not working? Looks like there's also a [`PredefinedMarginNames`](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.text.editor.predefinedmarginnames.aspx) class.

gwojan:
@jaredpar One thing I just noticed is if I don't have any additional windows like `Output`, `Error List` pinned like the images above show the VsVim margin consistently renders below the horizontal scroll bar.

If I pin something placement becomes random again.

gwojan:
[...]

PR #1753, opened 2016-03-05
add Order attribute to CommandMarginProvider

jhamm:
Was not able to reproduce the issue from the images. The command
margin continued to appear after the horizontal scrollbar with this
change.


jhamm:
fix #1659

Issue #1733, opened 2016-01-13
Go back to the line just yanked `` is not working


jasjuang:
After we yank a line, and lets say go down 5 lines to paste, `` should take the cursor back to the original line but it isn't

yy
5jp
`` ---> this is not working

Will appreciate if you can look into this.

jaredpar:
Thanks for reporting that.  I will look at getting that implemented.

jhamm:
Hi, I am not able to reproduce this behavior in vim/gvim.
in a buffer of:
````
1
2
3
4
5
6
7
8
9
0
````

from normal mode:
````
10G
m'
1G
yy
5jp
''
````

The cursor is on 11,1.

 If "yy" was overwriting the previous context mark, the cursor should be at position 1,1.

Can you provide more information about this please?

**Comment by jasjuang** [comment]
In this case it should be

```
10G
m'
1G
yy
5jp
`''
```

According to http://vim.wikia.com/wiki/Using_marks

''  jump back (to line in current buffer where jumped from)
`" jump to position where last exited current buffer

jhamm:
Yup, you are right. I was reading ` as '.

jhamm:
I have a working mark " at [jhamm/VsVim/fix-1733](https://github.com/jhamm/VsVim/tree/fix-1733).

I am now working on mark ``

When they are completed, I will create a PR.

jhamm:
This one can be closed, the PR has been merged.

jasjuang:
Thanks a lot for making this happen!

jhamm:
Glad I can help :smile:

PR #1760, opened 2016-03-15
Fix 1733

jhamm:
add " mark (last existed position)
add ' mark (last jump) when new VimBuffer is created
reorder :marks


jhamm:
fix #1733

jhamm:
There were a few changes in this one :smile:.

Let me know if I need to change any of it.

Src/VimCore/CoreTypes.fs
```diff
@@ -312,9 +312,10 @@ type LocalMark =
                 yield LocalMark.Letter letter
             for number in NumberMark.All do
                 yield LocalMark.Number number
+            yield LocalMark.LastInsertExit
```
jaredpar: Good catch.

Src/VimCore/MarkMap.fs
```diff
@@ -92,6 +94,27 @@ type MarkMap(_bufferTrackingService : IBufferTrackingService) =
             vimBufferData.VimTextBuffer.GetLocalMark localMark
         | Mark.LastJump -> 
             vimBufferData.JumpList.LastJumpLocation
+        | Mark.LastExitedPosition ->
+            let getZeroZero() =
+                let snapshotLine' = vimBufferData.TextBuffer.CurrentSnapshot.GetLineFromPosition 0
+                let snapshot = snapshotLine'.Start.Add(0)
```
jaredpar: Can simplify this to:

``` fsharp
SnapshotUtil.Get StartPoint vimBufferData.TextBuffer.CurrentSnapshot
```

Src/VimCore/MarkMap.fs
```diff
@@ -92,6 +94,27 @@ type MarkMap(_bufferTrackingService : IBufferTrackingService) =
...
+                | Some (line, column) -> (line, column)
+                | None -> (0, 0)
+
+            // not using TryGetPointInLine because None is returned when column and lineLength are both 0,
+            //  which can be valid in this case
+            match SnapshotUtil.TryGetLine vimBufferData.TextBuffer.CurrentSnapshot line with
+            | None -> getZeroZero()
+            | Some snapshotLine ->
+                if column >= snapshotLine.Length && not (column = 0 && snapshotLine.Length = 0) then
+                    getZeroZero()
```
jaredpar: Consider the case where `line` is a non-zero value and `column` is the end position on the line.  The function will then return the start position of the entire buffer instead of the current line.  Is that behavior intentional?

jhamm:
Those are good suggestions. I have added 2 new tests and made the changes. Thanks!

Issue #1801, opened 2016-06-30
Mouse selecting from right to left missing the last char.


xuboying:
Version 2.1.1.0
VS: 2013 VC++
Mouse selecting from right to left missing the last char(most right char).

jaredpar:
Thanks for reporting.  I will take a look at this.

xuboying:
Thanks!

slav:
This is also happening on VS 2015

ricksladkey:
Unless I am misunderstanding something, I cannot currently reproduce this bug, with either `behave=mswin` or `behave=xterm`.

ricksladkey:
I now understand this bug better and know that it was caused by a previous change that was reverted by PR #1829, which was merged, so this can be closed.

Issue #1792, opened 2016-06-22
double click word in normal mode does not expands to last letter


xuboying:
VS 2013 VC++
vsvim 2.1.0.0
See gif for details

![screenshot9](https://cloud.githubusercontent.com/assets/5218237/16263596/c2e7561e-38a7-11e6-9cd2-96cccfb69f93.gif)

jaredpar:
Thanks for reporting.  I was able to reproduce this issue.  It effects selections at the end of the line when certain vim options are set.  Should be fixed now in source.

xuboying:
Thanks!!

Issue #1791, opened 2016-06-22
VsVim temporary disable breaks cursor


elrob:
- ctrl + shift + F12 to disable VsVim temporarily (temporarily only - not disabling the extension completely that requires a restart)
- try to move the cursor to the end of the line and it doesn't work
- cursor can only reach the character before the end of the line

The temporary disable functionality is vital for me as I pair-program and often my partner can not use VIM. We need to be able to switch VSVim off and on quickly (without a restart). We've all now had to completely disable VSVim which makes me sad. I love the plug-in otherwise.

This issue has already been mentioned on the Q&A: https://visualstudiogallery.msdn.microsoft.com/59ca71b3-a4a3-46ca-8fe1-0e90e3f79329

jaredpar:
Thanks for reporting this issue.  I'm pretty sure I know what the issue here is and is similar to a bug I just fixed yesterday.

jaredpar:
This bug is fixed now.  If you wait for AppVeyor to complete it's run you can grab a build with the fix off of vsixgallery.com.

raztus:
I'm in the same scenario (pair-programming) and this was an ugly issue. Just installed the latest version from vsixgallery.com, and it fixes the issue. Thanks, @jaredpar!

jaredpar:
Yeah that particular function has been responsible for no less than 3 bugs in the 2.1.0 release.  Didn't quite realize all the ramifications that fix would have on selection.  Getting educated quickly 😄

MartinL0815:
Sorry, but the issue even occurs without disabling VsVim!

Another extension "Entrian Source Search" uses the VS API to get word boundaries and is since updating VsVim not always anymore to get a word at the end of a line for sourch pattern creation.

My observed problematic lines match the RE:

    ^\s+\w+$

If "Entrian Source Search" is triggered to use the word under the caret for the next search it retrieves the first RE group:

    ^\s+(\w+)\w$

So it misses the last character.

I disabled VsVim to see, if the last update causes the problem and even now selecting the text in such a line from Pos1 to End misses the last character of the line.
This character cannot be selected with VsVim being disabled.

As Jared stated the bug is already fixed and the fix should be available, right?

Does this fix includes this described buggy behavior?

jaredpar:
@MartinL0815 it's likely the latest version fixes this problem.  You can install it manually from this location to test out:

[...]

Issue #1789, opened 2016-06-20
Snippets are impaired with VsVim 2.1.0.0


cbates-:
VS2015 Pro, Win10 x64

It appears that VS snippets behavior has been degraded after installing VsVim 2.1.0.0.
I have a snippet with two replaceable items.  I insert the snippet

>xdep\<Tab\>

and the 1st replaceable item is selected; I start typing and replace the item.
When I tab, the cursor now jumps into the 2nd replaceable item, but w/o selecting it -- this is the change in behavior.  Previously, when tabbing to each replaceable item the entire item would be selected, allowing me to type the replacement value.
I hope that description is clear enough....

jaredpar:
Thanks for reporting this.  Are you able to share out this snippet so I can test it directly?  If not I can try and substitute in one of the standard snippets and see if I can replicate the behavior.  

Also: assuming this is C# code?

cbates-:
Thanks for the reply.
Attached is the snippet, it's pretty simple.
Let me know there is any other help I can provide...
C. Bates

jaredpar:
@cbates- the email reply didn't capture the snippet on Github.  Could you throw it up as a gist?

cbates-:
Not that familiar with gist, hope this works....
https://gist.github.com/cbates-/7de750b693191bab49e4ed8f7eb17bda
C. Bates

jaredpar:
That works.  Thanks!

On Mon, Jun 20, 2016 at 4:18 PM, cbates- <notifications@github.com> wrote:

> Not that familiar with gist, hope this works....
> https://gist.github.com/cbates-/7de750b693191bab49e4ed8f7eb17bda
> C. Bates
>
> On Mon, Jun 20, 2016 at 7:08 PM, Jared Parsons <notifications@github.com>
> wrote:
>
> > @cbates- <https://github.com/cbates-> the email reply didn't capture the
> > snippet on Github. Could you throw it up as a gist?
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly, view it on GitHub
> > <https://github.com/jaredpar/VsVim/issues/1789#issuecomment-227296502>,
> > or mute the thread
> > <
> https://github.com/notifications/unsubscribe/ABWN-XNztyOMU1uzSsOlGnuOk2x1EpR5ks5qNx1jgaJpZM4I52dF
> >
> > .
> >
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
[...]

PR #1829, opened 2016-09-15
Don't track `virtualedit` on arbitrary mouse position changes

jaredpar:
This essentially reverts a lot of the behavior in 4279e17676102eba70d182ab8bb2924b8534b58d.

When this change first came through I thought it was a good idea.  It seemed natural to properly enforce the `virtualedit` setting on any caret position change, not just when VsVim was processing key strokes.

Since this has shipped though customers have been sending a swarm of scenarios where this change broke behavior:

- #1789
- #1791
- #1792
- #1801

Fixing all these bugs is beginning to feel like a whack a mole problem.  There is simply no way to generalize the fix such that it excludes all of the places customers expect selection to hit EOL irrespective of the `virtualedit` setting.  As such going to revert back to the old behavior of not checking that on arbitrary mouse position changes.


jaredpar:
closes #1801

Issue #1738, opened 2016-01-26
Yank with a setting clipboard=unnamed should be placed to the register 0 too.


idealist1508:
How to reproduce:

```
:set clipboard=unnamed
yy
:reg 0 
  -> is empty
```

idealist1508:
I was trying to fix it, but i got an Error
The type 'ClipboardOptions' is not defined in
https://github.com/idealist1508/VsVim/blob/CopyToUnnamedFix/Src/VimCore/Register.fs#L569

I don't understand why? 
@jaredpar: Can you help me with it.

jaredpar:
@idealist1508 not sure why you would get that error as that type is defined in the project.  Are you sure nothing else changed which could have moved / deleted that type?

idealist1508:
@jaredpar Yes, I am sure. I only add a new parameter to https://github.com/idealist1508/VsVim/commit/ffb4b1c685803af89bd0769ef2514542c716816e#diff-96e864afc9fb1f7f309e9318eb130074R569
Can you try to compile code from my branche https://github.com/idealist1508/VsVim/tree/CopyToUnnamedFix? Thanks

jaredpar:
The reason you're getting that error is you're trying to use `ClipboardOptions` before it's been defined.  F# processes files in the order they appear in Solution explorer (top down).  In this case `ClipboardOptions` is defined in `VimSettingsInterface.fs` but you're using it in `Registers.fs`.  The latter file gets processed first hence the error.

idealist1508:
You are right. I read about this (and forgot). Thanks.

ernestobarros:
Missing this fix - to make everything work smoothly.

idealist1508:
I can look at this again at weekends. But I am only the beginner in F#, so if somebody fix it i will not cry.

idealist1508:
Hier is my Fix #1830

PR #1830, opened 2016-09-18
Fix for #1738: Yank with a setting clipboard=unnamed should be placed to the register 0 too. 

idealist1508:
Please review my changes


Src/VimCore/RegisterMap.fs
```diff
@@ -100,9 +100,19 @@ type internal RegisterMap (_map : Map<RegisterName, Register>) =
     /// Updates the given register with the specified value.  This will also update 
     /// other registers based on the type of update that is being performed.  See 
     /// :help registers for the full details
-    member x.SetRegisterValue (reg : Register) regOperation (value : RegisterValue) = 
+    member x.SetRegisterValue (register : Option<Register>) regOperation (value : RegisterValue) (clipboardOptions : Vim.ClipboardOptions) =
```
jaredpar: I feel like this might be the wrong approach.  The `RegisterMap` and `Register` types are designed to be unaware of settings.  All of the settings aware code was meant to be layered on top of them.  The approach I had in mind for this fix was to build on top of the existing code which attempts to access `IRegisterMap` in a settings aware manner.

Src/VimCore/CommandUtil.fs
```diff
@@ -944,8 +944,13 @@ type internal CommandUtil
 
     /// Get the appropriate register for the CommandData
     member x.GetRegister (commandData : CommandData) =
-        let name = 
+            match commandData.RegisterName with
+            | Some name -> Some (_registerMap.GetRegister name)
+            | None -> None
 
+    /// Get the appropriate register for the CommandData
+    member x.GetRegisterForPut (commandData : CommandData) =
```
jaredpar: Consider if there  was a `SetRegisterValue` method here.  This method could take into account the `clipboard` option and update both the clipboard and unnamed register here.  This would be a nice pairing with `GetRegister` which has similar logic.

Src/VimCore/RegisterMap.fs
idealist1508: Yeah,  I had a bad feeling doing this change, but i have no idea how can i do it better.

Do you mean something like that?
* the function SetRegisterValue takes a list of registers (or may be one register) to update, and updates only those/this registers without any extra logic
* some other function, which depend on configuration, parses user input, converts it to a list of registers for update and calls for each register the function SetRegisterValue.

Src/VimCore/CommandUtil.fs
idealist1508: But we need still two differnt methods for GetRegister. One should  get a concrete register to read value and the other one should get a list of registers to update.

Src/VimCore/RegisterMap.fs
jaredpar: I was thinking of something like `CommandUtil::SetRegisterValue (commandData : CommandData) text`.  The implementation could look at the state of `ClipboardOptions` and update both registers if necessary.  It already has access to settings so this should be doable.

idealist1508:
I am sorry. I'll pass.  Maybe i can look at this in a month  or so.

jaredpar:
I ended up fixing this as part of some overhaul of the register system.  See the following PR: 

https://github.com/jaredpar/VsVim/pull/1878

Issue #1837, opened 2016-09-23
gg with long count jumps to top rather than bottom


rcorre:
In vim, if you give `gg` a count larger than the number of lines in the file, it jumps to the bottom of the file. In vsvim, it jumps to the top.

jaredpar:
Thanks for reporting.  I will take a look at that.

jaredpar:
This was fixed by 2cd5e84cee0c6ef7e8b72e5adf2335faa6ae5785

PR #1853, opened 2016-10-23
Fixing bug #1837. When you give the 'gg' command a count greater than…

chaitanyagurrapu:
… the number of lines in the file, it now goes to the bottom of the file, just like in Vim


jaredpar:
Thanks for the contribution!

PR #1883, opened 2016-12-14
Proof of concept 'l' over surrogate pairs

john-patterson:
There is an issue where if you try to navigate over a unicode surrogate
pair, you cannot move right over it. If you are before it, you will end up
in the character and get rounded back down to where you were. If you are
on the left of the character and try to move back to the right over the
character, you will end up in the character and get rounded back to the
intended position.

This fixes that in the case of insert mode, but the code is ugly and the
seam that I'm explointing isn't the most elegant. This serves as a good
base to start the fix for both normal and insert mode.

This is a partial fix to [issue 1786](https://github.com/jaredpar/VsVim/issues/1786).

### DIFF
--- a/Src/VimCore/CommonOperations.fs
+++ b/Src/VimCore/CommonOperations.fs
@@ -446,8 +446,11 @@ type internal CommonOperations
 
         /// Move the caret right.  Don't go off the end of the line
         let moveRight () =
-            if x.CaretPoint.Position < x.CaretLine.End.Position then
-                let point = SnapshotPointUtil.AddOne x.CaretPoint
+            let point =
+                x.CaretPoint
+                |> SnapshotPointUtil.AddOne
+                |> (fun y -> if Char.IsSurrogate (y.GetChar()) then SnapshotPointUtil.AddOne y else y)
+            if point.Position <= x.CaretLine.End.Position then
                 x.MoveCaretToPoint point ViewFlags.Standard
                 true
             else


Src/VimCore/CommonOperations.fs
```diff
@@ -446,8 +446,11 @@ type internal CommonOperations
 
         /// Move the caret right.  Don't go off the end of the line
         let moveRight () =
-            if x.CaretPoint.Position < x.CaretLine.End.Position then
-                let point = SnapshotPointUtil.AddOne x.CaretPoint
+            let point =
+                x.CaretPoint
+                |> SnapshotPointUtil.AddOne
+                |> (fun y -> if Char.IsSurrogate (y.GetChar()) then SnapshotPointUtil.AddOne y else y)
```
jaredpar: This code will throw in the case where the caret is positioned on the last character of a file.  This occurs because `GetChar` throws when on the end of the `ITextSnapshot`.  

Also I think we need to change this to look for the Low Surrogate.  Otherwise moving right to the start of a surrogate pair would cause it to jump to the middle.

Src/VimCore/CommonOperations.fs
```diff
@@ -446,8 +446,11 @@ type internal CommonOperations
 
         /// Move the caret right.  Don't go off the end of the line
         let moveRight () =
-            if x.CaretPoint.Position < x.CaretLine.End.Position then
-                let point = SnapshotPointUtil.AddOne x.CaretPoint
+            let point =
+                x.CaretPoint
+                |> SnapshotPointUtil.AddOne
+                |> (fun y -> if Char.IsSurrogate (y.GetChar()) then SnapshotPointUtil.AddOne y else y)
+            if point.Position <= x.CaretLine.End.Position then
```
jaredpar: Think we need to add a similar fix for `CharLeft`.  Even though the caret functionally moves left today it's broken for other scenarios.  Consider when my caret is to the right of a surrogate pair.  In that case I would expect the normal mode command `yh` would yank the full unicode character.  Just as `yl` would yank the unicode character when the caret was at the start of the character. 

In order to do that `CharLeft` needs to fully encompass the character.

jaredpar:
Closing outdated PRs

PR #1966, opened 2017-06-16
Fixed divide by 0 error in modulus operator

celie56:


Src/VsVimShared/VsVimHost.cs
```diff
@@ -607,13 +607,13 @@ private bool MoveFocusHorizontally(int indexDelta)
             topLevelWindows.Sort((a, b) => a.Left < b.Left ? -1 : 1);
             var indexOfActiveDoc = topLevelWindows.FindIndex(win => win == _dte.ActiveWindow);
             var movedIndex = indexOfActiveDoc + indexDelta;
-            var newIndex = (movedIndex < 0 ? movedIndex + topLevelWindows.Count : movedIndex % topLevelWindows.Count);
+            var newIndex = (movedIndex < 0 ? movedIndex + topLevelWindows.Count : movedIndex % (double)topLevelWindows.Count);
```
jaredpar: Rather than casting to `double` here I think we should just catch the case where count is `0` and handle that specifically. Likely by just returning false.

jaredpar:
Fixed via #2044

Issue #1995, opened 2017-09-05
Repeat last colon command [Feature-Request]


wongjiahau:
Based on the Vim [wikia](http://vim.wikia.com/wiki/Repeat_last_colon_command), the last executed colon command (Ex command) can be repeated by using `@:` , and further repeat can be done with `@@`.  
One useful usage is when I want to repeat the last substitute command, say `:s/old/new` .

mletterle:
Duplicate of #1115

PR #1996, opened 2017-09-05
Laying foundation to complete #1995 (Repeat last ex command)

wongjiahau:
I'm sorry if I've done any syntax mistake, because I'm newbie to F#.
I think currently that is the best I can do for now to implement this feature.


jaredpar:
Closing out-dated PRs

PR #2012, opened 2017-10-17
Issue 2003: Fix focus not taken on command mode or search entry

AshrafMichail:
The bug is CommandMarginController calculates EditKind changes on OnKeyStart but the
_vimBuffer used to compute that doesn't reflect the state until OnKeyEnd.

The state stays wrong during the entire duration of command mode/search unless the users clicks.

Fix is to correct state in OnKeyInputEnd.

### DIFF
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginControl.xaml.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginControl.xaml.cs
@@ -125,6 +125,11 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
             InitializeComponent();
         }
 
+        internal bool IsCaretAtStart()
+        {
+            return _commandLineInput.CaretIndex == 1;
+        }
+
         internal void UpdateCaretPosition(EditPosition editPosition)
         {
             if (IsEditReadOnly)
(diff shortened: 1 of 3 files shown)


jaredpar:
Looks like there were a couple of unit test failures here. Have you had a chance to look at them?

AshrafMichail:
Yes, I took a look and it did find a regression.  It seems there’s some missing cases to prevent cursor from going before the mode indicator character.

I plan to fix this, but it’s low priority.

I can close the pull request and submit later if you like.


  *   Ashraf Michail

jaredpar:
Go ahead and leave it open for now if you're planning on fixing up the issues.

AshrafMichail:
Ok.  I am fixing it in the background but have one last issue I need to get to when I have time.

BTW, I’m assuming to get full unit tests to pass I need visual studio 2015 installed, is that the case?  I’m getting missing intellisense objects that seem to be visual studio 2015 objects.


  *   Ashraf Michail

AshrafMichail:
Still a work in progress, there are a number of places in the code that rely on state being EditKind.None when we are in command mode that are causing problems.

jaredpar:
Looks great! Thanks for doing this.

PR #2055, opened 2018-01-05
Remove "This is a test" from License.txt

jennings:
I was happy to update VsVim today, but noticed some surprising text in the installation dialog. I'm guessing "This is a test" isn't supposed to be there?

image


jaredpar:
Oh man ... I wonder how long this has been there . Thanks for fixing!

PR #2097, opened 2018-04-22
Fix command/search history

ricksladkey:
- Fix accidental handling of Down same as Up
- Add Ctrl-N and Ctrl-P as synonyms for Down and Up
- Fix backspacing past the beginning of command/search isn't like Escape
- Fix history previous after history next to an empty command (with test)

### DIFF
--- a/Src/VimCore/HistoryUtil.fs
+++ b/Src/VimCore/HistoryUtil.fs
@@ -155,7 +155,8 @@ type internal HistorySession<'TData, 'TResult>
             _historyClient.Beep()
         | HistoryState.Index (list, index) -> 
             if index = 0 then
-                _clientData <- _historyClient.ProcessCommand _clientData ""
+                _command <- ""
+                _clientData <- _historyClient.ProcessCommand _clientData _command
                 _historyState <- HistoryState.Empty
             else
                 x.DoHistoryScroll list (index - 1)
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
@@ -333,7 +333,7 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
         /// </summary>
         private bool HandleHistoryNavigation(KeyInput keyInput)
         {
-            var handled = _vimBuffer.Process(KeyInputUtil.VimKeyToKeyInput(VimKey.Up)).IsAnyHandled;
+            var handled = _vimBuffer.Process(keyInput).IsAnyHandled;
             var prefixChar = GetPrefixChar(_editKind);
             if (handled && _editKind != EditKind.None && prefixChar.HasValue)
             {
@@ -375,7 +375,7 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
                     e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Up));
                     break;
                 case Key.Down:
-                    e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Up));
+                    e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Down));
                     break;
                 case Key.Home:
                     if ((e.KeyboardDevice.Modifiers & ModifierKeys.Shift) == 0)
@@ -393,8 +393,12 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
                     break;
                 case Key.Left:
                 case Key.Back:
-                    // Ignore backspace if at start position
-                    e.Handled = _margin.IsCaretAtStart();
+                    if (_margin.IsCaretAtStart())
+                    {
+                        _vimBuffer.Process(KeyInputUtil.EscapeKey);
+                        ChangeEditKind(EditKind.None);
+                        e.Handled = true;
+                    }
                     break;
                 case Key.R:
                     if (e.KeyboardDevice.Modifiers == ModifierKeys.Control)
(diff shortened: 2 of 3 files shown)


ricksladkey:
I found that my old friend VimApp cannot be run or debugged from Visual Studio 2017 with the current VimApp.csproj file. This probably means @jaredpar has not used VimApp recently, but I find basic development of VsVim much easier than debugging Visual Studio with the experimental instance.

Since my own fix was a bit of hack, and since I can't imagine anyone else who would be better suited to figuring out why running/debugging doesn't work, I'll leave that problem for @jaredpar to solve!

Src/VimCore/HistoryUtil.fs
```diff
@@ -155,7 +155,8 @@ type internal HistorySession<'TData, 'TResult>
             _historyClient.Beep()
         | HistoryState.Index (list, index) -> 
             if index = 0 then
-                _clientData <- _historyClient.ProcessCommand _clientData ""
+                _command <- ""
+                _clientData <- _historyClient.ProcessCommand _clientData _command
                 _historyState <- HistoryState.Empty
```
ricksladkey: This is the fix that the new test tests below.

Test/VimCoreTest/CommandModeIntegrationTest.cs
```diff
@@ -534,6 +534,15 @@ public void NextCommandAlternateKeystroke()
...
 
+            [WpfFact]
+            public void PreviousCommandAfterEmptyHistory()
+            {
+                Create("");
+                _commandHistoryList.AddRange("dog", "cat");
+                _vimBuffer.ProcessNotation(":<C-p><C-n><C-p><C-p>");
+                Assert.Equal("dog", _commandMode.Command);
+            }
+
```
ricksladkey: This is very easy test to do by hand with vim to see that it is a good test.

Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -393,8 +393,12 @@ internal void HandleKeyEvent(KeyEventArgs e)
...
                 case Key.Left:
                 case Key.Back:
-                    // Ignore backspace if at start position
-                    e.Handled = _margin.IsCaretAtStart();
+                    if (_margin.IsCaretAtStart())
+                    {
+                        _vimBuffer.Process(KeyInputUtil.EscapeKey);
+                        ChangeEditKind(EditKind.None);
+                        e.Handled = true;
+                    }
```
ricksladkey: This makes backspacing past the beginning exit command/search like vim.

Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -375,7 +375,7 @@ internal void HandleKeyEvent(KeyEventArgs e)
                     e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Up));
                     break;
                 case Key.Down:
-                    e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Up));
+                    e.Handled = HandleHistoryNavigation(KeyInputUtil.VimKeyToKeyInput(VimKey.Down));
```
ricksladkey: This look like it was accidental, but previous Down a synonym for Up.

Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -333,7 +333,7 @@ private void UpdateCommandLine(string commandLine)
         /// </summary>
         private bool HandleHistoryNavigation(KeyInput keyInput)
         {
-            var handled = _vimBuffer.Process(KeyInputUtil.VimKeyToKeyInput(VimKey.Up)).IsAnyHandled;
+            var handled = _vimBuffer.Process(keyInput).IsAnyHandled;
```
ricksladkey: This looks like it was accidental and the intent was to use the supplied `keyInput` for the processing.

Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -425,6 +429,18 @@ internal void HandleKeyEvent(KeyEventArgs e)
...
+                    {
+                        e.Handled = HandleHistoryNavigation(KeyInputUtil.ApplyKeyModifiersToChar('p', VimKeyModifiers.Control));
+                    }
+                    break;
+                case Key.N:
+                    if (e.KeyboardDevice.Modifiers == ModifierKeys.Control)
+                    {
+                        e.Handled = HandleHistoryNavigation(KeyInputUtil.ApplyKeyModifiersToChar('n', VimKeyModifiers.Control));
+                    }
+                    break;
```
ricksladkey: This `Ctrl+N` and `Ctrl+P` as synonyms for `Down` and `Up` like vim.

jaredpar:
>  This probably means @jaredpar has not used VimApp recently

I actually use VimApp as well for the reasons you described. It has been about a month or so though since I used it. Perhaps a recent update of VS2017 broke it and I just haven't hit it yet. What problem were you running into?

Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
jaredpar: Yeah that definitely looks like a typo in the original code.

jaredpar:
Thanks @ricksladkey!

ricksladkey:
Here's what happens when I select VimApp as the StartUp Project and press `F5` or `Ctrl+F5`:

image

It does work if I specify "Launch -> Executable" and supply the full pathname to an executable, but that is risky because it doesn't obey the configuration and so I might build the Release executable and debug the Debug one.

PR #2099, opened 2018-04-24
Recenter view when moving offscreen by more than half a screen

### LINKED ISSUE #1501: Center the caret position in the window when moving off screen
When I perform a buffer motion with VsVim and the caret position would be off-screen below the current position, the window is usually adjusted so that the line containing the caret is the last line of the screen. By constrast, it seems to be the convention in gvim (and even Visual Studio itself) for off-screen motions to place the line containing the caret vertically centered in the window.

I know this is a small thing but I looked for an hour or so on how to fix this (or at least make it configurable), but I cannot seem to find the right part of the code.  I did narrow it down to the call to `ITextCaret.EnsureVisible`, which is documented to behave in the way that I am trying to fix, and yet that interface doesn't seem to have way to get the "normal" behavior.

Any tips on how I would go about changing this?

ricksladkey:
Fixes #1501

### DIFF
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -1782,6 +1782,38 @@ module TextViewUtil =
         let caret = GetCaret textView
         caret.EnsureVisible()
 
+    /// Check whether we should recenter the view on point
+    let CheckRecenterOnPoint (textView : ITextView) point =
+        let textViewLine = textView.GetTextViewLineContainingBufferPosition point
+        let onScreen = textViewLine.VisibilityState = Formatting.VisibilityState.FullyVisible
+        if not onScreen then
+
+            // The line the point is on is not on screen.
+            // We handle three cases.  The point line:
+            // 1) precedes the first visible line by less than half a screen
+            // 2) follows the last visible line by less than half a screen
+            // 3) is further away than half a screen
+            // Vim uses pseudo-scrolling for cases 1) and 2) and centers
+            // the point vertically within the screen for 3).
+            let pointLine = SnapshotPointUtil.GetContainingLine point
+            let pointIndex = pointLine.LineNumber
+            let textViewLines = textView.TextViewLines
+            let firstVisibleLine = SnapshotPointUtil.GetContainingLine textViewLines.FirstVisibleLine.Start
+            let firstIndex = firstVisibleLine.LineNumber
+            let lastVisibleLine = SnapshotPointUtil.GetContainingLine textViewLines.LastVisibleLine.Start
+            let lastIndex = lastVisibleLine.LineNumber
+            let scrollLimit = int(ceil(textView.ViewportHeight / textView.LineHeight / 2.0))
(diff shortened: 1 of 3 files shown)


Src/VimCore/EditorUtil.fs
```diff
@@ -1782,6 +1782,43 @@ module TextViewUtil =
         let caret = GetCaret textView
         caret.EnsureVisible()
 
+    /// Check whether we should recenter the view on point
+    let CheckRecenterOnPoint (textView : ITextView) point =
+
+        // Be careful because using ITextViewLines can result in an exception.
```
jaredpar: When can this exception happen?

Src/VimCore/EditorUtil.fs
```diff
@@ -1782,6 +1782,43 @@ module TextViewUtil =
...
+                // The line the point is on is not on screen.
+                // We handle three cases.  The point line:
+                // 1) precedes the first visible line by less than half a screen
+                // 2) follows the last visible line by less than half a screen
+                // 3) is further away than half a screen
+                // Vim uses pseudo-scrolling for cases 1) and 2) and centers
+                // the point vertically within the screen for 3).
+                let pointLine = SnapshotPointUtil.GetContainingLine point
+                let pointLineNumber = pointLine.LineNumber
+                let textViewLines = textView.TextViewLines
```
jaredpar: The `ITextView.TextViewLines` property can also return `null` in some cases. `TextViewUtil.GetTextViewLines` tries to encapsulate this, and the exception oddity, in an option.

Src/VimCore/EditorUtil.fs
ricksladkey: I did not personally experience these exceptions. I added the try block because I saw similar code elsewhere in VsVim, for example in `CommandUtil.ScrollLines`:

```fsharp
        try
...
        with
        // Dealing with ITextViewLines can lead to an exception (particularly during layout).  Need
        // to be safe here and handle the exception.  If we can't access the ITextViewLines there
        // isn't much that can be done for scrolling
        | _ -> ()
```
The documentation suggests certain methods can throw `InvalidOperationException`, for example `GetTextViewLineContainingBufferPosition` for the reason:

> The view has not completed initialization.

Rather than, say, "during layout." It seems wise to prevent these exceptions, should they occur, especially since what we are doing is entirely cosmetic.

Src/VimCore/EditorUtil.fs
ricksladkey: I've only recently understood the advantage of the utility functions, for example for currying, for syntax, and for cases like this where an option result is a better match for F#. I'll change this code to use the method you describe.

ricksladkey:
I converted the direct use of `TextViewLines` to use `GetTextViewLines` based on your comment.

I also noticed another special case of scrolling off-screen that I missed in the previous version that involves scrolling to regions near the bottom of the file. In that case, if centering the screen on point would leave blank space after the last visible line, vim positions the last line of the buffer as the last line of the screen.

jaredpar:
Thanks!

PR #2102, opened 2018-04-26
Fix '$' line and '%' line range in buffers with a trailing linebreak

### LINKED ISSUE #2101: Last line and entire buffer include too many lines when the buffer ends in a linebreak
If the buffer contains:

```
cat\r\n
dog\r\n
```

and this substitute command is issued:

`%s/^/x/`

then the buffer will contain:

```
xcat\r\n
xdog\r\n
x
```

in other words, we have inadvertently modified the last phantom line.

Although buffers without a trailing linebreak are a fact of life and must be supported in a modern editor, traditional Unix/DOS/Windows/vi/vim text files always end in a linebreak. When then do, vim and VsVim should treat them traditionally. That is:

```
cat\r\n
dog\r\n
```

should be treated as though it had two lines.

ricksladkey:
Fixes #2101


jaredpar:
Thanks!

Issue #757, opened 2012-02-10
Insert mode Ctrl-G support


philipmat:
I use `inoremap <Space> <Space><C-g>u` in my _vimrc file to have Vim generate an undo point on (roughly) every word.

However, in VsVim 1.1.2 when pressing space it inserts `gu` (space,g,u) instead.

ricksladkey:
PR #2106 implemented this feature.

PR #2106, opened 2018-05-01
Fix undo for inserts with corrections

### LINKED ISSUE #771: Undo still is to fine-granular
If I do the following in VsVim:

```
<i>Hello<Del><Del><Del>allo Welt<Escape>
```

Than I undo step by step:

```
Hallo Welt
<u>
H
<u>
He
<u>
Hel
<u>
Hell
<u>
Hello
<u>
```

So I need 6 undoes in VsVim for one insert.
In VIM one `<u>` undoes the whole insert!

### LINKED ISSUE #1096: Insert-mode mappings should not affect the undo stack
Extracted from #1095.

[These mappings in my vimrc](https://github.com/roryokane/dotvim/blob/2c1ccd2d115d0782fd03633e3768fb8d80e28f5d/vimrc#L298-L302) demonstrate the problem:

> ``` vim
> " keep indentation if I press Esc right after o or O
> nnoremap o ox<BS>
> nnoremap O Ox<BS>
> " also keep indentation if I press Esc right after a newline in Insert mode
> inoremap <Esc> x<BS><Esc>
> ```
> 
> These mappings work around a problem with Vim where it deletes all indentation on a newly created blank line _unless_ you have typed some text onto it already. The mappings type an "x" and delete it immediately so that if I press Esc afterwards, the indentation remains.
> 
> This mappings have an undesirable effect on VsVim. If I use `o` to open a new line and then undo that, VsVim undoes it in five steps: add an "x" back, delete the "x" (from the `<Esc>` mapping), add an "x" back, delete the "x" (from the `o` mapping), and finally delete the new line. Vim would undo all that in one step. Since I never typed the "x"s, I would prefer that they not go in the Undo history.
> 
> The difference is caused by VsVim handling Undo differently from Vim. Vim always undoes everything done in Insert mode all at once. But VsVim uses Visual Studio's standard undo in Insert mode, so it remembers that the "x"s were added and deleted.

> [The problem is that] undo is handled differently in VsVim. I think either VsVim should group commands invoked by insert-mode mappings into the previous undo group, or, like Vim, it should group all commands in insert mode into a single undo group.

ricksladkey:
This change fixes undoing all changes from a single insertion with a single undo.

Previously the presence of backspace or other non-text insertion actions during an ordinary insert would cause the insert to be completely undone only after multiple undo steps.

* Fixes #771
* Fixes #1096

### DIFF
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -178,6 +178,9 @@ type ActiveEditItem =
     /// In Replace mode, will overwrite using the selected Register
     | OverwriteReplace
 
+    /// In the middle of an undo operation.  Waiting for the next key
+    | Undo 
+
     /// No active items
     | None
 
@@ -307,6 +310,7 @@ type internal InsertMode
                 ("<C-o>", RawInsertCommand.CustomCommand this.ProcessNormalModeOneCommand)
                 ("<C-p>", RawInsertCommand.CustomCommand this.ProcessWordCompletionPrevious)
                 ("<C-r>", RawInsertCommand.CustomCommand this.ProcessPasteStart)
+                ("<C-g>", RawInsertCommand.CustomCommand this.ProcessUndoStart)
             |]
             |> Seq.map (fun (text, rawInsertCommand) ->
                 let keyInput = KeyNotationUtil.StringToKeyInput text
@@ -710,14 +714,17 @@ type internal InsertMode
             _commandRanEvent.Trigger x args
 
             // If there is an existing undo transaction, close it out and start a new one.
-            match _sessionData.Transaction with
-            | None -> ()
-            | Some transaction ->
-                transaction.Complete()
-                let transaction = _undoRedoOperations.CreateLinkedUndoTransactionWithFlags "Insert after motion" LinkedUndoTransactionFlags.CanBeEmpty
-                _sessionData <- { _sessionData with Transaction = Some transaction }
+            x.BreakUndoSequence "Insert after motion" 
 
         ProcessResult.OfCommandResult result
+
+    member x.BreakUndoSequence name =
+        match _sessionData.Transaction with
+        | None -> ()
+        | Some transaction ->
+            transaction.Complete()
+            let transaction = _undoRedoOperations.CreateLinkedUndoTransactionWithFlags name LinkedUndoTransactionFlags.CanBeEmpty
+            _sessionData <- { _sessionData with Transaction = Some transaction }
 
     /// Paste the contents of the specified register with the given flags 
     ///
@@ -824,6 +831,12 @@ type internal InsertMode
         _sessionData <- { _sessionData with ActiveEditItem = if _isReplace then ActiveEditItem.OverwriteReplace else ActiveEditItem.Paste }
         ProcessResult.Handled ModeSwitch.NoSwitch
 
+    /// Start a undo session in insert mode
+    member x.ProcessUndoStart keyInput =
+        x.CancelWordCompletionSession()
+        _sessionData <- { _sessionData with ActiveEditItem = ActiveEditItem.Undo }
+        ProcessResult.Handled ModeSwitch.NoSwitch
+
     /// Process the second key of a paste operation.  
     member x.ProcessPaste keyInput = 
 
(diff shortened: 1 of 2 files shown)


jaredpar:
Believe the logic for insert undo is a bit more complex. Essentially combinations like moving the caret left plus typing new text creates new undo units within the insertion. The rules for this aren't documented but it had to be implemented in order to have proper <kbd>.</kbd> (repeat insert command). Trying to find that PR but having trouble at the moment.

ricksladkey:
If you can find cases that fail (and I'll try some stuff myself), I could try to address them in another PR.

jaredpar:
I think we should address these up front. Undo is an area that I get a lot of feedback on when it's done outside expectations. In particular I've gotten very strong feedback when undo has reverted significantly more than users were expecting. 

As for what the logic is here as best as I've been able to determine it's mostly captured in `InsertMode.CreateCombinedEditCommand`. I've found it's easier to link the actions which are known to be safe to combine vs. excluding the unsafe ones.

ricksladkey:
OK, I found a case like you previously mentioned. Starting with an empty buffer:

`idog cat<Left><Left><Left>bat <Esc>`

Gives:

`dog bat cat`

and one undo `u` gives:

`dog cat`

and the second empties the buffer.

After my change, the whole undo is undone in one fell swoop.

If I handle this case (and presumably similar cases), is that what you are looking for?

ricksladkey:
I pushed another commit to break the undo sequences for the case I described, with a correspond test.

ricksladkey:
I added another commit that implements `<C-g>u` to break the undo sequence because it was easy with the new infrastructure.

* Fixes #757

Test/VimCoreTest/NormalModeIntegrationTest.cs
```diff
@@ -2734,6 +2734,91 @@ public void Issue1266()
...
+            /// Undo of change word with backspaces
+            /// </summary>
+            [WpfFact]
+            public void Undo_ChangeWordWithBackspaces()
+            {
+                Create("aaa bbb ccc");
+                _vimBuffer.ProcessNotation("1Gwcwddd<BS><BS>ef<Esc>");
+                Assert.Equal("aaa def ccc", _textView.GetLine(0).GetText());
+                _vimBuffer.ProcessNotation("u");
+                Assert.Equal("aaa bbb ccc", _textView.GetLine(0).GetText());
```
jaredpar: Can we add some tests for the up and down arrow keys as well? Those are the other case where I know of that the undo sequence can get broken up.

ricksladkey: Sure, I added one test that uses `<C-Left>` and another that uses `<Down>`. The implementation detects edit vs. non-edit using the flag specified in the key binding, i.e. whether `CommandFlags.InsertEdit` is present. None of the arrow keys or arrow keys with modifiers have that flag set.

Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -816,6 +831,12 @@ type internal InsertMode
         _sessionData <- { _sessionData with ActiveEditItem = if _isReplace then ActiveEditItem.OverwriteReplace else ActiveEditItem.Paste }
         ProcessResult.Handled ModeSwitch.NoSwitch
 
+    /// Start a undo session in insert mode
+    member x.ProcessUndoStart keyInput =
+        x.CancelWordCompletionSession()
+        _sessionData <- { _sessionData with ActiveEditItem = ActiveEditItem.Undo }
```
jaredpar: How does the transaction get reset in this case? Believe that is necessary in order to implement the command properly. Is it done incidentally as a part of handling this.

ricksladkey: This is in support of `<C-g>u` command (`:help i_Ctrl-G_u` in vim). There are (at least) two commands related to undo in vim, only the first of which I implemented.

In this context `ProcessUndoStart` means "process the first key of a multi-key sequence related to undo. Perhaps the name `ProcessUndoStart`  sounds like it is undoing something but in reality it is just the intermediate state halfway through responding to `<C-g>` followed by some key.

As far as the undo transaction, it will get closed out and reopened once the `u` key is typed, all without ever leaving insert mode.

jaredpar:
Thanks! This is a tricky area. Want to get it merged so I can start dogfooding and looking for cases to test out here. Even with all the tests, still an area that makes me nervous 😄

ricksladkey:
Thanks. Yes, our worst nightmare is something that glitches undo so bad it doesn't work at all right when it is needed the most. The dreaded "lost work." Users entrust the editor with the precious contents their unsaved buffers and bad behavior is not to be taken lightly!

jaredpar:
Indeed. At least the last time VsVim had an undo glitch redo still functioned. So work wasn't lost it just took several minutes to undo and then several more minutes to re-appear. 😄

Issue #1852, opened 2016-10-22
G does not move to the last line if it is an empty line.


chaitanyagurrapu:
In Vim, if the last line is an empty line, pressing G will go to that line. In VsVim it doesn't do that. One possible fix is to change the 'GetLastLine' behavior in EditorUtils. I have a PR up with the change. Although, it is more of a roll back as the original code is from here https://searchcode.com/codesearch/view/9488852/# so I am sure it was originally written and later deleted by @jaredpar 

Alternately, there seems to be a fix possible in VsVim itself as I came across these unit tests ...

![image](https://cloud.githubusercontent.com/assets/853059/19619601/bd154892-98ae-11e6-8537-97ad18f9f2f2.png)

Clearly there should be a way to move to the last line on a 'G' without changing editor utils. I can have a look and try to submit a PR if you could specify what would be a preferred fix.

paul2t:
I think that this issue is back with the latest version of vsvim (2.6.0.0)
When the last line is empty:
`G` goes to the line before the last one.
Moreover, `j` also stops at the line before the last.
`}` and `)` are working fine.

ricksladkey:
@paul2t The best way to diagnose this is to open the same file in VsVim and gvim and test the various commands. If you do this, you will see the `G` and `j` work the same way in Vsvim as they do in gvim. However, you are right that `}` and `)` can move to the phantom line. If you want to open a new issue to fix `}` and `)`, I'll take a look at it.

paul2t:
Oh ok, you are right. Sorry, I didn't read your previous comment correctly.
I understand now. in vim if the last line is empty, it is not shown. Unlike visual studio. That's why I thought this was a bug.
Thanks for your work, this is a great plugin.

PR #2107, opened 2018-05-01
Fix 'G' in a buffer that ends with a linebreak

ricksladkey:
This PR reverts "Fix G on blank last line" but keeps and fixes the corresponding test.

The incorrect behavior was introduced in commit 98b6da61ff2f4cf46eab6b9c74cc23ededb72022 and is not yet visible in the released version.

The issue that commit was intended to fix, #1852, should indeed stay closed because the reported behavior is not a bug.


ricksladkey:
Appveyor failure appears to be spurious.

PR #2112, opened 2018-05-05
Phantom line fixes

ricksladkey:
- Add concept of normalized lines to the editor APIs
- Add 'before save' event to VimHost
- Implement vim 'endofline' and 'fixeondofline' local options
- Add 'settings changed' and 'before save' event handlers in VimBuffer
- Eliminate special cases manually handling normalized lines
- Fix common cases where the cursor could be moved to the phantom line
- Fix number of lines reported by 'C-g' to agree with vim
- Fix simple insertion into an empty buffer creates exactly one normalized line
- Adapt a handful of test that conflicted with the new 'endofline' management
- Add many tests relating to the phantom line

The main goal of these changes is to make VsVim behave more like vim when editing "traditional" text files that end in a final newline. From a user's point of view, the main difference is that it is now much more difficult to accidentally position the cursor after the final newline of a traditional text file. In all cases, the guiding principle is to behave exactly as vim would for the same file contents.

For example, if the buffer contains:

```
cat\r\n
dog\r\n
```

positioning the cursor on the `c` of `cat` and pressing `w` two times in vim moves the cursor to the `g` in `dog`. By contrast, VsVim would previously move the cursor to the non-line after the last line, the so-called "phantom" line.

The phantom line is the empty line following a final newline in a buffer. In this case, the Visual Studio editor APIs treats the buffer as having one more line than vim does. To address this, we introduce the concept of "normalized" lines and use them where appropriate for motion and editor operations. Most special cases can now be handled by simply using the correct flavor of the utility function.

Vim handles the modern phenomenon of files without a final newline using the options 'endofline' and 'fixendofline', both of which default to true. Setting 'fixendofline' dictates that all lines should be normalized lines, i.e. end with a linebreak. If 'fixendofline' is not set, vim uses the special local setting 'endofline' to manage files without a final newline. Upon creation of a buffer, vim record whether all lines have linebreaks in the 'endofline' variable and obeys that setting when writing the file.

In VsVim, in deference to the Visual Studio ecosystem and VsVim tradition, 'fixendofline' defaults to false. Recent versions of Visual Studio support the EditorConfig option 'insert_final_newline' which means roughly the same thing as 'fixendofline' but is considerably more flexible. By default 'fixendofline' is not set and VsVim does not interfere with EditorConfig's behavior.


Src/VimCore/EditorUtil.fs
```diff
@@ -43,6 +43,16 @@ module internal EditorCoreUtil =
     let IsInsideLineBreak (point: SnapshotPoint) (line: ITextSnapshotLine) = 
         point.Position >= line.End.Position && not (IsEndPoint point)
 
+    let EndsWithLineBreak (snapshot: ITextSnapshot) = 
+        let lineNumber = snapshot.LineCount - 1 
+        lineNumber > 0 && snapshot.GetLineFromLineNumber(lineNumber).Length = 0
+
+    let GetNormalizedLineCount (snapshot: ITextSnapshot) =
```
jaredpar: Seems most of the functionality comes down to this line. May want to add a comment here explaining what the normalized line count is and why it's important.

Src/VimCore/InsertUtil.fs
```diff
@@ -296,6 +296,16 @@ type internal InsertUtil
 
     member x.Insert text =
         if _editorOperations.InsertText(text) then
+            if _localSettings.EndOfLine then
+
+                // Ensure that the caret line has a linebreak.
+                let caretPoint = x.CaretPoint
+                let caretLine = x.CaretLine
```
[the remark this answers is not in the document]
ricksladkey: The concept of the last line is a confusing term because of the difference between the last vim line and the last text buffer line. Only the very last line of the text buffer can have a missing linebreak. So the logic asks unambiguously "does the line that the insertion will be performed on have a line break?"

Perhaps it would be clearer if these variables were named `insertionPoint` and `insertionLine`.

Src/VimCore/EditorUtil.fs
ricksladkey: Good idea.

PR #2118, opened 2018-05-08
Option backslash fixes

### LINKED ISSUE #2117: String settings cannot be set to a string containing a blank
The `:set <setting>=<value>` command does not work correctly if the value contains a blank.

Example:

`:set shell=c:\Program\ Files\Git\\bin\sh.exe`

produces a parse error. The rules for how vim parses what it calls "string option values" are fairly complicated, but at a minimum, `\<Space>` is needed to allow a space to be included.

`:help option-backslash`

in vim provides all the gory details.

ricksladkey:
Implement the rules documented by vim in `:help option-backslash`.

Fixes #2117


jaredpar:
I couldn't get help with `:help option-escape`. Is that help coming from a plugin? I'm running gVim 8.0.5 on this particular machine.

ricksladkey:
Oops, I don't know how many places I put that in incorrectly but the correct vim help name is `:help option-backslash`. I'm using vim 8.0 as well.

Src/VimCore/Interpreter_Parser.fs
```diff
@@ -1548,6 +1548,49 @@ type Parser
...
+                    let char = _tokenizer.CurrentChar
+                    if isFileName && CharUtil.IsFileNameChar char then
+                        builder.AppendChar '\\'
+                        builder.AppendChar char
+                    else
+                        builder.AppendChar char
+                    _tokenizer.MoveNextChar()
+
+            | char ->
+                if CharUtil.IsBlank char || "|".Contains(char.ToString()) then
```
jaredpar: Why the Contains vs. just `'|' = char`?

ricksladkey: This mimicked a longer list from the previous code:

```f#
CharUtil.IsLetterOrDigit c || @"-:\.,<>~[]".Contains(c.ToString())
```

So `Contains` here is a placeholder for any non-blank unescaped character that terminates the string.

According to `:help option-backslash`, the only unescaped characters that can terminate an option-backslash string are vertical bar (which I don't think we support elsewhere yet?) and double quote (which starts a comment).

So technically, double quote needs to be added to it, but I don't think anything else does.

PR #2121, opened 2018-05-10
Implement filter ('!') command and associated operations

### LINKED ISSUE #1540: Feature request: filter text through an external program using !
Recently installed GOW, a nice little directory of precompiled unix utilitites for windows (like awk, sed, sort, wc, etc.). Would be really cool to be able to send text from VsVim to one of those utilities, reading its output back into the file (ala VIM, as described in section 10.9 of its help (usr_11.txt)). 

E.g. do a visual select of some lines of text and sort them using:

:'<,'>!sort

In VsVim I get a "No range allowed" message.

### LINKED ISSUE #1987: After substitute in selection and destroying selection I'm in command mode (VS2013/VS2015)
Hi,

If I select text with (e.g.) with `"Shift+V"` and start a substitution `":'<,'>s/…/…/"`, the selection is active at the end.
"Canceling" the selection by hitting `"Esc"` opens the command prompt `":"` and I need an additional hit on escape to get back to the text.

Can you please test, if you can reproduce and fix it?

Best regards,

Martin

### LINKED ISSUE #2033: How do you read the content of the status bar
I have some example .bat file that just echos "Test". When I execute the file via '!example.bat' Then the stdout stuff gets printed into the status bar.

How can I copy that stdout into a new buffer I opened with split?

### LINKED ISSUE #2078: After replacing in visual mode and moving, <Esc> switches to command mode
I'm seeing an issue that took me quite a while to find a reproduction pattern for. I think it's actually two issues:

1. Select a multi-line region with `Shift-v` (two lines is enough)
```
foo test foo
foo test foo
foo test foo
```
2. Replace test with nothing: `:'<,'>s/test//` + enter  
  => *test* gets replace, but also the last line remains in Visual mode (first issue)
3. Press `k` once or twice to move up => more lines get selected
4. Hit `<Esc>` to remove selection => removes selection but also switches to `:` command mode (second issue)

### LINKED ISSUE #2120: Using a ':' command from visual mode doesn't return to normal mode
In vim if a `:` command is executed while in visual mode, when the command completes normally by pressing the enter key, the buffer returns to normal mode. In VsVim, the buffer returns first to visual mode with an altered visual selection (depending on the command executed). Then, if `<Esc>` is pressed, visual mode returns back to command mode. Finally, if `<Esc>` is pressed a second time, normal mode is finally entered. Upon hitting `<Enter>` from command mode entered from visual mode, it should just go straight to normal mode.

ricksladkey:
Changes:

- Implement the filter command `:[range]!command` to filter lines through external program
- Implement the filter lines operation `!!`
- Implement the filter motion operation `![motion]`
- Implement the filter operation from visual mode `V[motions]!`
- Enhance the `RunCommand` interface to return the exit code and standard error output
- Add timeout when running external programs to prevent hanging the editor
- Hide console window associated with shell/filter commands
- Fix `:` command from visual mode doesn't return to command mode
- Fix expansion of last command using `!`
- Fix double-escaping of `!` character in filter and shell commands
- Fix recording of previous shell/filter command

Issues:

- Fixes #1540
- Fixes #1987
- Fixes #2033
- Fixes #2078
- Fixes #2120

### DIFF
--- a/Src/VimCore/Modes_Command_CommandMode.fs
+++ b/Src/VimCore/Modes_Command_CommandMode.fs
@@ -27,6 +27,7 @@ type internal CommandMode
     let mutable _historySession: IHistorySession<int, int> option = None
     let mutable _bindData = BindDataError
     let mutable _keepSelection = false
+    let mutable _isPartialCommand = false
 
     /// Currently queued up command string
     member x.Command 
@@ -78,7 +79,10 @@ type internal CommandMode
 
             // It is possible for the execution of the command to change the mode (say :s.../c) 
             if _buffer.ModeKind = ModeKind.Command then
-                ProcessResult.Handled ModeSwitch.SwitchPreviousMode
+                if _isPartialCommand then
+                    ProcessResult.OfModeKind ModeKind.Normal
+                else
+                    ProcessResult.Handled ModeSwitch.SwitchPreviousMode
             else 
                 ProcessResult.Handled ModeSwitch.NoSwitch
         | BindResult.Cancelled ->
@@ -131,11 +135,12 @@ type internal CommandMode
         _historySession <- Some historySession
         _bindData <- historySession.CreateBindDataStorage().CreateBindData()
         _keepSelection <- false
+        _isPartialCommand <- false
 
         arg.CompleteAnyTransaction
         let commandText = 
             match arg with
-            | ModeArgument.PartialCommand command -> command
+            | ModeArgument.PartialCommand command -> _isPartialCommand <- true; command
             | _ -> StringUtil.Empty
 
         if not (StringUtil.IsNullOrEmpty commandText) then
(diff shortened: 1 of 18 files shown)


Src/VimCore/Modes_Command_CommandMode.fs
```diff
@@ -134,11 +135,12 @@ type internal CommandMode
         _historySession <- Some historySession
         _bindData <- historySession.CreateBindDataStorage().CreateBindData()
         _keepSelection <- false
+        _isPartialCommand <- false
```
jaredpar: Think this should be cleared in `OnLeave` as well.

ricksladkey:
Thanks, I added a new commit to address your `OnLeave` feedback.

This is failing because of the `Undo_InsertCharacterAboveCaret` test in the master branch for new 'endofline' management. I just submitted a tiny PR to fix that #2134. When that PR is merged, this one should pass checks.

PR #2125, opened 2018-05-11
Clean up line ranges

### LINKED ISSUE #1958: The substitution command shouldn't require the `.` when substituting from the current line.
Currently, substituting from the current line requires a `.`: `:.,5s/...`

But from my [experience](http://www.vim.org/download.php) (couldn't verify this [here](http://vim.wikia.com/wiki/Ranges)), you should be able to omit the `.`:

`:,5s/...` will perform the substitution from the current line to line 5.

### LINKED ISSUE #1976: Relative line number yanking from command line is not working.
Only absolute line number yanking (or any operation) is working from command line, doing `-10,-6co.`, visual studio shows error dialog saying invalid line number.

ricksladkey:
Changes:

- Fix default lhs and rhs for ',' and ';' line ranges
- Fix interpretation of `<line-specifier><number>`, e.g. `.2` after `,` or `;`
- Fix interpretation of `[line-specifier]-[number]` and `[line-specifier]+[number]`
- Add tests for a variety of line range variants

Fixes:

- Fixes #1958
- Fixes #1976

### DIFF
--- a/Src/VimCore/Interpreter_Parser.fs
+++ b/Src/VimCore/Interpreter_Parser.fs
@@ -1342,7 +1342,16 @@ type Parser
             _tokenizer.MoveNextToken()
             LineRangeSpecifier.EntireBuffer
         else
-            match x.ParseLineSpecifier() with
+            let startLine = x.ParseLineSpecifier()
+            let startLine =
+                match startLine with
+                | None ->
+                    if _tokenizer.CurrentChar = ',' || _tokenizer.CurrentChar = ';' then
+                        Some LineSpecifier.CurrentLine
+                    else
+                        None
+                | Some left -> startLine
+            match startLine with
             | None -> LineRangeSpecifier.None
             | Some left ->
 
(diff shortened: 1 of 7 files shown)


Src/VimCore/Interpreter_Parser.fs
```diff
@@ -1299,12 +1299,8 @@ type Parser
...
                     None
 
-            elif _tokenizer.CurrentChar = '+' then
-                _tokenizer.MoveNextToken()
-                x.ParseNumber() |> Option.map LineSpecifier.AdjustmentOnCurrent
-            elif _tokenizer.CurrentChar = '-' then
-                _tokenizer.MoveNextToken()
-                x.ParseNumber() |> Option.map (fun number -> LineSpecifier.AdjustmentOnCurrent -number)
+            elif _tokenizer.CurrentChar = '+' || _tokenizer.CurrentChar = '-' then
+                Some LineSpecifier.CurrentLine
```
jaredpar: Ah nice simplification. Does look like I was doing the work for the adjustment case twice here. Had to open it up locally to see where this was already being accounted for.

PR #2127, opened 2018-05-12
Record last visual selection before modifying text buffer

### LINKED ISSUE #2079: Reselecting after undo-ing visual block deletion selects incorrect range
Given the following text.

```
1foo test foo
2foo test foo
3foo test foo
4foo test foo
```
1. `Shift-v`, `j` to select the first two lines
2. `d` to delete them (cursor is not on *3foo* line)
3. `u` to undo (cursor is now on *1foo* line)
4. `gv` to re-establish previous selection

=> lines 3 & 4 are selected
Expected: lines 1 & 2 to be selected.

ricksladkey:
- Fixes #2079

An alternative solution to this bug is much more invasive: modify the `IBufferTrackingService` API so that it accepts an `ITextSnapshot` instead of an `ITextBuffer`. This would allow the creation of tracking points and spans for a snaphot prior to the current snapshot of the text buffer.  However it would require changing all callers.

### DIFF
--- a/Src/VimCore/Modes_Visual_VisualMode.fs
+++ b/Src/VimCore/Modes_Visual_VisualMode.fs
@@ -188,8 +188,8 @@ type internal VisualMode
         let lastVisualSelection = VisualSelection.CreateForSelection _textView _visualKind _globalSettings.SelectionKind _vimBufferData.LocalSettings.TabStop
 
         // We can't wait until after the command to record the last visual selection
-        // because if the command modified the text buffer, the snapshot will be newer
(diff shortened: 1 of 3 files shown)


jaredpar:
You are a machine lately with these PRs! Sorry I'm just now beginning to catch up with them. Been a busy few weeks.

ricksladkey:
No problem, I found some time to work on VsVim issues close to my heart and that led me to overcome the entry barrier. After that work, I wanted to help chip away at the bug list to begin to pay back my appreciation for the project!

PR #2132, opened 2018-05-14
Goto definition fixes

### LINKED ISSUE #2131: Go to definition operations leave cursor positioned at the end of an identifier
Commands like `gd`, `gD` and `<C-]>` work well, but they position the cursor after the identifier that was searched for, instead of on it, like `g*` does.

ricksladkey:
Changes:

- Fix clearing selection after F12 and `<C-,>` to only clear new selections
- Remove duplicate selection clearing for `<C-]>`
- When going to a definition, place the cursor at the start of the identifier
- Also clear selection for Ctrl-F12 with VS2017 (e.g. C# language service)

Fixes:

- Fixes #2131

### DIFF
--- a/Src/VsVimShared/ICommandTarget.cs
+++ b/Src/VsVimShared/ICommandTarget.cs
@@ -29,7 +29,7 @@ namespace Vim.VisualStudio
     {
         CommandStatus QueryStatus(EditCommand editCommand);
 
-        bool Exec(EditCommand editCommand, out Func<Func<int>, int> wrapper);
+        bool Exec(EditCommand editCommand, out Action preAction, out Action postAction);
     }
 
     internal interface ICommandTargetFactory
--- a/Src/VsVimShared/ICommandTarget.cs
+++ b/Src/VsVimShared/ICommandTarget.cs
@@ -29,7 +29,7 @@ namespace Vim.VisualStudio
     {
         CommandStatus QueryStatus(EditCommand editCommand);
 
-        bool Exec(EditCommand editCommand, out Action action);
+        bool Exec(EditCommand editCommand, out Func<Func<int>, int> wrapper);
     }
 
     internal interface ICommandTargetFactory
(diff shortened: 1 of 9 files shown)


Src/VsVimShared/ICommandTarget.cs
```diff
@@ -29,7 +29,7 @@ internal interface ICommandTarget
     {
         CommandStatus QueryStatus(EditCommand editCommand);
 
-        bool Exec(EditCommand editCommand, out Action action);
+        bool Exec(EditCommand editCommand, out Func<Func<int>, int> wrapper);
```
jaredpar: One of the intents of this design over IOleCommandTarget was essentially to take the concept of next out of the implementation. Essentially no need to call into the next implementation. Why do we need to add this back here?

ricksladkey: The `GoToDefinition` code needs to run an action *before* the command (to collect the list of text views that have pre-existing selections) and to run an action *after* the command (to clear any new selections). This can be solved by either:

1) Splitting the current action into two actions: a before hook and an after hook
2) Changing the API to a allow single wrapper that accepts an opaque function to call the target

I chose the second method because:

1) The before and after hooks for `GoToDefinition` have to communicate state between each other (namely the list of text views with selections)
2) It's a relatively easy general mechanism in the functional-style with considerable flexibility

The major advantages of the new approach is that the wrapper can:

1) Bypass the command entirely
2) Use the success or failure of the command to affect the hooked actions

Only the latter feature is currently being used.

It didn't occur to me that the before hook and after hook could communicate state by capturing a variable in a closure but I think that will work.

ricksladkey:
I converted my changes to maintain the feature that `IOleCommandTarget ` doesn't need to execute the corresponding command.

In doing so, I failed to perfectly restore some of the original code so it might look like I am gratuitously reformatting the old code is some places, but really there was rewriting and then some manual undoing.

PR #2136, opened 2018-05-16
Lazy control key fixes

### LINKED ISSUE #1905: Add support for alternate file switching using  CTRL-^ or :e#
http://stackoverflow.com/questions/19971023/how-to-go-back-to-previous-opened-file-in-vim

ricksladkey:
Like vim, allow special character control keys that are normally produced with control and shift to be produced only with the control key.

<H2>Changes</H2>

- Translate `<C-2>` on input to `<C-@>`
- Translate `<C-6>` on input to `<C-^>`
- Translate `<C-->` on input to `<C-_>`
- Translate `<C-/>` on input to `<C-?>`
- Add support for mapping `<C-\>`

<H2>Fixes</H2>

- Partially fixes #1905 (mapping lazy control characters)

### DIFF
--- a/Src/VimWpf/Implementation/Misc/AlternateKeyUtil.cs
+++ b/Src/VimWpf/Implementation/Misc/AlternateKeyUtil.cs
@@ -166,7 +166,8 @@ namespace Vim.UI.Wpf.Implementation.Misc
                 return true;
             }
 
-            if ((modifierKeys & ModifierKeys.Control) != 0 &&
+            if ((modifierKeys == ModifierKeys.Control ||
+                modifierKeys == (ModifierKeys.Control | ModifierKeys.Shift)) &&
                 s_wpfControlKeyToKeyInputMap.TryGetValue(key, out keyInput))
             {
                 return true;


jaredpar:
Why does <C-6> map to <C-^>? Is this behavior inherent to gVim?

I've observed that caps is essentially ignored for alpha keys when combined with C. I wasn't aware that applied to non-alpha keys as well.

ricksladkey:
Quoting vim docs `:help ctrl-^`:

```
			Mostly the ^ character is positioned on the 6 key,
			pressing CTRL and 6 then gets you what we call CTRL-^.
			But on some non-US keyboards CTRL-^ is produced in
			another way.
```

ricksladkey:
Empirically, I bound all of the non-letter control keys (like `<C-^>`) in vim and found that you could press them with just control, e.g. `<C-6>` vs. `<C-S-6>` It's true that `^` isn't necessarily above `6`, depending on the keyboard, but vim specifically list `<C-6>` as an alias for `<C-^>`.

PR #2138, opened 2018-05-16
Work on restore bug

jaredpar:
Explicitly excluding ValueTuple from our build until the restore issues
are resolved.

https://github.com/NuGet/Home/issues/6936


ricksladkey:
I discovered that if I disable NuGet restore in Visual Studio and restore from the command line, I can build and debug, so I have a workaround for now.

PR #2139, opened 2018-05-17
Edit alternate file and global marks

### LINKED ISSUE #1872: :marks doesn't always display global bookmarks
Microsoft Visual Studio Professional 2015
Version 14.0.23107.0 D14REL
VsVim 2.1.2.0
Marks created via "m-A", "m-D", etc. don't always appear in list of marks.
It looks like only the global marks for the current buffer are listed.
gVim lists all global marks, regardless of buffer.

For example, I have marks 'A', 'D', and 'T' set.
Jump to mark 'T': 'T
:marks
mark line  col file/text
 '     39    0 C:\work\scratch.txt
 T     38    0 C:\work\scratch.txt
 "      1    0 C:\work\scratch.txt
 ^     40    0 C:\work\scratch.txt
 .     39   43 C:\work\scratch.txt
 <      2    0 C:\work\scratch.txt
 >      3   27 C:\work\scratch.txt

Jump to mark 'A': 'A
:marks
mark line  col file/text
 '    424   28 C:\Users\Char...
 A    424   28 C:\Users\Char...
 "     69    0 C:\Users\Charl...

### LINKED ISSUE #1890: Maybe support the ctrl+6 for switching between buffers like vim?
That would be a very convinent way to switch to and back without spliting window i think.

### LINKED ISSUE #1905: Add support for alternate file switching using  CTRL-^ or :e#

### LINKED ISSUE #1972: Navigating to an uppercase mark placed in a closed file does not reopen file
VS 2017 15.3.0 Preview 3.0 [26621.2.d15rel]
VsVim 2.4.1.0

**Note:** In the repro below, marks behave as I would expect until after A.txt's tab is closed.

### Repro steps:

1. Create an uppercase mark `mA` in A.txt
2. Close A.txt's tab
3. Navigate to uppercase mark `'A` from another file.

### Expected behavior:

A.txt is reopened in a new tab, and the cursor is placed at the start of the marked line.

### Actual behavior:

Navigating to the the mark fails. VsVim shows a "Mark not set" status.

Even after A.txt is manually reopened, attempting to navigate to `'A` from another tab continues to fail with a "Mark not set" status. This was even true for tabs I successfully was able to do this from prior to closing A.txt.

### Other Strangeness

After A.txt is manually reopened, attempting to navigate to `'A` from A.txt's newly opened tab mostly succeeds. The weird part is that `'A` will now put the cursor the middle of the line precisely where the mark was set like I would expect `` `A`` to do instead of putting the cursor at the start of the line. `` `A`` continues to work as expected, but only from A.txt's tab.

### LINKED ISSUE #2191: Jumping exactly to a mark may leave caret in virtual space
Special marks like `>` are sometimes set at the very end of a non-empty line, past where vim would normally allow the caret to be set (unless for example `:set virtualedit=onemore` is set). Jumping to such a mark with `` `> `` will put the caret in virtual space.

ricksladkey:
Implement what vim calls "edit alternate file" (see `:help ctrl-^`). It also extends VsVim's handling of global marks to support unloaded buffers.

### Changes

- Add most recently used buffer list
- Implement `<C-^>` to switch to the nth most recently focused buffer
- Add file history
- Implement, for example, `:e#` and `:e#[count]` to edit recent files
- Add the `:files` command (with the aliases `:buffers` and `:ls`)
- Fix display of all global marks with `:marks` command
- Migrate global marks when buffers are unloaded
- Recreate global marks when buffers are reloaded
- Support navigating to a global mark in an unloaded buffer
- Fix cursor position when navigating to a mark in a another buffer
- Support setting initial cursor position when opening new documents
- Prevent putting caret in virtual space for end-of-line marks

### Issues

- Fixes #1890
- Fixes #1905
- Fixes #1872
- Fixes #1972
- Fixes #2191

### Discussion

This doesn't exactly replicate all vim semantics, but tries to keep the spirit of the commands in the context of Visual Studio. Specifically, `<C-^>` will never take you to a file that isn't open and `[count]<C-^>` will take you the the nth most recent focused buffer, not to a a specific buffer with an unchanging number from the `:files` list. Furthermore, `<C-^>` is not like `:e#` but more like `:tabe#` in that it doesn't replace the contents of the existing window with a new buffer. Nevertheless, `:e#` *can* take you to a file that is not open, but similarly, `:e#[count]` takes you to the nth most recent file.

The intent is that `<C-^>` and `:tabe#` do what you would expect them to do in Visual Studio and small counts might be useful to switch a little further back, but probably memorizing buffer numbers peculiar to the current editing session is not the best/fastest way to switch to a specific non-recent buffer in Visual Studio.

Global mark changes were included in the same PR because they both relate heavily to the files associated with text buffers.

### DIFF
--- a/Src/VimApp/VimAppHost.cs
+++ b/Src/VimApp/VimAppHost.cs
@@ -25,7 +25,7 @@ namespace VimApp
         private readonly IFileSystem _fileSystem;
         private readonly IDirectoryUtil _directoryUtil;
         private readonly IContentTypeRegistryService _contentTypeRegistryService;
-        private readonly Dictionary<ITextBuffer, string> _bufferMap;
+        private readonly Dictionary<IWpfTextView, string> _viewMap;
         private IVimWindowManager _vimWindowManager;
         private IVim _vim;
 
@@ -63,7 +63,7 @@ namespace VimApp
             _contentTypeRegistryService = contentTypeRegistryService;
             _fileSystem = fileSystem;
             _directoryUtil = directoryUtil;
-            _bufferMap = new Dictionary<ITextBuffer, string>();
+            _viewMap = new Dictionary<IWpfTextView, string>();
         }
 
         public override void VimCreated(IVim vim)
@@ -89,7 +89,14 @@ namespace VimApp
 
         public override string GetName(ITextBuffer textBuffer)
         {
-            return _bufferMap.TryGetValue(textBuffer, out string name) ? name : "";
+            foreach (var pair in _viewMap)
+            {
+                if (pair.Key.TextBuffer == textBuffer)
+                {
+                    return pair.Value;
+                }
+            }
+            return "";
         }
 
         public override int GetTabIndex(ITextView textView)
@@ -126,6 +133,16 @@ namespace VimApp
         // the active window, not existing
         public override bool LoadFileIntoExistingWindow(string filePath, ITextView textView)
         {
+            foreach (var pair in _viewMap)
+            {
+                if (pair.Value == filePath)
+                {
+                    var point = pair.Key.Caret.Position.VirtualBufferPosition;
+                    NavigateTo(point);
+                    return true;
+                }
+            }
+
             var vimWindow = MainWindow.ActiveVimWindowOpt;
             if (vimWindow == null)
             {
@@ -148,7 +165,7 @@ namespace VimApp
                     }),
                     DispatcherPriority.ApplicationIdle);
 
-                _bufferMap.Add(createdTextView.TextBuffer, filePath);
+                _viewMap.Add(createdTextView, filePath);
 
                 return true;
             }
@@ -161,11 +178,24 @@ namespace VimApp
 
         public override bool LoadFileIntoNewWindow(string filePath)
         {
+            foreach (var pair in _viewMap)
+            {
+                if (pair.Value == filePath)
+                {
+                    var point = pair.Key.Caret.Position.VirtualBufferPosition;
+                    NavigateTo(point);
+                    return true;
+                }
+            }
+
             try
             {
                 var textDocument = TextDocumentFactoryService.CreateAndLoadTextDocument(filePath, TextBufferFactoryService.TextContentType);
                 var wpfTextView = MainWindow.CreateTextView(textDocument.TextBuffer);
                 MainWindow.AddNewTab(System.IO.Path.GetFileName(filePath), wpfTextView);
+
+                _viewMap.Add(wpfTextView, filePath);
+
                 return true;
             }
             catch (Exception ex)
--- a/Src/VimApp/VimAppHost.cs
+++ b/Src/VimApp/VimAppHost.cs
@@ -10,6 +10,7 @@ using System.Windows.Input;
 using Microsoft.VisualStudio.Utilities;
 using Vim.UI.Wpf;
 using System.Windows.Threading;
+using System.Collections.Generic;
 
 namespace VimApp
 {
@@ -24,6 +25,7 @@ namespace VimApp
         private readonly IFileSystem _fileSystem;
         private readonly IDirectoryUtil _directoryUtil;
         private readonly IContentTypeRegistryService _contentTypeRegistryService;
+        private readonly Dictionary<ITextBuffer, string> _bufferMap;
         private IVimWindowManager _vimWindowManager;
         private IVim _vim;
 
@@ -61,6 +63,7 @@ namespace VimApp
             _contentTypeRegistryService = contentTypeRegistryService;
             _fileSystem = fileSystem;
             _directoryUtil = directoryUtil;
+            _bufferMap = new Dictionary<ITextBuffer, string>();
         }
 
         public override void VimCreated(IVim vim)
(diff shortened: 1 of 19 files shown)


ricksladkey:
I have another change for marks that depends on this one. It adds mark glyphs that appear in the indicator column. I'll submit that one if this PR gets merged.

ricksladkey:
I'm going to post a teaser screenshot for mark glyphs in the hope that it will increase the chances that this PR gets merged!

image

ricksladkey:
@jaredpar This pull request has been around for quite a while, and of course you should review it in your own time frame, but I'm just wondering whether:

- you don't like it from some reason,
- it's big enough that it requires an in depth review, or
- somehow it's not on your radar or otherwise didn't notice it

I have another PR that depends on this one, so I am hopeful that this one will eventually get merged!

I'm happy to wait as long as not reviewing or merging it is intentional, and not because this PR is too far down in the pile.  Thanks.

jaredpar:
Sorry. This PR just requires me to page in a lot of insert mode to review correctly. I've sat down to review this a few times now and ended up getting distracted and losing my train of thought. I will try and get some time to review this over the weekend.

Src/VimApp/VimAppHost.cs
```diff
@@ -69,6 +72,25 @@ public override void VimCreated(IVim vim)
...
+                {
+                    if (viewInfo.TextView == textView)
+                    {
+                        viewInfo.VimBuffer.Close();
+                        viewInfo.TextView.Close();
+                        _viewMap.Remove(viewInfo.TextView);
+                        break;
+                    }
+                }
+                vimWindow.Clear();
```
jaredpar: Why do we need to clear the windows here?

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4215,7 +4230,7 @@ type IVimHost =
     abstract LoadFileIntoExistingWindow: filePath: string -> textView: ITextView -> bool
 
     /// Loads the new file into a new existing window
-    abstract LoadFileIntoNewWindow: filePath: string -> bool
+    abstract LoadFileIntoNewWindow: filePath: string -> line: int -> column: int -> bool
```
jaredpar: Did you consider using an option here vs. using -1 as the "default" value?

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4491,17 +4509,23 @@ and IMarkMap =
     /// Get the mark for the given char for the IVimTextBuffer
     abstract GetMark: mark: Mark -> vimBufferData: IVimBufferData -> VirtualSnapshotPoint option
 
+    /// Get the mark info for the given char for the IVimTextBuffer
+    abstract GetMarkInfo: mark: Mark -> vimBufferData: IVimBufferData -> (char * string * int * int) option
```
jaredpar: Should we add a small record here vs returning a 4 way tuple?

jaredpar:
Sorry again for taking so long on this. Left some minor feedback / Qs.

Src/VimCore/CoreInterfaces.fs
ricksladkey: I did! I was ambivalent because using `Option` is awkward from C# and using `Nullable<T>` is awkward from F#. And of course `-1` is a "magic number", but I think of it as similar to `String.IndexOf` returning -1, which is a time-honored tradition in many programming languages.

In effect, there are three functionally-equivalent ways of doing the same thing and I chose the language-neutral way. I'm happy to switch to another method but I couldn't deduce which one would be preferred by myself.

ricksladkey:
No apology necessary! I will try to address your review feedback.

Src/VimApp/VimAppHost.cs
ricksladkey: I was using VimApp to test things like `:e #`, :e file.txt`, `:tabe file.txt`, `:q`,  and especially navigating to marks in a closed buffer. A lot of the `VimHost` was unimplemented and I tried to fix it up so most of those things work like they work in Visual Studio.

I admit I wasn't too worried about getting it just perfect because:

1) It didn't used to work at all
2) VimApp is mostly just used by developers

It is intriguing to think of really investing in VimApp enough to make it competitive with gvim for non-Visual Studio editing!

Src/VimCore/CoreInterfaces.fs
jaredpar: Yeah I've been frustrated with finding a good way to have default values that work equally well in C# and F#. I experimented a bit with F# optional parameters but they fall a bit short as they don't appear optional in C#. To use them I ended having to define two overloads:

1. One with all the parameters marked as optional
1. One with all all the optional parameters omitted that just forwards to the other 

The (2) version is callable from C#. Still though it's less than ideal 

https://github.com/jaredpar/VsVim/blob/master/Src/VimCore/CoreInterfaces.fs?utf8=%E2%9C%93#L870

I've been working on a proposal to get F# to emit their optional parameters in a way that would make it much easier to use from C#. But that's a ways away.

Src/VimApp/VimAppHost.cs
ricksladkey: I reverted my vim app changes for the most part, since my implementation was not fully thought out.

Src/VimCore/CoreInterfaces.fs
ricksladkey: I converted the two individual parameters to use `FSharpOption` which seems no worse than the `-1` approach!

Src/VimCore/CoreInterfaces.fs
ricksladkey: I created a `MarkInfo` type which `GetMarkInfo` now returns.

PR #2141, opened 2018-05-18
Repeat command

### LINKED ISSUE #2140: @: command does not work
In vim after performing an ex command I am able to repeat it using "@:"
I love vsvim and idk how many others use this, but could be nice.

ricksladkey:
Implement the vim `:@:` command to repeat commands (see `:help :@:` in vim).
Changes:

- Add 'last line command' command
- Add capability to merge new line range with previous line command
- Implement `[range]:@:` to repeat last command optionally with new line range

Fixes:

- Fixes #2140


ricksladkey:
@gwojan In vim documentation and VsVim discussions, what vim calls an "ex command", i.e. something you can type at the colon prompt, is usually notated with a colon to distinguish it from commands/keystrokes applicable in normal mode.  So we refer to the `e` ex command as `:e` but the `.` command to repeat as just `.`. Since `@:` is a command you type at the colon prompt, it becomes `:@:`.

ricksladkey:
@gwojan Actually, you are right, `@:` is indeed a normal command that does what `:@:` would do, but from normal mode.  I'll add that feature, which should be pretty straightforward.

Src/VimCore/Vim.fs
```diff
@@ -154,6 +155,9 @@ type internal VimData(_globalSettings: IVimGlobalSettings) as this =
         member x.LastSubstituteData 
             with get() = _lastSubstituteData
             and set value = _lastSubstituteData <- value
+        member x.LastLineCommand
```
jaredpar: Why can't `LastCommandLine` be used here vs. adding the new property?

jaredpar: Because they're different types ... Need to finish coffee before I review code. :( 

---
In reply to: [189437792](https://github.com/jaredpar/VsVim/pull/2141#discussion_r189437792) [](ancestors = 189437792)

Src/VimCore/Modes_Command_CommandMode.fs
```diff
@@ -50,6 +50,7 @@ type internal CommandMode
                 command
 
         let lineCommand = _parser.ParseLineCommand command 
+        _vimData.LastLineCommand <- Some lineCommand
```
jaredpar: Why set this here vs. line 60 where `LastCommandLine` is set. This would mean that within operations like `RunLineCommand` the values of `LastCommandLine` and `LastLineCommand` represented different commands.

jaredpar: This will store `LineCommand.ParseError` in case of syntax errors. Is that intended? 

---
In reply to: [189437911](https://github.com/jaredpar/VsVim/pull/2141#discussion_r189437911) [](ancestors = 189437911)

Src/VimCore/Vim.fs
ricksladkey: The main reason we cannot use a string as the last command is that we need to support repeating the repeat. For example, `:s/pattern/replacement/`, then `:@:`, and then `:@:` again. For the second repeat, LastCommandLine is `":@:"`, but we want to repeat the substitute.

A secondary reason is that we need to support merging a new line range with a previous line command. If the last command line is a string, we would have to partially reparse it.

Src/VimCore/Modes_Command_CommandMode.fs
jaredpar: Gotcha on syntax error. 

---
In reply to: [189438568](https://github.com/jaredpar/VsVim/pull/2141#discussion_r189438568) [](ancestors = 189438568)

PR #2144, opened 2018-05-19
Remove last of VS2010 code

jaredpar:
VsVim stopped supporting VS2010 a few releases ago. This removes the
last of the code that helped support that release.


jaredpar:
CC @ricksladkey

Issue #2113, opened 2018-05-05
Cursor goes to phantom line


ricksladkey:
For buffers that contain a final newline, the cursor is allowed to move to from the last character of the last line to the phantom line in the following situations:

Normal Mode:

- Down arrow key (`<Down>`)
- Right arrow key (`<Right>`)
- Down motion (`j`)
- Right motion when `whichwrap` is set (`l`, `<Space>`)
- Word motions (`w`, `W`, `e`, `E`)
- Section forward (`]]`)
- Goto line with a line number greater than the number of lines in the buffer (`[count]G`)

Insert Mode:

- Right arrow key (`<Right>`)
- Word motion (`<C-Right>`)
 
In all these cases, in vim the cursor does not move.

Example buffer that exhibits the problems:

```
cat\r\n
dog\r\n
```

PR #2145, opened 2018-05-19
Paging and scroll fixes

### LINKED ISSUE #1093: dH doesn't work like dL
dL deletes from cursor to last line of window. dH should delete from cursor to first line in window, but deletes ONLY the first line.

Love the extension BTW! Used to pay $100 for a commercial product that wasn't even as good.

### LINKED ISSUE #1888: H/L in combination with `set scrolloff=2`
When using `set scrolloff=2`, H will jump to the top of the screen instead of top+2, thus causing the window to scroll. L has the same issue.

E.g. when the top line is 20 and I press H, the top line is now 18 with the cursor placed on 20.

ricksladkey:
### Changes
- Prevent `<C-f>` from reaching the phantom line
- Prevent `<C-d>` from reaching the phantom line
- Prevent display of the phantom line when going to the end of the file with `G`
- Systematically use the `GetTextViewLines` API to avoid exception handling boilerplate
- Ensure screen motions (`H` and `L`) only move to fully visible lines
- Fix screen motions (`H` and `L`) combined with scrolloff
- Handle scrolling by line (`<C-e>` and `<C-y>`) when line wrapping is in effect
- Fix scrolling by line can position the cursor partially offscreen
- Increase coverage of paging and scrolling scrolling tests

### Fixes
- Fixes additional cases not mentioned in the closed issue #2113
- Fixes #1888
- Closes #1093 (already fixed)

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -2833,7 +2833,7 @@ type internal CommandUtil
 
         // In case something went wrong when calculating the row count,
         // it should always be at least at big as count.
-        let rowCount = min rowCount count
+        let rowCount = max rowCount count
 
         _textView.ViewScroller.ScrollViewportVerticallyByLines(direction, rowCount)
 


Src/VimCore/CommandUtil.fs
```diff
@@ -1779,23 +1779,20 @@ type internal CommandUtil
     /// Happens when the middle mouse button is clicked.  Need to paste the contents of the default
     /// register at the current position
     member x.PutAfterCaretMouse() =
-        try
+        match TextViewUtil.GetTextViewLines _textView with
```
jaredpar: Consider combining the match clauses.

Src/VimCore/EditorUtil.fs
```diff
@@ -1818,13 +1818,19 @@ module TextViewUtil =
...
     let GetTextViewLines (textView: ITextView) =
-        try
-            let textViewLines = textView.TextViewLines
-            if textViewLines <> null then Some textViewLines
-            else None
-        with 
+        if textView.IsClosed || textView.InLayout then
+
             // TextViewLines can throw if the view is being laid out.  Highly unlikely we'd hit
             // that inside of Vim but need to be careful
```
jaredpar: It's actually likely during startup that we end up executing VsVim commands during a layout. Over the years I've had a number of bug reports in this area. Essentially every startup sequence for different file types is a bit of a race condition. As such the editor ends up firing off events while it's still in the middle of a layout. Once we're past startup though it's pretty unlikely.

Src/VimCore/CommandUtil.fs
ricksladkey: Gotcha, we can match on a tuple because we need both of them to be some value.

Src/VimCore/EditorUtil.fs
ricksladkey: Unfortunately I can think of no easy way to force F# clients to use our bulletproof API. At least when trying to do something in VsVim, people like me who search around for something similar to build off of will now only find uses of the preferred API.

PR #2150, opened 2018-05-20
Fix accidental use of hard coded register 'c' vs. variable c

### LINKED ISSUE #2068: In command mode, ctrl-r, 0 register is not working
I searched related to issues. 
I found this bug was already fixed #1252 .

But when I test ctrl-r, 0 in command mode, it is still not working.
Which option do I need to change , or else what am I wrong?

I checked :register and the result is as below:.
--- Registers ---
"+   asdf
"0   auto-style1

I tested in Visual Studio 2013 and vsvim v2.5.

ricksladkey:
- Fixes #2068

### DIFF
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
@@ -847,7 +847,7 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
                     // Now we need to update the command line.  During edits the controller is responsible
                     // for manually updating the command line state.  Also we have to keep the caret postion
                     // correct
-                    var name = RegisterName.OfChar('c');
+                    var name = RegisterName.OfChar(c);
                     if (name.IsSome())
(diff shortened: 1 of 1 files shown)


Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -847,7 +847,7 @@ private void UpdateForPasteWait(WpfTextChangedEventArgs e)
                     // Now we need to update the command line.  During edits the controller is responsible
                     // for manually updating the command line state.  Also we have to keep the caret postion
                     // correct
-                    var name = RegisterName.OfChar('c');
+                    var name = RegisterName.OfChar(c);
```
jaredpar: I checked just to be sure and yep this is my fault 😦

ricksladkey: Ha, ha, I bet you were manually testing with the `c` register!

PR #2153, opened 2018-05-21
Fix screen motions combined with scrolloff

### LINKED ISSUE #1093: dH doesn't work like dL

### LINKED ISSUE #1888: H/L in combination with `set scrolloff=2`

ricksladkey:
### Changes
- Fix screen motions combined with scrolloff
### Fixes
- Fixes #1888
- Closes #1093

### DIFF
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -2324,8 +2324,15 @@ type internal MotionUtil
             if range.Count = 0 then
                 None
             else
+
+                // The minimum distance from the top of the window is zero if
+                // the top of the buffer is visible and scrolloff otherwise.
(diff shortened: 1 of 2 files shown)


ricksladkey:
I merged this PR with #2145 so they wouldn't conflict with each other and because they cover related topics.

PR #2154, opened 2018-05-21
Support escaped key notation in string expression constants

### LINKED ISSUE #1845: Escaped sequences don't work for "execute"
image

generates

image

rather than actually executing carriage returns and escape key presses.

ricksladkey:
- Fixes #1845

### DIFF
--- a/Src/VimCore/Interpreter_Parser.fs
+++ b/Src/VimCore/Interpreter_Parser.fs
@@ -1490,12 +1490,31 @@ type Parser
                 moveNextChar()
                 if afterEscape then
                     match c with
+                    | 'e' -> builder.AppendChar (char 0x1b)
                     | 't' -> builder.AppendChar '\t'
                     | 'b' -> builder.AppendChar '\b'
                     | 'f' -> builder.AppendChar '\f'
                     | 'n' -> builder.AppendChar '\n'
                     | 'r' -> builder.AppendChar '\r'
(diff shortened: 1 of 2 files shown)


Src/VimCore/Interpreter_Parser.fs
```diff
@@ -1490,12 +1490,31 @@ type Parser
                 moveNextChar()
                 if afterEscape then
                     match c with
+                    | 'e' -> builder.AppendChar (char 0x1b)
                     | 't' -> builder.AppendChar '\t'
                     | 'b' -> builder.AppendChar '\b'
                     | 'f' -> builder.AppendChar '\f'
                     | 'n' -> builder.AppendChar '\n'
                     | 'r' -> builder.AppendChar '\r'
-                    | '\\' -> builder.AppendChar '\\'
```
jaredpar: Why did this case get removed? Seems to be valid. Tried the following in gVim `:echo "f\\o"` and it printed `f\o`

ricksladkey: Because `\[char]` where `[char]` is not special already just appends that character. In other words, `\\` would already append `\` (by falling through to the wildcard match) so this line, while not harmful, is redundant.

In summary, `\` after `\` is not special, it's just another escaped character.

ricksladkey: I added a commit to specifically test that backslash escapes itself.

PR #2155, opened 2018-05-21
Implement v and V for operator motions

### LINKED ISSUE #1800: `dvb` does not delete
In version 2.1.1, `dvb` moves the cursor as if `b` were pressed, but it does not actually delete anything.

jaredpar:
Operators can use v and V to modify motions in the following way:

- v: force the motion to be characterwise 
- V: force the motion to be linewise

Full details available at `:help o_v` 

closes #1800


Src/VimCore/CoreInterfaces.fs
```diff
@@ -3413,25 +3468,25 @@ type MotionBinding =
     /// Simple motion which comprises of a single KeyInput and a function which given 
     /// a start point and count will produce the motion.  None is returned in the 
     /// case the motion is not valid
-    | Simple of KeyInputSet * MotionFlags * Motion
+    | Static of KeyInputSet * MotionFlags * Motion
 
     /// Complex motion commands take more than one KeyInput to complete.  For example 
     /// the f,t,F and T commands all require at least one additional input.  The bool
     /// in the middle of the tuple indicates whether or not the motion can be 
     /// used as a cursor movement operation
```
ricksladkey: These comments are slightly out of date, e.g. simple/complex vs. static/dynamic and referring to a obsolete bool.

Test/VimCoreTest/NormalModeIntegrationTest.cs
```diff
@@ -1712,6 +1712,124 @@ public void VisualLines()
...
+
+            [WpfFact]
+            public void CharacterWiseInclusive()
+            {
+                Create("the", "dog");
+                _vimBuffer.ProcessNotation("yVfh");
+                Assert.Equal("the" + Environment.NewLine, UnnamedRegister.StringValue);
+            }
+        }
+
```
ricksladkey: I ran all these test by hand in vim and they passed.

Src/VimCore/CoreInterfaces.fs
jaredpar: Thanks. Fixed.

PR #2157, opened 2018-05-22
Backspace in replace mode

### LINKED ISSUE #653: Backspace should undo in Replace Mode
The backspace command needs to do an undo in replace mode.

ricksladkey:
When backspacing in replace mode the editor should effectively "undo" any previous replacements. This is one of the oldest issues, filed in 2011!

- Fixes #653

### DIFF
--- a/Src/VimCore/InsertUtil.fs
+++ b/Src/VimCore/InsertUtil.fs
@@ -604,6 +604,7 @@ type internal InsertUtil
             | InsertCommand.Overwrite s -> x.Replace s
             | InsertCommand.ShiftLineLeft -> x.ShiftLineLeft ()
             | InsertCommand.ShiftLineRight -> x.ShiftLineRight ()
+            | InsertCommand.UndoReplace -> x.UndoReplace ()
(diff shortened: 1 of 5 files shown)


Src/VimCore/InsertUtil.fs
```diff
@@ -41,6 +41,7 @@ type internal InsertUtil
     let _vimHost = _vimBufferData.Vim.VimHost
     let _vimTextBuffer = _vimBufferData.VimTextBuffer
     let mutable _combinedEditStartPoint: ITrackingPoint option = None
+    let mutable _replaceUndoLine: ITextSnapshotLine option = None
```
jaredpar: Did you consider using `ITrackingLineColumn` here? The goal here seems to be to track a line number across edits to the buffer. That is a bit better done with `ITrackingLineColumn`.

ricksladkey: This is literally a snapshot of the the old line before any insert edits were made on it. A tracking line column would not be able to provide us with that historical information.

jaredpar:
I realize I'm behind in reviewing this one too 😦 Will get this done this weekend.

PR #2163, opened 2018-05-26
Blocks and token matching

### LINKED ISSUE #2159: ci( yi( di( etc does not work if parenthesis contains a string literal with a "("
Trying to use vi( or any _i( combination inside a parenthesis containing a "(" inside a string literal does not work, for example trying to copy the statement in the if () does not work:

`if ("hello".IndexOf('(') == 0)`

ricksladkey:
### Changes

- Convert `i(` (paren block motion), etc. and `%` (token matching) to use the same infrastructure
- Fix paren blocks, etc. to handle string and character literals
- Add tests for block and token matching corner cases

### Fixes

- Fixes #2159

### Discussion

There is a long comment in `GetBlock` discussing how vim behaves and how VsVim handles the problem. The idea is not to behave exactly as vim does, because it is not hard to trip up vim either, but to handle the most common cases that previously tripped up VsVim.

### DIFF
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -428,8 +428,8 @@ type internal BlockUtil() =
         // Compute a tuple of three quantities:
         // - Whether the target point is inside a string literal
         // - The start point of the string, as long as it there
-        //   is no unbalanced start character between the beginning
-        //   and the target
+        //   is no unmatched start character in string literals
+        //   before the target
         // - The quote character used to delineate the string
         let isInStringLiteral (target: SnapshotPoint) (sequence: SnapshotPoint seq) =
             let mutable escape = false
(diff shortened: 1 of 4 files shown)


Src/VimCore/MotionUtil.fs
```diff
@@ -348,6 +348,210 @@ type MatchingTokenKind =
...
+        // Is the char at the given point escaped?
+        let isEscaped point =
+            match SnapshotPointUtil.TrySubtractOne point with
+            | None -> false
+            | Some point -> SnapshotPointUtil.GetChar point = '\\'
+
+        // Is the char at the given point double escaped?
+        let isDoubleEscaped point =
+            match SnapshotPointUtil.TrySubtract point 2, SnapshotPointUtil.TrySubtract point 2 with
+            | None, _
```
jaredpar: Huh, didn't realize F# had this type of fall through behavior.

ricksladkey: The underscore matches a tuple. In any case, I didn't change that code; I just had to move it to a new type for reuse. So you probably wrote the original code!

PR #2166, opened 2018-05-30
Surrogate pairs

### LINKED ISSUE #1786: Cannot navigate right over unicode characters occupying planes higher than the BMP plane (0)
I like putting emojis in unit tests instead of data. I noticed the other day I could not navigate (`l` in normal mode, right arrow in all modes) over some emojis, but if I was in normal mode and used the chord `2l` I would skip to the place I expected `l` or right arrow to take me. After examining the hex dump of files with various character in the same spot, ones that would let me over them and ones that would not, I determined that it occurs when using a unicode character that is 4 bytes in length. Doing some research I noted this is precisely the set of **utf8mb4** encoded characters in the planes above BMP.

ricksladkey:
### Changes

- Expand `GetCharacterSpan` to handle surrogate pairs
- Add `SnapshotCharacterSpan` to operate on logical columns relying on `GetCharacterSpan`
- Modify 'get next/previous point on line' to use snapshot character spans
- Modify 'get next/previous point with wrap' to use character spans
- Modify normal/insesrt mode motions to avoid directly incrementing/decrementing buffer positions
- Fix `a` (insert after) to append after the current character span
- Fix inclusive/exclusive conversion not to directly increment/decrement the span length
- Add snapshot character span and normal mode integration tests for surrogate pairs

### Fixes

- Fixes #1786

### Discussion

This is not a "quick fix" that simply allows cursor motions to navigate over emojis. It is a deeper refactoring to enable all motions and operators to work correctly in the presence of surrogate pairs. As a result, for example, we no longer rely on "caret affinity" to correctly adjust the cursor if we happen to land in the middle of a surrogate pair.

However, there are still many areas that may need that need more work to fully implement support for surrogate pairs. For example, almost every use of `SnapshotPointUtil.GetChar` probably needs to be changed to return the string representing the character at point, or to upgrade from a 16-bit `char` type to a 32-bit `int`.

### DIFF
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -409,8 +409,7 @@ type SnapshotCharacterSpan =
 
     /// The column number of the column
     /// (Warning: don't use the column number as a buffer position offset)
-    /// TODO: Rename this to "ColumnNumber" analogous to LineNumber.
-    member x.Column = x._columnNumber
+    member x.ColumnNumber = x._columnNumber
 
     /// The position or text buffer offset of the column
     member x.Position = x._positions.[x._columnNumber]
@@ -487,7 +486,7 @@ type SnapshotCharacterSpan =
 
     /// Debugger display
     override x.ToString() =
-        sprintf "Point: %s Line: %d Column: %d" (x.Point.ToString()) x.LineNumber x.Column
+        sprintf "Point: %s Line: %d Column: %d" (x.Point.ToString()) x.LineNumber x.ColumnNumber
 
 /// The Text Editor interfaces only have granularity down to the character in the 
 /// ITextBuffer.  However Vim needs to go a bit deeper in certain scenarios like 
@@ -1726,7 +1725,7 @@ module SnapshotPointUtil =
     /// None will be returned (note this handles surrogate pairs)
     let TryGetPreviousCharacterSpanOnLine (point: SnapshotPoint) count =
         let column = SnapshotCharacterSpan(point)
-        if column.Column >= count then
+        if column.ColumnNumber >= count then
             let previousColumn = column.Subtract count
             Some previousColumn.Point
         else
@@ -1737,7 +1736,7 @@ module SnapshotPointUtil =
     /// surrogate pairs)
     let TryGetNextCharacterSpanOnLine (point: SnapshotPoint) count =
         let column = SnapshotCharacterSpan(point)
-        if column.Column + count <= column.ColumnCount then
+        if column.ColumnNumber + count <= column.ColumnCount then
             let nextColumn = column.Add count
             Some nextColumn.Point
         else
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -1284,35 +1284,60 @@ module SnapshotPointUtil =
         line.ExtentIncludingLineBreak
 
     // Get the span of the character which is pointed to by the point.  Normally this is a 
-    // trivial operation.  The only difficulty if the Point exists on an empty line.  In that
-    // case it is the extent of the line
+    // trivial operation.  The only difficulty if the Point exists at the end of a line
+    // (in which case the span covers the linebreak) or if the character is part of a
+    // surrogate pair.
     let GetCharacterSpan point = 
+        let snapshot = GetSnapshot point
         let line = GetContainingLine point
         let endSpan = new SnapshotSpan(line.End, line.EndIncludingLineBreak)
-        match endSpan.Contains(point) with
-            | true -> endSpan
-            | false -> new SnapshotSpan(point,1)
+        if endSpan.Contains(point) then
+            endSpan
+        elif point.Position = snapshot.Length then
+            new SnapshotSpan(point, 0)
+        else
+            let c = point.GetChar()
+            let snapshot = point.Snapshot
+            if System.Char.IsHighSurrogate(c) then
+                if point.Position = snapshot.Length then
+                    new SnapshotSpan(point, 1)
+                elif System.Char.IsLowSurrogate(point.Add(1).GetChar()) then
+                    new SnapshotSpan(point, 2)
+                else
+                    new SnapshotSpan(point, 1)
+            elif System.Char.IsLowSurrogate(c) then
+                if point.Position = 0 then
+                    new SnapshotSpan(point, 1)
+                elif System.Char.IsHighSurrogate(point.Subtract(1).GetChar()) then
+                    new SnapshotSpan(point.Subtract(1), 2)
+                else
+                    new SnapshotSpan(point, 1)
+            else
+                new SnapshotSpan(point, 1)
 
     /// Get the next point in the buffer with wrap
-    let GetNextPointWithWrap point = 
-        let tss = GetSnapshot point
-        let line = GetContainingLine point
-        if point.Position >= line.End.Position then
-            let num = line.LineNumber+1
-            if num = tss.LineCount then SnapshotUtil.GetStartPoint tss
-            else tss.GetLineFromLineNumber(num).Start
+    let GetNextPointWithWrap point =
+        let snapshot = GetSnapshot point
+        let nextPoint =
+            GetCharacterSpan point
+            |> SnapshotSpanUtil.GetEndPoint
+        if nextPoint.Position = snapshot.Length then
+            SnapshotUtil.GetStartPoint snapshot
         else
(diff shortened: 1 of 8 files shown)


Src/VimCore/EditorUtil.fs
```diff
@@ -254,49 +296,140 @@ type SnapshotLineRange  =
...
+    member x.Width = x._positions.[x._columnNumber + 1] - x.Position
+
+    /// The snapshot span covering the logical character at the column
+    member x.Span = new SnapshotSpan(x.Snapshot, x.Position, x.Width)
+
+    // The number of columns in the line containing the column
+    member x.ColumnCount = x._positions.Length - 1
+
+    /// Whether the "character" at column is a linebreak
+    member x.IsLineBreak = x.Position = x._line.End.Position
```
jaredpar: Why this vs. IsInsideLineBreak?

Src/VimCore/EditorUtil.fs
```diff
@@ -254,49 +296,140 @@ type SnapshotLineRange  =
...
-        let column = _column + count
-        if column > 0 && column < _snapshotLine.LengthIncludingLineBreak then
-            SnapshotColumn(_snapshotLine, _column + count)
-        else
-            let point = x.Point.Add count
-            SnapshotColumn(point)
+    /// Constructor for a column corresponding to a snapshot point
+    /// that "parses" the whole containing line and builds a lookup
+    // table of physical positions corresponding to logical columns
+    new (point: SnapshotPoint) =
```
jaredpar: Part of the motivation for the old constructor taking an `ITextSnapshotLine` was performance. Every call to `point.GetContainingLine` produces a new instance of `ITextSnapshotLine`. Using a `SnapshotPoint` based ctor meant that iterating every point in an `ITextSnapshot` allocated a new `ITextSnapshotLine` instance. That was actually a problem and traces lead me to the current form of taing an `ITextSnapshotLine`. Forced the caller to be a bit smarter about when the allocation should occur.

Src/VimCore/EditorUtil.fs
```diff
@@ -21,18 +21,60 @@ module internal EditorCoreUtil =
     let IsEndPoint (point: SnapshotPoint) = 
         point.Position = point.Snapshot.Length
 
+    /// TODO: Should this handle surrogate pairs and linebreaks?
```
jaredpar: Definitely need the variants which iterate points forward without surrogate pairs context. If we need a suggorate pair move forward that should probably be a different method.

Src/VimCore/EditorUtil.fs
```diff
@@ -21,18 +21,60 @@ module internal EditorCoreUtil =
...
+    /// Get the span of the character which is pointed to by the point.  Normally this is a 
+    /// trivial operation.  The only difficulty is if the point exists at the end of a line
+    /// (in which case the span covers the linebreak) or if the character is part of a
+    /// surrogate pair.
+    let GetCharacterSpan (line: ITextSnapshotLine) (point: SnapshotPoint) = 
+
+        // We require that the point belong to the line.
+        Contract.Requires(point.Snapshot = line.Snapshot)
+        Contract.Requires(point.Position >= line.Start.Position)
+        Contract.Requires(point.Position <= line.EndIncludingLineBreak.Position)
```
jaredpar: Can't we change the input here to a `SnapshotColumn` and avoid all of these asserts?

Src/VimCore/EditorUtil.fs
```diff
@@ -254,49 +296,140 @@ type SnapshotLineRange  =
 /// This is the representation of a point within a particular line.  It's common
 /// to represent a column in vim and using a SnapshotPoint isn't always the best
 /// representation.  Finding the containing ITextSnapshotLine for a given 
-/// SnapshotPoint is an allocating operation and often shows up as a critical 
-/// metric in profiling.  This structure pairs the two together in a type safe fashion
+/// SnapshotPoint is an allocating operation and often shows up as a critical
+/// metric in profiling.  This structure pairs the two together in a type safe fashion.
+///
+/// Conceptually, a "column" corresponds to a single logical character:
```
jaredpar: I'm concerned about merging the functionality here. I think of SnapshotColumn as a core primitive like SnapshotPoint. If there is a version for surrogate pairs I would expect that to be a separate type as well. One which pretty easily converts back and forth.

Src/VimCore/EditorUtil.fs
ricksladkey: With the new snapshot column, a linebreak is a single column position, and so you are either at the linebreak or you aren't. In other words, it's impossible for the snapshot column to be positioned in the middle of a linebreak (or in the middle of a surrogate pair, for that matter).

Src/VimCore/EditorUtil.fs
ricksladkey: My logic behind adapting `SnapshotColumn` is that every single caller is already expecting to operate on logical characters. If we were to create a new type, there would be no callers using `SnapshotColumn`.

To put it another way, any user of the old `SnapshotColumn` needs to handle surrogate pairs itself, which is 1) hard, and 2) messy (there can be many surrogate pairs on a line), and 3) inefficient. It's inefficient because to know the logical columns on a line you need to parse the text of the whole line. If we ever have to do that, it's better to do it once when we create the column.

This is the beauty of `SnapshotColumn` : VsVim already has it as an efficient abstraction over the raw text. By adapting `SnapshotColumn` to allow variable-width columns, all of VsVim suddenly becomes robust for documents containing surrogate pairs.

Src/VimCore/EditorUtil.fs
ricksladkey: The new API retains all of the performance advantages of the old API. The old API also had a constructor that just supplies a snapshot point. However, the new API has a constructor that takes a `SnapshotColumn` *instead* of an `ITextSnapshotLine` because not only do we need to carry over the snapshot line (like the old API), we also need to carry over the computed positions table. The positions table is an array and we want to avoid its allocation as well.

But believe me, I was extremely careful to avoid introducing any unnecessary inefficiency. Unfortunately, because our text buffer is effectively encoded, it is impossible to make a surgical change and have all of VsVim "just work."

Src/VimCore/EditorUtil.fs
ricksladkey: I suspected as much, and didn't risk attempting to change it. But can you give me any practical example of a situation where VsVim needs to be able to have a snapshot point pointing to the second character of a surrogate pair? I could not think of any.

Src/VimCore/EditorUtil.fs
ricksladkey: I don't know that I will convince you, and if not, you can at least use all the tests.

But, I will try one more line of defense.  How many times in VsVim have you used `SnapshotPointUtil.AddOne` or `SubtractOne` and ended up in the middle of a linebreak and had to special case the code? The middle of a line break and the middle of a surrogate pair are traps waiting to catch you. From vim's point of view (and from Visual Studio's point of view) both a line break and a surrogate pair are a single logical character. Visual Studio won't even let you place the caret at those locations. Like it or not, the presence of surrogate pairs means that a `TextBuffer` is effectively encoded. Unless we have an abstraction that allows us to decode the buffer and operate efficiently on logical characters, the alternative is a multitude of unnecessary and probably incomplete special cases.

ricksladkey: I will take another crack at this basically leaving `SnapshotColumn` untouched.

Src/VimCore/EditorUtil.fs
ricksladkey: Now reverted.

Src/VimCore/EditorUtil.fs
ricksladkey: I improved the efficiency of the new API even more!

Src/VimCore/EditorUtil.fs
ricksladkey: Removed those comments.

Src/VimCore/EditorUtil.fs
ricksladkey: I restored `SnapshotColumn` and created a new type `SnapshotCharacterSpan`. If you don't like that name, I can rename it.

Src/VimCore/EditorUtil.fs
ricksladkey: To clarify, the old `SnapshotColumn` is back and unchanged. The new `SnapshotCharacterSpan` still has an `IsLineBreak` property.

PR #2168, opened 2018-05-30
Block put

### LINKED ISSUE #1900: `P` and `p` only paste on the first line of a block selection
Steps to reproduce:

Given the Text:

```
test

foobar 1
foobar 2
foobar 3
```

- Move the cursor to `test` and yank it (e.g. `y i W`)
- Move the cursor to the first `foobar`
- Block-select all `foobar` occurrences (e.g. `0 C-v j j e`)
- Try to paste the yanked `test` to each line (`p`)

In vim, one would end up with all three occurences of `foobar` replaced by `test`, i.e.:
(tested using [vim.js](https://coolwanglu.github.io/vim.js/emterpreter/vim.html))
```
test

test 1
test 2
test 3
```

In Visual Studio, only the first line in the block selection gets replaced, the others are simply deleted:
(tested using both VS2015 and VS2010 w/ VsVim 2.2.0.0)
```
test

test 1
 2
 3
```

ricksladkey:
### Changes

- After a character-wise yank, position the cursor at the beginning of what was yanked
- When putting simple text over a block selection, replicate the text being put

### Fixes

- Fixes #1900

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -3251,6 +3251,10 @@ type internal CommandUtil
     member x.YankMotion registerName (result: MotionResult) =
         let value = x.CreateRegisterValue (StringData.OfSpan result.Span) result.OperationKind
         _commonOperations.SetRegisterValue registerName RegisterOperation.Yank value
+        match result.OperationKind with
+        | OperationKind.CharacterWise ->
+            TextViewUtil.MoveCaretToPoint _textView result.Start
+        | _ -> ()
         CommandResult.Completed ModeSwitch.NoSwitch
 
     /// Yank the lines in the specified selection
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -1766,7 +1766,7 @@ type internal CommandUtil
             | OperationKind.LineWise ->
                 x.CaretLine.EndIncludingLineBreak
 
-        x.PutCore point stringData registerValue.OperationKind moveCaretAfterText
+        x.PutCore point stringData stringData registerValue.OperationKind moveCaretAfterText
 
     /// Put the contents of the register into the buffer after the cursor and respect
     /// the indent of the current line.  Used for the ']p' command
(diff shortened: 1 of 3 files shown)


Src/VimCore/CommandUtil.fs
```diff
@@ -1970,6 +1970,18 @@ type internal CommandUtil
                     col |> Seq.iter (fun span -> edit.Delete(span) |> ignore)
                     edit.Apply() |> ignore
 
+                    // For character-wise put of simple text over a
+                    // block selection, replicate the text into a block.
+                    let newStringData =
+                        match operationKind, stringData with
+                        | OperationKind.CharacterWise, StringData.Simple text ->
+                            seq { for _ in col do yield text }
```
jaredpar: Consider: `Seq.repeat col.Count text`

Src/VimCore/CommandUtil.fs
```diff
@@ -1970,6 +1970,18 @@ type internal CommandUtil
...
                     edit.Apply() |> ignore
 
+                    // For character-wise put of simple text over a
+                    // block selection, replicate the text into a block.
+                    let newStringData =
+                        match operationKind, stringData with
+                        | OperationKind.CharacterWise, StringData.Simple text ->
+                            seq { for _ in col do yield text }
+                            |> NonEmptyCollectionUtil.OfSeq 
+                            |> Option.get
```
jaredpar: Hmm. May be able to simplify these first three lines to:

``` fsharp
NonEmptyCollectionUtil.Map (fun _ -> text) col 
```

That would avoid the `Option.get`.

Src/VimCore/CommandUtil.fs
```diff
@@ -1819,12 +1819,12 @@ type internal CommandUtil
...
             | OperationKind.LineWise -> x.CaretLine.Start
 
-        x.PutCore point stringData registerValue.OperationKind moveCaretAfterText
+        x.PutCore point stringData stringData registerValue.OperationKind moveCaretAfterText
 
     /// Put the contents of the specified register after the cursor.  Used for the
     /// normal 'p', 'gp', 'P', 'gP', ']p' and '[p' commands.  For linewise put operations
     /// the point must be at the start of a line
-    member x.PutCore point stringData operationKind moveCaretAfterText =
+    member x.PutCore point originalStringData stringData operationKind moveCaretAfterText =
```
jaredpar: Think we could simplify this a bit. After looking at this `originalStringData` and `stringData` only differ in the case that `stringData` is `StringData.Simple` instance and that's only useful in positioning the caret below. Suggestion I had is the following:

1. Move this to `Option string` and use a name along the lines of  `originalSimpleText`. 
1. Use this value to create a `StringData` to match on: 

``` fsharp
let m = 
  match originalSimpleText
  | Some s -> StringData.Simple s
  | None -> stringData
```

Even then the text is still duplicated between `originalSimpleText` and `stringData` and there is an implicit assumption they are equal. I can't really think of a great way around that though. Other than making the new parameter a `bool` and doing a more complex match.

ricksladkey: The bool ended up being the clearest way to indicate precisely what was desired, i.e "move the caret as if the string data were simple." Thanks for the suggestion to improve it.

Src/VimCore/CommandUtil.fs
```diff
@@ -1766,7 +1766,7 @@ type internal CommandUtil
             | OperationKind.LineWise ->
                 x.CaretLine.EndIncludingLineBreak
 
-        x.PutCore point stringData stringData registerValue.OperationKind moveCaretAfterText
```
jaredpar: Really like this approach.

PR #2169, opened 2018-05-31
Command margin and macros

### LINKED ISSUE #2077: Macro search forward and backward
If you have an xlm like text. You have two elements and you are in the second element's row. If you record a macro "/>?<" (first search to the tag's end, then search backwards to the tag's start). It's OK, but when you play the macro in new elements the cursor will be at the first element tag's start and not the second elmenet.

### LINKED ISSUE #2115: Search forward in two macros conflict
Consider this snippet
```
<tr>
            <td width="25%">@Caption("Planned Date")</td>
            <td width="25%">@Text("")</td>
            <td width="25%">@Caption("Manager")</td>
            <td width="25%">@Text("")</td>
</tr>
```
I record two macros one to target the "P" and remove "Planned Date", one to target the "M" and replace "Manager", "/P c2w" and "/M cw" respectively. What I find is that when performing the macro, VsVim doesn't correctly remember the search term but instead repeats the last search. So if I search /P before executing the macros, then the macros function as "/P c2w" and "/P cw" and same if I search for /M. Let me know if there's anything else I could test for you. You should be able to reproduce with the snippet above,.

ricksladkey:
Currently it is impossible to record a macro that either searches for something or contains a `:` command, because keystrokes entered while the command margin has the focus are never processed by the vim buffer that the command margin belongs to. To fix this problem, when enter is pressed in the command margin, the individual characters of the search text or command text are forwarded to the vim buffer for processing, enabling the macro recorder to record them.

### Changes

- Ensure margin command text is processed by the vim buffer (for macro recording)
- Prevent status messages from being squashed when replaying a macro

### Fixes

- Fixes #2077
- Fixes #2115

### DIFF
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
@@ -56,7 +56,14 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
 
             internal void Clear()
             {
-                Message = null;
+                if (!InEvent)
+                {
+                    // This might be a nested key input event, for example
+                    // if a macro is being replayed. Only clear the event
+                    // state for the outermost key input event so we don't
+                    // lose messages.
+                    Message = null;
+                }
                 SwitchModeEventArgs = null;
             }
         }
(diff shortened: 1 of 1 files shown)


Deadairx:
Just pulled this down and verified, it's working for me. Thanks for fixing it so quick!

ricksladkey:
@Deadairx Thanks for trying it out, Cody.  Much appreciated!

PR #2171, opened 2018-05-31
Implement ':r!' to read from a command into the buffer

### LINKED ISSUE #2167: `:!` and `:r!` behave strangely
I've been trying to run `cmd` commands from VsVim on Windows 10 and I'm having some trouble with it:

If I run a command (e.g. `:! dir`), an empty CLI window opens up. As long as it stays open, Visual Studio is blocked (the GUI doesn't react anymore), and when I close the CLI window, the output of the command is finally displayed in the VsVim prompt. This also happens with commands that are non-built-in commands (e.g. `:! powershell -command ls`).

If I try to run an application detached from VsVim using `start` (e.g. `:! start cmd`), the additional application opens up correctly, but Visual Studio stays blocked until it is closed again. Interestingly, the CLI window that opens up on normal commands gets closed almost immediately.

If I try to insert the output of a command using `:r!` (e.g. `:r! dir` or `:r! powershell -command ls`), the VsVim prompt displays the message `Can't run X` (`X` being the specified command).

I'm currently running VsVim v. 2.5.0.0 in VS2015 on Windows 10.

ricksladkey:
### Changes

- Support command flavor of `:read`
- Avoid editor hang executing `start [command]`
- Fix `:read` handling of line zero (to mean before line 1)
- Add unit tests for command flavor of `:read`

### Fixes

- Fixes #2167

### DIFF
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -2268,6 +2268,12 @@ module EditUtil =
                     inner (index + length)
         inner 0
 
+    /// Split a string into lines of text
+    let SplitLines (text: string) =
+        let text = RemoveEndingNewLine text
+        let text = NormalizeNewLines text "\n"
+        StringUtil.Split '\n' text
+
 /// In some cases we need to break a complete string into a series of text representations
 /// and new lines.  It's easiest to view this as a sequence of text values with their 
 /// associated line breaks
--- a/Src/VimWpf/VimHost.cs
+++ b/Src/VimWpf/VimHost.cs
@@ -255,15 +255,18 @@ namespace Vim.UI.Wpf
             // Use a (generous) timeout since we have no way to interrupt it.
             var timeout = 30 * 1000;
 
+            // Avoid redirection for the 'start' command.
+            var doRedirect = !arguments.StartsWith("/c start ");
+
             // Populate the start info.
             var startInfo = new ProcessStartInfo
             {
                 FileName = command,
                 Arguments = arguments,
                 UseShellExecute = false,
-                RedirectStandardInput = true,
-                RedirectStandardOutput = true,
-                RedirectStandardError = true,
+                RedirectStandardInput = doRedirect,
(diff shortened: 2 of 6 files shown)


Src/VimCore/EditorUtil.fs
```diff
@@ -2268,6 +2268,12 @@ module EditUtil =
                     inner (index + length)
         inner 0
 
+    /// Split a string into lines of text
+    let SplitLines (text: string) =
+        let text = RemoveEndingNewLine text
+        let text = NormalizeNewLines text "\n"
```
jaredpar: Why `"\n"` here vs. using `EditUtil.NewLine` helper to get the preferred new line character?

Src/VimWpf/VimHost.cs
```diff
@@ -255,15 +255,18 @@ public virtual RunCommandResults RunCommand(string command, string arguments, st
             // Use a (generous) timeout since we have no way to interrupt it.
             var timeout = 30 * 1000;
 
+            // Avoid redirection for the 'start' command.
+            var doRedirect = !arguments.StartsWith("/c start ");
```
jaredpar: Shouldn't this be a case insensitive comparison?

Src/VimCore/EditorUtil.fs
ricksladkey: Because we are only temporarily normalizing the lines into something that `String.Split` can handle. The return value of the method is an array of strings without any newlines.

Src/VimWpf/VimHost.cs
ricksladkey: Good point.

Src/VimCore/EditorUtil.fs
jaredpar: Ah I see now.

Src/VimWpf/VimHost.cs
ricksladkey: I made that change.

PR #2178, opened 2018-06-06
Enable debug output via the OutputWindow

### LINKED ISSUE #1916: Feature to enable debug output
I would love to see a debug feature which writes to a log file so at times when vsvim isn't working correctly there is some diagnosic info available.

At least once a day vsvim stops working and I need to restart Vs. Perhaps there's an issue with another plugin? Not sure but it would be a good excuse for me to open my first pull request ☺️

jaredpar:
There is now an option to display all of the vim trace / debug output in the output window. This is particularly valuable for debugging key stroke issues as well as other problems. 

closes #1916


Src/VsVimShared/VsVimShared.csproj
```diff
@@ -8,6 +8,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.0.0.0" />
+    <Reference Include="EnvDTE80">
+      <HintPath>..\..\References\Vs2010\envdte80.dll</HintPath>
```
jaredpar: Ugg. Missed that.

ricksladkey:
Pretty sure this will be useful for me!

PR #2179, opened 2018-06-06
Block caret colors

### LINKED ISSUE #749: Could the block caret observe both fore and background color
It seems the setting for "VsVim Block Caret" only observes the Foreground color and uses that as "background" color.
Would it be possible to make it observe both FG and BG colors?

For example, given the following settings:
- VsVim Block Caret : FG = Blue, BG = Yellow and 
- Normal Text : FG = Red, BG = Black

I would like the caret to blink OFF: Red on Black (this happens), ON: Blue on Yellow (this doesn't happen - the caret is Red on Blue).

### LINKED ISSUE #1945: Change color of block caret.
Hello,
Change color of "block caret" is not the same that I have set to. For example I want my color of block caret to be 255 ; 204 ; 0 but in editor ends up being 179 ; 146 ; 13.

The color I have setup:
image

The color in editor (little bit darker):
image

Does anyone have the same issue? Does anyone know how to fix it? 
edit: I think it's because caret is somehow transparent ¯\_(ツ)_/¯
Can this transparency be disabled?

ricksladkey:
Although VsVim currently supports a configurable block caret background color, the foreground color is always displayed as the character would appear if the caret were somewhere else. This change fixes that and allows the user to configure and display both foreground and background colors for the block caret. Doing this vastly improves readability and accessibility because it allows high contrast for the character under the cursor. In fact, it works so well that it makes it practical to turn off cursor blinking entirely.

Here is a screenshot of the new  block caret in action:

image

### Changes

- Allow block caret background to be configured
- Set default caret opacity to fully opaque (100%)
- Adapt VsVim options page to allow configuring block caret foreground and background
- Rewrite the block caret to actually draw the text of the character under the cursor
- Reverse the previous interpretation of background/foreground for the block caret

### Important Note

This change might produce somewhat surprising results when first installed.

First, the old VsVim used to use the configurable foreground color as the background of the block caret, which was wrong and confusing. This change fixes that and is therefore technically not backward-compatible, settings-wise. The price is a visit to the options page to fix the block caret colors.

Second, the old VsVim did not allow the block caret foreground to be set. Therefore, on first use, the block caret foreground will default to white, which is suitable for a light background window color. However, if a dark theme is in use, black or some other dark color might be a better choice and using white will temporarily *reduce* the contrast of the block caret. Again, a simple visit to the options page may be necessary to get the best results.

Third, some users may have previously overridden the block caret opacity using `:set vsvimcaret=XX`. Chances are that they did this to try to *improve* the readability. However, increasing the opacity of the new block caret will now almost certainly increase its readability. As a result, setting the opacity less than 100% is still supported, but as an aesthetic choice.

### Fixes

- Fixes #749
- Fixes #1945

### DIFF
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
@@ -406,6 +406,7 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
             var textRunProperties = _classificationFormatMap.DefaultTextProperties;
             var typeface = textRunProperties.Typeface;
             var fontSize = textRunProperties.FontRenderingEmSize;
+            var textHeight = offset + height;
 
             if (_caretOpacity < 1.0 && backgroundBrush is SolidColorBrush solidBrush)
             {
@@ -426,7 +427,10 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
                 FontStyle = typeface.Style,
                 FontSize = fontSize,
                 Width = width,
-                Height = offset + height,
+                Height = textHeight,
+                LineHeight = textHeight,
+                LineStackingStrategy = LineStackingStrategy.BlockLineHeight,
+                BaselineOffset = 0,
             };
 
             var element = new Canvas
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
@@ -303,7 +303,7 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
         private FormattedText CreateFormattedText()
         {
             var textRunProperties = _classificationFormatMap.DefaultTextProperties;
-            return new FormattedText("^", CultureInfo.CurrentUICulture, FlowDirection.RightToLeft, textRunProperties.Typeface, textRunProperties.FontRenderingEmSize, Brushes.Black);
+            return new FormattedText("A", CultureInfo.CurrentUICulture, FlowDirection.LeftToRight, textRunProperties.Typeface, textRunProperties.FontRenderingEmSize, Brushes.Black);
         }
 
         /// <summary>
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
@@ -393,21 +393,43 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
         private CaretData CreateCaretData()
         {
             var color = TryCalculateCaretColor();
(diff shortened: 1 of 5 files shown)


ricksladkey:
Here is a screenshot showing that surrogate pairs are handled correctly:

image

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -291,117 +293,154 @@ private Point GetRealCaretVisualPoint()
...
+                            && point.Position < _textView.TextSnapshot.Length - 1
+                            && Char.IsLowSurrogate(point.Add(1).GetChar()))
+                        {
+                            caretCharacter = new SnapshotSpan(point, 2).GetText();
+                        }
+                        else
+                        {
+                            caretCharacter = pointCharacter.ToString();
+                        }
+                        width = line.GetCharacterBounds(point).Width;
```
jaredpar: Why does the width of a single point get used here when there is a surrogate pair?

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -291,117 +293,154 @@ private Point GetRealCaretVisualPoint()
...
-            Canvas.SetTop(caretData.Image, point.Y + caretData.YDisplayOffset);
+            Canvas.SetLeft(caretData.Element, point.X);
+            Canvas.SetTop(caretData.Element, point.Y + caretData.YDisplayOffset);
         }
 
         private FormattedText CreateFormattedText()
         {
             var textRunProperties = _classificationFormatMap.DefaultTextProperties;
-            return new FormattedText("^", CultureInfo.CurrentUICulture, FlowDirection.RightToLeft, textRunProperties.Typeface, textRunProperties.FontRenderingEmSize, Brushes.Black);
+            return new FormattedText("A", CultureInfo.CurrentUICulture, FlowDirection.LeftToRight, textRunProperties.Typeface, textRunProperties.FontRenderingEmSize, Brushes.Black);
```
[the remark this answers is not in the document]
ricksladkey: It makes no difference what character is used when the font is fixed-width, which it nearly always is.

When I first looked at the code, the `"^"` string caused me to scratch my head. It wasn't clear to me why that particular string was chosen. I initially guessed that it was related to control character display somehow. Anyway, I tried to improve it by changing it to space, but it turns out that WPF won't measure the width of trailing whitespace. So finally I settled on the first alphabetic character as the best stand-in for a "typical character".

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: The `GetCharacterBounds` API effectively means "get the bounds of the character to the right of the caret, taking into account any surrogate pairs". In other words, the text view line itself it well aware of the surrogate pair.

PR #2187, opened 2018-06-10
Prevent vim buffer creation unless the view is in the text editor category

### LINKED ISSUE #1970: Should VsVim be enabled on F# Interactive mode?
F# Interactive is both broken and using VsVim, I assume VsVim is breaking some assumptions about  how the repl works as it is functionally useless.

ricksladkey:
The F# Interactive window is not in the "Text Editor" fonts and colors category so VsVim fonts and colors are not available. In general, unless a window *is* in the "Text Editor" fonts and colors category, VsVim shouldn't be active in that window.

### Changes

-Prevent vim buffer creation unless the view is in the text editor category

### Fixes

- Fixes #1970

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/VsAdapter.cs
+++ b/Src/VsVimShared/Implementation/Misc/VsAdapter.cs
@@ -318,6 +318,36 @@ namespace Vim.VisualStudio.Implementation.Misc
             return value.Equals(true) || value.Equals(1);
         }
 
+        internal bool IsTextEditorView(ITextView textView)
+        {
+            // Our fonts and colors won't work unless this view is in the "Text Editor"
+            // fonts and colors category.
+            var GUID_EditPropCategory_View_MasterSettings =
+                new Guid("{D1756E7C-B7FD-49a8-B48E-87B14A55655A}"); // see {VSIP}/Common/Inc/textmgr.h
+            var textEditorGuid =
+                new Guid("{A27B4E24-A735-4d1d-B8E7-9716E1E3D8E0}"); // Text editor category
+
+            var vsTextView = _editorAdaptersFactoryService.GetViewAdapter(textView);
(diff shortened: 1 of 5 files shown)


Src/VsVimShared/Implementation/Misc/VsAdapter.cs
```diff
@@ -318,6 +318,36 @@ internal bool IsWatchWindowView(ITextView textView)
             return value.Equals(true) || value.Equals(1);
         }
 
+        internal bool IsTextEditorView(ITextView textView)
+        {
+            // Our fonts and colors won't work unless this view is in the "Text Editor"
+            // fonts and colors category.
```
jaredpar: FWIW: Not having the fonts and colors work is less than ideal but it's not necessarily a deal breaker for me.

Src/VsVimShared/VsVimHost.cs
```diff
@@ -771,6 +771,11 @@ public override bool ShouldCreateVimBuffer(ITextView textView)
                 return false;
             }
 
+            if (!_vsAdapter.IsTextEditorView(textView))
```
jaredpar: Changing this particular function always makes me a bit nervous. VS is such a diverse ecosystem and there are always components out there which find new and odd ways to create windows out that surprise me. In the past I've ended up disabling VsVim for scenarios that I wasn't aware of. 

The good news about changing this though is if we get it wrong people complain ... fast. And while impactful to users it can also be fixed rather fast. 

The advice I was given by the editor team when I started VsVim was to focus on the `TextViewRoles` of an `ITextView` for controlling where VsVim should be displayed. If a particular window doesn't want VsVim then it shouldn't use the structured role, etc ... That advice works up to a point but there are just so many real examples where it falls down (like this one).

ricksladkey: Gulp! Fingers crossed.

PR #2193, opened 2018-06-11
Prevent putting caret in virtual space for end-of-line marks

### LINKED ISSUE #2191: Jumping exactly to a mark may leave caret in virtual space

ricksladkey:
Fixes #2191


ricksladkey:
This had a conflict with PR #2139, so I merged it into that one and am closing this one.

Issue #2181, opened 2018-06-08
Implement the 'showcmd' feature


dhmgit:
When the 'showcmd' setting is on in Vim, the last line of the screen shows partially typed commands and other context information.

http://vimdoc.sourceforge.net/htmldoc/options.html#'showcmd'

I find this feature very helpful, so I've implemented it in my VsVim fork for my own use.  I've never submitted a PR, so I'm adding this feature request as a first step.  If this project is open to new contributors, I'd like to submit my work for review.

Thanks to everyone who has contributed to this project for greatly enhancing my VS editing experience.  VsVim is awesome!

ricksladkey:
I don't speak for the project, but I can say that I am simply an enthusiastic user, like you, who wants to give back to a project that is important to me. I definitely think you should submit a PR for review.

jaredpar:
>  If this project is open to new contributors, I'd like to submit my work for review.

Absolutely open to contributions :smile:

PR #2194, opened 2018-06-11
Showcmd option

dhmgit:
Proposed implementation of feature request: #2181 

I'm a bit unsatisfied with the frequent double-update of the showcmd area in CommandMarginController, once on KeyInputEnd, and again on KeyInputProcessed.  There may be a more efficient way of positioning these updates, but not being 100% comfortable with the key-processing model, I'm playing it safe here.  Note that AFAIK, KeyInputProcessed only needs to be handled due to the fact that buffered keys could be processed in KeyMappingTimeoutHandler.OnTimerTick after a timeout.

### DIFF
--- a/Src/VimCore/CommandRunner.fs
+++ b/Src/VimCore/CommandRunner.fs
@@ -73,6 +73,8 @@ type internal CommandRunner
         | Some count -> count
         | None -> 1
 
+    member x.Inputs = List.rev _data.Inputs
+
     /// Try and get the VisualSpan for the provided kind
     member x.GetVisualSpan kind = VisualSpan.CreateForSelection _textView kind _localSettings.TabStop
 
@@ -395,6 +397,7 @@ type internal CommandRunner
         member x.HasCount = x.HasCount
         member x.RegisterName = x.RegisterName
         member x.Count = x.Count
+        member x.Inputs = x.Inputs
         member x.KeyRemapMode = 
             match _runBindData with
             | None -> KeyRemapMode.None
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginUtil.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginUtil.cs
@@ -213,7 +213,9 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
         public static string GetShowCommandText(IVimBuffer vimBuffer)
         {
             if (vimBuffer.IncrementalSearch.InSearch)
+            {
                 return string.Empty;
+            }
 
             switch (vimBuffer.ModeKind)
             {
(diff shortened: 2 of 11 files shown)


Src/VimWpf/Implementation/CommandMargin/CommandMarginUtil.cs
```diff
@@ -211,5 +209,61 @@ string ICommandMarginUtil.GetStatus(IVimBuffer vimBuffer)
...
+                    return GetVisualModeShowCommandText(vimBuffer, (IVisualMode) vimBuffer.Mode);
+            }
+
+            return string.Empty;
+        }
+
+        private static string GetVisualModeShowCommandText(IVimBuffer vimBuffer, IVisualMode visualMode)
+        {
+            if (visualMode.CommandRunner.Inputs.Any())
+                return string.Concat(visualMode.CommandRunner.Inputs.Select(x => x.Char));
```
ricksladkey: It's a minor point, but my impression from the code base is that the preferred style is braces around even single statements. But what @jaredpar thinks is what really matters.

Test/VimWpfTest/CommandMarginUtilTest.cs
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Vim.UI.Wpf.UnitTest
+{
+    public class CommandMarginUtilTest : VimTestBase
+    {
+        private readonly MockRepository _factory;
+        private readonly Mock<IIncrementalSearch> _search;
+        private readonly MockVimBuffer _vimBuffer;
+
+        public CommandMarginUtilTest()
```
ricksladkey: Wow, you really got the full-blown experience by cooking up one of these test harnesses from scratch!

Src/VimCore/CommandRunner.fs
```diff
@@ -73,6 +73,8 @@ type internal CommandRunner
         | Some count -> count
         | None -> 1
 
+    member x.Inputs = List.rev _data.Inputs
```
ricksladkey: I always try to remember triple slash comments for members (shows when you hover in the caller).

dhmgit:
@ricksladkey, thank you for the detailed comments.  I've adding processing for normal-mode ICommandRunner inputs as well as translation of ASCII control chars to (^) notation.

Src/VimCore/CommandRunner.fs
jaredpar: It sees the only place we ever use `CommandRunnerData.Inputs` is this function now. Every use is reversing and allocating a new list. I wonder if we should just change `CommandRunnerData.Inputs` to store the forward list here and avoid the reverse on every access of this member. It would save a few allocations.

jaredpar:
Thanks for the contribution!

Src/VimCore/CommandRunner.fs
dhmgit: I wondered about the same.  Not being very familiar with the codebase at this point, I wasn't sure if there was a good reason for stacking the inputs as they came in.

jaredpar: I think this code dates back from when I was first writing VsVim and learning F#. I suspect I was attempting to be clever and avoid a full list allocation on every key press by storing the keys in reverse. Then doing a one time reverse once the sequence was completed. Looks like I never actually used that value though 😦

dhmgit: I'm just learning F# myself, so I didn't realize lists were singly-linked until digging deeper.  Appending is not only O(n) but also reallocates the original list.  That's not what I expected!
I think we're talking about very small lists in any realistic scenario (that I can think of), but keeping the inputs in order of entry also makes more semantic sense to me.

As soon as I get a chance, I'll take care of this detail and submit another PR.

PR #2198, opened 2018-06-12
Improve out-of-the-box fidelity with vim

ricksladkey:
### Changes

- Set the default behavior to GVim74 instead of GVim73 (gvim has defaulted to `:behave mswin` for five years now)
- Remove "advanced key" exceptions to "handle all with VsVim" because VsVim emulates all of them

# Discussion

Vim assumes that on Windows, new users will be most familiar with Windows-like `selectmode` and `<C-c>` to copy, and `<C-v>` to paste. Yet VsVim currently defaults to `behave=xterm` and even users who select "handle all with VsVim" need to manually toggle half a dozen settings to get the full VsVim keyboard.

VsVim can get the best of both worlds by defaulting to `behave mswin` (the effective meaning of GVim74) and making it easy to truly handle all keys with VsVim.

### DIFF
--- a/Src/VsVimShared/Implementation/OptionPages/KeyboardSettingsControl.xaml.cs
+++ b/Src/VsVimShared/Implementation/OptionPages/KeyboardSettingsControl.xaml.cs
@@ -30,7 +30,6 @@ namespace Vim.VisualStudio.Implementation.OptionPages
             typeof(KeyboardSettingsControl));
 
         private readonly ObservableCollection<KeyBindingData> _keyBindingList = new ObservableCollection<KeyBindingData>();
-        private readonly HashSet<KeyBindingData> _advancedSet = new HashSet<KeyBindingData>();
         private readonly IVim _vim;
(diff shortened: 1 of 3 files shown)


Src/VsVimShared/Implementation/OptionPages/KeyboardSettingsControl.xaml.cs
```diff
@@ -134,31 +126,6 @@ private void OnDisableAllVimKeysClick(object sender, RoutedEventArgs e)
             _keyBindingList.ForEach(x => x.HandledByVsVim = false);
         }
 
-        /// <summary>
-        /// Is this an advanced key stroke like CTRL+V which we don't want to automatically 
-        /// convert.  
-        /// </summary>
-        private bool IsAdvanced(KeyStroke keyStroke)
```
jaredpar: I've honestly never liked this function and behavior. Always felt like a hack. But I was a bit reluctant to take over literally everything here. Feel like there is a chance of backlash around this but think it's worth trying out.

PR #2205, opened 2018-06-13
Return to normal mode when an unfocused buffer's selection is cleared

### LINKED ISSUE #2204: Go to definition or find in files may create zero-width selection
When the user uses Visual Studio services like go to definition or find in files, VsVim programmatically clears the selection to prevent the buffer going into visual or selection mode. However, if the buffer was already in a visual mode or a select mode, it stays in that mode, but with a zero-width selection.

This is nonsensical for visual line mode and counterproductive for all visual and select modes, because the user expects to arrive at the target in normal mode. Unless they notice the altered selection and hit `<Esc>`, beginning typing normal commands will not have the desired effect.

ricksladkey:
Fixes #2204

### DIFF
--- a/Src/VimCore/Modes_Visual_SelectMode.fs
+++ b/Src/VimCore/Modes_Visual_SelectMode.fs
@@ -248,7 +248,7 @@ type internal SelectMode
 
     interface IMode with
         member x.VimTextBuffer = _vimTextBuffer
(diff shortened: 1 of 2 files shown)


jaredpar:
Good find on that bug! I've hit it many times as well.

Issue #2180, opened 2018-06-07
Setting notimeout is not respected


dhmgit:
Setting notimeout does not prevent partially typed commands from timing out.  In contrast, Vim will wait indefinitely for a command to be completed.

This may be naive given my lack of experience with the codebase, but simply adding a return in KeyMappingTimeoutHandler.OnKeyInputBuffered seems to fix the issue for me.  Also, a nearby comment seems to indicate that this was the original intent.

```
if (!globalSettings.Timeout)
{
     _timer.Stop();
    return; // added
}
```

ricksladkey:
This is fixed and can be closed! Thanks, @dhmgit

PR #2207, opened 2018-06-13
Fix notimeout

dhmgit:
Proposed fix for #2180


Src/VimWpf/Implementation/Misc/KeyMappingTimeoutHandler.cs
```diff
@@ -80,6 +80,7 @@ private void OnKeyInputBuffered(object sender, KeyInputSetEventArgs args)
                     if (!globalSettings.Timeout)
                     {
                         _timer.Stop();
+                        return;
```
jaredpar: That comment above this even explicitly says "and return". Not sure how we missed this originally. Thanks for finding this!

PR #2210, opened 2018-06-16
Turn off the IME for normal and other non-input modes

### LINKED ISSUE #1027: Ignore IME state in normal mode
Would you mind giving a switch to ignore IME state in normal mode?

ricksladkey:
### Fixes

- Fixes #1027

### Discussion

The IME is the "Input Method Editor" used with many languages to enter non-latin characters. When it is off, the keyboard keys generate their normal characters. While the IME aids in the entry of non-latin characters, it makes entry of latin characters more difficult. Since vi/vim/VsVim all use many latin characters as commands, it is inconvenient to have the IME enabled when in, say, normal mode.

This change passively monitors the focused VsVim document and coordinates the adjustment of the IME state so that the user controls the IME state when VsVim is in any kind of input mode (like insert mode), but the IME coordinator turns off the IME when VsVim is in any kind of non-input mode (like normal mode).

Input modes are defined as:

- Insert mode
- Replace mode
- Any select mode
- External edit mode
- Disabled mode
- Normal mode when entering a search pattern

All other modes are non-input modes.

When entering an input mode from a non-input mode, the IME coordinator will use the IME setting saved when last exiting an input mode. In all other case, it will disable the IME.

Since this change does not modify VsVim itself and since the IME setting has no effect when there is no IME available, it should be a harmless change. However, feedback from users who use languages with an IME is encouraged.


TKumadaki:
@ ricksladkey Because I am using IME, this feature is very glad.
But I want you to choose three options.

1. Use this function.
2. Do not use this function.
3. This function is used, but when entering Insert mode, Ime's state will not be restored.

About 2.
There is a solution to turn Ime off when hooking the keyboard and pressing the Esc key.
Users using this solution may not want to use this feature.
Specifically, if using both Vim and VsVim

About 3.
I do not remember the state of Ime I last entered.
So it is easier for Ime to be off when entering input mode.
Of course, I think that other people have better memory than I am.

Perhaps if can determine the state of Ime by the color of the caret,
I think that this function will be easy to use.

The Japanese version of GVim is designed to recognize the state of Ime with the color of the caret.

Since it is a personal opinion, I also want to hear opinions from other people.

ricksladkey:
@TKumadaki I know @jaredpar's preference is to use the same features/settings as gvim so I am thinking that implementing the settings `imimsert` and `imdisable` should let you accomplish 2. and 3. The vim documentation suggests the following mapping to implement 3:

```
:inoremap <ESC> <ESC>:set iminsert=0<CR>
```
The cursor color is a nice idea, but I think I'll hold off on that change for this PR.

Src/VimWpf/Implementation/ImeCoordinator/ImeCoordinator.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+
+        private void OnInputModeRelatedEvent(IVimBuffer vimBuffer)
+        {
+            var wasInInputMode = _inInputMode;
+            var isInInputMode = IsInputMode(vimBuffer);
+            _inInputMode = isInInputMode;
+
+            if (wasInInputMode != isInInputMode)
```
jaredpar: Imagine there are two buffers both in insert mode with `ImeState` set to `InputMethodState.On`. Now I hit escape in buffer 1, switch to buffer 2 and hit escape. Won't we end up with with both `_lastInputMethodState` and `ImeState.Current` set to `Off`?

Src/VimWpf/Implementation/ImeCoordinator/ImeCoordinator.cs
```diff
@@ -0,0 +1,121 @@
...
+        [ImportingConstructor]
+        internal ImeCoordinator(IVim vim)
+        {
+            _vim = vim;
+
+            _inInputMode = true;
+            _lastInputMethodState = InputMethodState.DoNotCare;
+        }
+
+        void VimBufferCreated(IVimBuffer vimBuffer)
```
jaredpar: Missing `private`

jaredpar: Missing `private` here

jaredpar:
@ricksladkey it does seem like we should tie this to a setting. Briefly looking through those settings (`iminsert` and `imdisable`) they seem pretty relevant here. Should we implement them as a part of this PR?

ricksladkey:
> Should we implement them as a part of this PR?

@jaredpar Yes, I think we should. I asked for feedback from someone who uses an IME regularly and @TKumadaki made some good points.

I'll make those changes shortly.

ricksladkey:
I implemented all of `imcmdline`, `imdisable`, `iminsert` and `imreplace` and tacked on `<C-^>` in insert mode.

ricksladkey:
Interesting related discussion over in https://github.com/vim/vim/issues/2071

ricksladkey:
Wow, those options are crazy and I'm now a lot less certain of the correctness of this PR! It was also way more work than I had counted on when I started this PR. But I like to see a thing through to its completion and I think the overall structure is clean and maintainable, so I'm hoping feedback from daily IME users can ferret out any issues that didn't show up in my testing.

I will try to think about how this could be tested to lock in the major points but the combinations of cases are a little overwhelming: all corner cases...

TKumadaki:
I would like to help you, what should I do?
Should I use VsVim in your repository?

ricksladkey:
@TKumadaki If @jaredpar merges this, you can try out a build with the latest sources:

https://github.com/jaredpar/VsVim/wiki/FAQ#how-can-i-install-a-build-with-the-latest-sources

from the Open Vsix Gallery:

http://vsixgallery.com/extension/VsVim.Microsoft.e214908b-0458-4ae2-a583-4310f29687c3/

ricksladkey:
@TKumadaki It's quite a bit more work, but if you really want to test this before @jaredpar merges it, you are willing to build it yourself, and you have already cloned https://github.com/jaredpar/VsVim, then you can checkout this PR by follow these instructions:

https://help.github.com/articles/checking-out-pull-requests-locally/

For example:

```
git fetch origin pull/2210/head:ime-coordinator
```

TKumadaki:
@ricksladkey
According to your advice, I built VsVim.

As long as I used it a bit, I had no problem.

I am planning to use it in the office for about a week.
Then I would like to report the results.

If you have anything else you want to do, please let me know.

ricksladkey:
@TKumadaki Thanks a lot; I eagerly await your report!

TKumadaki:
@ricksladkey
I will report the points I noticed although I have not completed it for a week.

1. About f and r in normal mode
   In Vim, pressing f and r in normal mode restores the state of the last ime
   It is not so in VsVim.

2. About Ctrl- ^
   ^ The position of the keyboard is different in Japanese and English.
   Japanese keyboards, Ime is toggled with ctrl - &.
   Vim work correctly with Japanese keyboard.

I do not mind these problems.
But you and others may not.
So I decided to report.

TKumadaki:
@ricksladkey
sorry.
With vim on Japanese keyboard, ime is toggled with Ctrl-& or Ctrl-^.

ricksladkey:
@TKumadaki Thanks, I was aware of the `f` and `r` behavior after reading the vim docs but didn't implement it yet. I will take a look at how hard it would be.

ricksladkey:
@TKumadaki It turned out to be fairly easy to support `f`, `F` and `r`. In my testing it works just like vim. If you could upgrade to the new version and test it yourself that would be great!

TKumadaki:
@ ricksladkey I upgraded and confirmed that the behavior of f and r was fixed.
I noticed today, but when I complete the f and r commands, or cancel the command
The IME was not off.
This problem was also solved.
I Continue to use this VsVim.

I will ask you
Should I report the points I noticed immediately?
Should I put it together after a week?

I am worried about your burden.

ricksladkey:
@TKumadaki Frequent reports are not a burden to me. Thanks for your contribution to making this feature successful!

TKumadaki:
@ricksladkey The problem has not occurred so far.
There is something I want to report though it is not a problem.

This is the behavior when the tab is changed.
When tab is changed with IME on in normal mode
The next tab is IME on.

I think IME should be Off at this time.
But GVim does the same thing.

So I do not wish to fix it.
However, I am a little disappointed with the behavior of GVim.

TKumadaki:
@ricksladkey I used VsVim with the following settings.

1. Use as initial setting
   I used it for 3 days.

2. `: inoremap <ESC> <ESC>: set iminsert = 0 <CR>`
  I used it for 3 days.

3. `set iminsert = 2,: inoremap <ESC> <ESC>: set iminsert = 2 <CR>`
  I used it for 1 day.

About 1.,As far as I remember, the state of the IME has been successfully restored.

As for 2.,there was no problem.

For 3., I think that this setting always turns on IME in Insert Mode.
  However, it turned into Insert mode about 2 times with the IME turned off.
  After that, even if I did the same work I did not reproduce.
  It may be my misunderstanding.

If this PR is merged, for 1 and 2
If there is a problem I can report to the issue.

About 3, I usually do not use this setting.
I would like feedback from other users.

Src/VimWpf/Implementation/ImeCoordinator/ImeCoordinator.cs
```diff
@@ -0,0 +1,423 @@
...
+                                return InputMethodState.On;
+                            }
+                            else
+                            {
+                                return InputMethodState.Off;
+                            }
+                        }
+
+                    default:
+                        throw new ArgumentException("inputMode");
```
jaredpar: Can use `nameof(inputMode)`

Src/VimWpf/Implementation/ImeCoordinator/ImeCoordinator.cs
```diff
@@ -0,0 +1,423 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Windows.Input;
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace Vim.UI.Wpf.Implementation.ImeCoordinator
+{
+    [Export(typeof(IVimBufferCreationListener))]
+    internal sealed class ImeCoordinator : IVimBufferCreationListener, IDisposable
```
jaredpar: Is this `IDisposable` invoked during normal VS operation? I don't believe the Vim implementation will probe `IVimBufferCreationListener` for `IDisposable`.

ricksladkey: I gathered from the MEF comments elsewhere in VsVim that if MEF creates it and it's `IDisposable`, that MEF will dispose it. It's not particularly critical as this is basically a singleton for the duration of Visual Studio like VsVim itself.

Src/VimWpf/Implementation/ImeCoordinator/ImeCoordinator.cs
ricksladkey: I'm still catching up to now ancient C# features!

Src/VimCore/VimSettings.fs
```diff
@@ -159,6 +159,10 @@ type internal GlobalSettings() =
             (HistoryName, "hi", SettingValue.Number(VimConstants.DefaultHistoryLength), SettingOptions.None)
             (IncrementalSearchName, "is", SettingValue.Toggle false, SettingOptions.None)
             (IgnoreCaseName,"ic", SettingValue.Toggle false, SettingOptions.None)
+            (ImeCommandName, "imc", SettingValue.Toggle false, SettingOptions.None)
+            (ImeDisableName, "imd", SettingValue.Toggle false, SettingOptions.None)
+            (ImeInsertName, "imi", SettingValue.Number 0, SettingOptions.None)
+            (ImeSearchName, "ims", SettingValue.Number -1, SettingOptions.None)
```
jaredpar: Shouldn't the default for "ims" be 0?

ricksladkey: Not sure if it is released yet, but vim master has changed the default to -1, as discussed in this issue: https://github.com/vim/vim/issues/2071.

jaredpar: I stand corrected :)

PR #2215, opened 2018-06-17
Make session data reversible for macro registers

### LINKED ISSUE #2214: Macros containing special keys are not reloaded properly
If a macro is recorded using a special key (e.g. `<Left>`), and Visual Studio is then exited and restarted, the macro will not work properly. The session data serialization file format does not distinguish macro registers from string registers, and cannot faithfully encode the contents of macro registers containing special keys.

ricksladkey:
Fixes #2214

### DIFF
--- a/Src/VimCore/Vim.fs
+++ b/Src/VimCore/Vim.fs
@@ -26,6 +26,9 @@ type SessionRegisterValue = {
 
     [<field: DataMember(Name = "value")>]
     Value: string
+
+    [<field: DataMember(Name = "isMacro")>]
+    IsMacro: bool
 }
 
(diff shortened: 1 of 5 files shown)


Src/VimCore/Vim.fs
```diff
@@ -784,13 +791,23 @@ type internal Vim
             _ as ex -> VimTrace.TraceError ex
 
     member x.LoadSessionDataCore filePath =
-        let sessionData = x.ReadSessionData (x.GetSessionDataFilePath())
+        let sessionData = x.ReadSessionData filePath
```
jaredpar: Nice catch.

PR #2226, opened 2018-06-20
Append empty transaction when completing an active linked undo transaction

### LINKED ISSUE #2225: Undo after adjacent inserts undoes too much
If you create an empty buffer and type `iabc<Esc>` to insert `abc` and then `adef<Esc>` to append `def`, then you get one line containing `abcdef`. If you then press `u`, VsVim will undo both insertions. In vim, separate insertions, even adjacent ones, are separate steps in the undo sequence.

ricksladkey:
Fixes #2225

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -888,7 +888,8 @@ type internal CommandUtil
     /// Used for the several commands which make an edit here and need the edit to be linked
     /// with the next insert mode change.
     member x.CreateTransactionForLinkedChange name action =
-        let transaction = _undoRedoOperations.CreateLinkedUndoTransaction name
+        let flags = LinkedUndoTransactionFlags.EndsWithInsert
+        let transaction = _undoRedoOperations.CreateLinkedUndoTransactionWithFlags name flags
 
         try
             x.EditWithUndoTransaction name action
--- a/Src/VimCore/UndoRedoOperations.fs
+++ b/Src/VimCore/UndoRedoOperations.fs
@@ -363,7 +363,16 @@ and UndoRedoOperations
                 | UndoRedoData.Linked (count, false) :: tail -> 
                     Contract.Assert(count > 0)
                     hasData <- true
-                    _undoStack <- UndoRedoData.Linked (count, true) :: tail
+
+                    // Append an empty transaction to the previous undo transaction
+                    // to break the undo sequence.
+                    match _textUndoHistory with
+                    | Some textUndoHistory ->
+                        use undoTransaction = textUndoHistory.CreateTransaction "Break undo sequence"
+                        undoTransaction.Complete()
+                    | None -> ()
+
+                    _undoStack <- UndoRedoData.Linked (count + 1, true) :: tail
                 | _ -> ()
 
                 // If a linked undo operation completes that contains 0 undo / redo items that very likely 
--- a/Src/VimCore/UndoRedoOperations.fs
+++ b/Src/VimCore/UndoRedoOperations.fs
@@ -352,9 +352,13 @@ and UndoRedoOperations
     ///     - transactions we orphaned in ResetState because of detected errors
     ///     - transactions closed out of order 
     member x.LinkedUndoTransactionClosed (linkedUndoTransaction: LinkedUndoTransaction) = 
+        let canBeEmpty = Util.IsFlagSet linkedUndoTransaction.Flags LinkedUndoTransactionFlags.CanBeEmpty
+        let endsWithInsert = Util.IsFlagSet linkedUndoTransaction.Flags LinkedUndoTransactionFlags.EndsWithInsert
+
         match x.UndoTransactionClosedCore linkedUndoTransaction _linkedUndoTransactionStack with
         | TransactionCloseResult.Expected -> 
-            if _linkedUndoTransactionStack.Count = 0 && Option.isSome _textUndoHistory then
+            match _linkedUndoTransactionStack.Count, _textUndoHistory with
+            | 0, Some textUndoHistory ->
                 let mutable hasData = false
 
                 // The linked undo transaction is now done, need to freeze the data on the undo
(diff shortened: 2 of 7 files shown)


Src/VimCore/UndoRedoOperations.fs
```diff
@@ -363,7 +363,16 @@ and UndoRedoOperations
...
                     Contract.Assert(count > 0)
                     hasData <- true
-                    _undoStack <- UndoRedoData.Linked (count, true) :: tail
+
+                    // Append an empty transaction to the previous undo transaction
+                    // to break the undo sequence.
+                    match _textUndoHistory with
+                    | Some textUndoHistory ->
+                        use undoTransaction = textUndoHistory.CreateTransaction "Break undo sequence"
+                        undoTransaction.Complete()
```
jaredpar: Putting the fix here means that every single linked undo transaction now has an extra transaction put on top of it. That feels like a pretty big hammer. Won't that impact all of the other places, in addition to the insert case, where we use linxed undo transactions? I'm thinking of all the cases in command util where this happens.

ricksladkey: I have been testing this for a few days with no ill effects, but I agree now that the fix is fairly invasive for the how seemingly minor the bug is (that `<C-g>u` in insert mode is useless).

I also came across a very helpful way to visualize the undo stack using the drop down on the undo button in the toolbar:

image

Anyway, to address your comment, I think we only need to add the empty undo transact after inserts, which will cut down on the number of undo transactions considerably. I'll make that change.

Src/VimCore/UndoRedoOperations.fs
```diff
@@ -363,7 +363,16 @@ and UndoRedoOperations
                 | UndoRedoData.Linked (count, false) :: tail -> 
                     Contract.Assert(count > 0)
                     hasData <- true
-                    _undoStack <- UndoRedoData.Linked (count, true) :: tail
+
+                    // Append an empty transaction to the previous undo transaction
+                    // to break the undo sequence.
+                    match _textUndoHistory with
```
jaredpar: This is asserted above in the `if`. Can probably simplify those two checks.

Src/VimCore/UndoRedoOperations.fs
jaredpar: Hmm. After debugging into this a bit I understand what is going on here better. The native VS undo manager, which is what backs undo in the VS case, will create a single transaction for text edits which can be merged so long as the text edit is forward. I'm a bit worried the current implementation of undo doesn't fully account for this. Going to chat with someone tomorrow about this.

ricksladkey: Yes, the problem does not manifest in either the unit tests or VimApp because of the simpler (non-merging) undo implementation. This is why I didn't catch it when I implemented `<C-g>u` in insert mode to break the undo sequence.

My original goal, after understanding that some undo transactions can be merged, was to somehow create an unmergeable insert transaction. That seems doable, except it turned out that Visual Studio itself is creating the insert transactions. Even if we could get it work, it gets into implementing our own undo transactions, undo primitives, merge strategies, etc. which itself is a lot of work when what we want is simple: please don't merge these adjacent insert transactions.

Src/VimCore/UndoRedoOperations.fs
ricksladkey: Fixed that.

Src/VimCore/UndoRedoOperations.fs
ricksladkey: Here is the a sample of the undo stack with the new "append empty undo transaction only when the linked transaction ends with an insert". I did one insert (`aabc<Esc>`), then another insert (`adef<Esc>`), then a normal command that modified the buffer (`:sort`):

image

As you can see, the normal command (which creates a linked undo transaction), does not cause the empty undo transaction to be appended.

Src/VimCore/CommandUtil.fs
```diff
@@ -888,7 +888,8 @@ type internal CommandUtil
     /// Used for the several commands which make an edit here and need the edit to be linked
     /// with the next insert mode change.
     member x.CreateTransactionForLinkedChange name action =
-        let transaction = _undoRedoOperations.CreateLinkedUndoTransaction name
+        let flags = LinkedUndoTransactionFlags.EndsWithInsert
+        let transaction = _undoRedoOperations.CreateLinkedUndoTransactionWithFlags name flags
```
jaredpar: Why do we want to do this for all the command based commands? I'm trying to think of one that would be affected by this.

Src/VimCore/UndoRedoOperations.fs
```diff
@@ -363,16 +367,28 @@ and UndoRedoOperations
...
+                    // transaction to the previous undo transaction to break the undo sequence.
+                    let extra =
+                        if endsWithInsert then
+                            use undoTransaction = textUndoHistory.CreateTransaction "Break undo sequence"
+                            undoTransaction.Complete()
+                            1
+                        else
+                            0
+
+                    _undoStack <- UndoRedoData.Linked (count + extra, true) :: tail
```
jaredpar: After thinking more about this I do agree this is the best approach. There really isn't a better option here. 

I considered ways we could detect this case and add the transaction only when needed. But given the VS implementation of `ITextUndoHistory` is largely unimplemented (never exposes a transaction directly) there really isn't much of an option. 

This is just kind of how it is with `ITextUndoHistory`. The API is actually really nice at a glance but the VS implementation is the one that matters and it effectively removes most of the API surface.

Src/VimCore/CommandUtil.fs
ricksladkey: Perhaps the function name isn't specific enough right now, but `CreateTransactionForLinkedChange` is only called from three places, and all of them are defined to end in insert mode.

ricksladkey: I'm seeing now the comment above `CreateTransactionForLinkedChange` that says:

> and need the edit to be linked with the next insert mode change

which is exactly what the flag specifies.

PR #2227, opened 2018-06-22
Disable 'outlining undo' on all vim buffers

### LINKED ISSUE #2184: Folding seems to become a part of the undo-chain
I noticed something that I find a bit weird ;-)
Repro:

1. Open a document that VS knows to fold
2. Delete a line
3. Open or close a fold
4. u for undo

Undo will undo the folding, not undo 2. A second undo will then restore the line.
I tested this two VsVims (2.4.xy and 2.5) in two VS (2015 and 2017 ).

ricksladkey:
Fixes #2184

### DIFF
--- a/Src/VsVimShared/HostFactory.cs
+++ b/Src/VsVimShared/HostFactory.cs
@@ -97,6 +97,9 @@ namespace Vim.VisualStudio
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


jaredpar:
That was much easier than I was anticipating. Nice to see there was an explicit option for this.

PR #2228, opened 2018-06-24
Go to window

### LINKED ISSUE #625: Add support for CTRL-W, CTRL-W
Move to next window

### LINKED ISSUE #1565: Add alternate move cursor commands
See `:help window-move-cursor` for more details.
Some of them have a different behaviour from stated in the documentation, like 1 Ctlr+W Ctrl+L which deletes the current line rather than moving the cursor to one window right of the current one.

(Btw, thanks a ton for this plugin! :+1: )

### LINKED ISSUE #1623: Ctrl+ww switching between panes
In vim if I do a `:vsplit`, I can switch between splits using `<C-ww>`. I think it would be great if VsVim could implement this functionality too!

### LINKED ISSUE #1836: Split navigation broken in diff view
Select a commit from the TeamExplorer History view and select `Compare with previous`. This opens a split diff view, but the `<ctrl-w>*` shortcuts for navigating between splits don't work.

ricksladkey:
### Changes

- Rewrite `<C-w> hjkl` to operate on vim buffers instead of Visual Studio windows
- Make `<C-w> hjkl` non-wrapping and support the count argument
- Implement `<C-w> HJKL` (go to far left/down/up/right window)
- Implement `<C-w> wW` (go to next/previous window, with count)
- Implement `<C-w> tb` (go to top/bottom window)
- Implement `<C-w> <Left><Down><Up><Right>` (aliases for `<C-w> hjkl`)
- Stub out `<C-w> p` (go to recent window, requires PR #2139)
- Detect binding conflicts in multi-key bindings, e.g. `<C-w><C-k>` and `<C-w><C-l>`

### Fixes

- Fixes #625
- Fixes #1565
- Fixes #1623
- Fixes #1836


Src/VimCore/CoreInterfaces.fs
```diff
@@ -4221,7 +4221,7 @@ type IVimHost =
     abstract Make: jumpToFirstError: bool -> arguments: string -> unit
 
     /// Move the focus to the ITextView in the open document in the specified direction
-    abstract MoveFocus: textView: ITextView -> direction: Direction -> unit
+    abstract GoToWindow: textView: ITextView -> direction: WindowKind -> count: int -> unit
```
jaredpar: Should `direction` be renamed to `windowKind` here?

ricksladkey: It should!

Issue #1038, opened 2013-02-07
Support % macro in command line 


zarucki:
I added couple custom tools to the visual studio (Tools->External Tools...). 

In particular:
- command to open current file in gVim when VsVim is not enough ;) (if there's an easier way to do this please let me know, unfortunately % doesn't seem to have a current file path when running :!vim %)
- command to run Kdiff3 on current file against base from repository

The order tools are defined in External Tools... dialog is mapped to the commands in visualstudio:
- First command is Tools.ExternalCommand1, second is Tools.ExternalCommand2

When I press ctrl+/ and enter "> Tools.ExternalCommand1" it works correctly, but when I run :vsc Tools.ExternalCommand1 nothing happens. It could be related to the context, the command is run in. Maybe when running commands through :vsc it doesn't know which file is the current one?

ricksladkey:
This feature was implemented in PR #2229.

Issue #869, opened 2012-04-24
Support `%` and `#` special characters in the command line


m0tive:
VsVim doesn't replace `%` with the current filename. This really useful when wanting to open the current file in another editor (eg `:! gvim %`).

Filename modifiers like `%:h` and `%:r` would also be good.

http://vimdoc.sourceforge.net/htmldoc/cmdline.html#cmdline-special

ricksladkey:
This feature was implemented in PR #2229.

PR #2229, opened 2018-06-26
Parsing and applying filename modifiers in cd and lcd commands.

dhmgit:
#869 
#1038

http://vimdoc.sourceforge.net/htmldoc/cmdline.html#filename-modifiers

**Overview**
This PR is intended to open a discussion on the best way to approach the above referenced issues.  My initial motivation for supporting symbolic paths was the desire to use `:autocmd` along with `:lcd %:p:h` to keep my local current directory in line with the file I was currently editing.  I think it would be relatively easy to apply the symbolic path interpretation to other commands as soon as the details are worked out, but this PR only applies it to :cd and :lcd.

**Functional Change**
Vim seems to consider the filename symbol '%' to be equivalent to the '%' register.  VsVim calculates the '%' register as the current filename (sans directory path), whereas gVim appears to calculate it as the current full file path minus any prefix it has in common with the current directory.  I.e. if the current directory is `c:\a\b\x` and the current file path is `c:\a\b\y\test.txt` then the '%' register will be `y\test.txt` in gVim and `test.txt` in VsVim.  In order to maintain consistency between the register and the filename symbol, I changed the '%' register calculator to be consistent with the observed gVim behavior.  I'm not sure if this would be considered an undesirable functional change.

**Architectural Concern**
The line between parsing and interpreting gets a little blurry where we convert a symbolic path string to a physical path string.  In the PR, I opted to pass IVimBufferData to the Parser to give it the required context to apply the filename-modifier sequence internally and return the physical path string to the Interpreter.  This seems like a lot of context data to be passing to a parser, but it was more or less in line with passing IVimData to allow Parser.ParseAtCommand to convert `@:` to the previously run command.  

Some alternatives:
1. Return a type-safe representation of the parsed symbolic path from the Parser and let the Interpreter replace the symbols with physical path components.  E.g. `%:p:h\test.txt` would parse as something like `[Filename;FullPathModifier;HeadModifier;LiteralText]` and then the Interpreter could convert this to the physical path string.  This is the most robust design I can think of, but it would also require the most work, and may be overkill.
2. Return the raw symbolic path from the Parser and let the Interpreter invoke the Parser again to convert it to a physical path, passing the additional contextual information (filename/file path) as function arguments.  The disadvantage here is that every command that accepts a symbolic path would need to be "known" by the Interpreter.  It would also require the Interpreter create a new Parser to "re-parse" the raw symbolic path.  Furthermore, adding support for more symbols (e.g. '#') would require expanding the number of context arguments that need to be passed back into the Parser.

**Notes**
- There are few quirks in the gVim behavior that don't quite align with the documentation.  I erred on the side of the the gVim behavior rather than a strict interpretation of the doc.


ricksladkey:
> In the PR, I opted to pass IVimBufferData to the Parser to give it the required context to apply the filename-modifier sequence internally and return the physical path string to the Interpreter.

If you look in PR #2139, you'll see that I implemented a "file history" just like "command history" and this information is stored in `IVimData`. The piece of information you want is just the zeroth item in the files list, so issuing `:files<CR>` when the current buffer is `Interpreter_Parser` produces, e.g.:

```
      # file history
      0 C:\Build\VsVim\Src\VimCore\Interpreter_Parser.fs
      1 C:\Build\VsVim\Src\VimCore\Vim.fs
      2 C:\Build\VsVim\Src\VimCore\Modes_Insert_InsertMode.fs
      3 C:\Build\VsVim\Src\VimCore\CommonOperations.fs
      4 C:\Build\VsVim\Src\VimCore\Interpreter_Interpreter.fs
      5 C:\Build\VsVim\Test\VimCoreTest\CommandUtilTest.cs
```

This makes it unnecessary to pass `IVimBufferData` to the interpreter.

The upshot is that if we could build your PR off of my PR, then we would get both `#` support and `%` support with modifiers, which would be great, and it solves your problem of how to get access to the necessary information.

dhmgit:
After submitting the original PR, my sense that passing the IVimBufferData to the Parser was wrong continued to nag at me.  I implemented alternative 1) from my previous comment (which ended up being easier than expected) and added a commit to the PR.  I like the overall SoC much better after this change.

@ricksladkey, I only saw your comment after pushing my second commit, but I think your suggestion will still be relevant.  I wanted a way of supporting '#' as well, and it looks like the file history you added would do the trick.  Thanks for the suggestion and I'll take a look at your PR soon.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4300,6 +4300,12 @@ and IVimBufferData =
     /// The current directory for this particular window
     abstract CurrentDirectory: string option with get, set
 
+    /// The current (rooted) file path for this buffer
+    abstract CurrentFilePath : string
+
+    /// The current file name for this buffer. This is CurrentFilePath, excluding any prefix shared with CurrentDirectory.
+    abstract CurrentFileName : string
```
jaredpar: These should both return `string option`. Not all instances of `IVimBufferData` will end up having a path. 

Still thinking about whether this is the right place to expose file paths. Up until now its mostly been hidden behind `IVimHost` and methods like `Save`. Think this is probably the right place though.

dhmgit: @jaredpar , do you know of steps to create a buffer without a file path in VS?  Although it makes logical sense to change the properties to `string option`, any newly created document in VS seems to have a backing file in the temp directory, e.g. `~vs6503.cs`.  What's worse is that despite significant time spent searching through VS extensibility docs, I can't find any supported way of determining whether not a file has been saved at least once.  I suppose I could check whether the file is in a known temp directory and matches the '~vs' pattern, but this is isn't exactly bulletproof.

Let me know if you decide you want me to rewrite the `IVimBufferData.CurrentFile*` properties as `IVim` or `IVimHost` functions.  Either would be an easy change as both interfaces are accessible via `IVimBufferData` properties.  I don't think I realized early on that they were just a 'dot' or two away.

dhmgit: Re: CurrentRelativeFileName, I agree we should probably favor .NET's naming conventions over Vim's.  Vim calls the '%' and '#' symbolic names "filename", but this actually caused me a lot of confusion until I figured out that it's not always just the file name -- it's the full file path minus any common prefix it shares with the current directory.  However, CurrentFilePathWithoutCurrentDirectorySharedPrefix is a bit absurd, so CurrentRelativeFileName it is!

jaredpar: > do you know of steps to create a buffer without a file path in VS?

There are a couple of examples of such buffers. Most notably is the `ITextBuffer` that is created to edit conditional break points in managed code. 

> any newly created document in VS seems to have a backing file in the temp directory, e.g.

Indeed that will be the case for documents, or rather anything that can be saved, in VS. Their implementation is very tied to the file system and it always has a path in that case.

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -1644,9 +1647,10 @@ type VimInterpreter
             | Some substituteData -> substituteData.SearchPattern, substituteData.Substitute
         x.RunSubstitute lineRange pattern replace flags 
 
-    member x.RunTabNew filePath = 
-        let filePath = filePath |> OptionUtil.getOrDefault ""
-        _vimHost.LoadFileIntoNewWindow filePath (Some 0) None |> ignore
+    member x.RunTabNew symbolicPath = 
+        let filePath = x.InterpretSymbolicPath symbolicPath
+        let resolvedFilePath = x.ResolveVimPath filePath
```
ricksladkey: Great! This was on my radar.

dhmgit:
@ricksladkey I have the following in my .vsvimrc

`autocmd BufEnter,BufWinEnter * cd %:p:h`

Here's the Vim Tip that lead me to investigate this feature: http://vim.wikia.com/wiki/Set_working_directory_to_the_current_file

I had originally started to implement the 'autochdir' setting, but the linked article suggests that using autocmd is the preferred way of tracking the current directory.

PR #2230, opened 2018-06-27
Effective change

### LINKED ISSUE #664: Completion with intellisense not repeatable.

### LINKED ISSUE #1353: Repeat command with . is inaccurate sometimes
I am using ReSharper annotations and am inserting [NotNull] before parameters
I have the following code:

private static bool CertificateExists(
    X509Store certificateStore,
    string issuerName)

I put the cursor on X in the X509Store. 
I hit i[NotNull]  escape (I do not use autocomplete or anything like that, I type all the  text)
I move the cursor one line down to the s in string and hit . to repeat the command.
This is the result:

private static bool CertificateExists(
    [NotNull] X509Store certificateStore,
    [_NotNul] string issuerName)

Note the _ and the missing l at the end of NotNull.

I have seen this happen in other places but  not everywhere.  I can reproduce this scenario every time though.

### LINKED ISSUE #2220: repeat of change command does not match observed behavior in vim (example with 2cl)
The issue:

In vsvim it seams that the dot command isn't not repeating the insert with a change command.
Example below
---
In this example we are replacing "->" with "."

**In vsvim:**

```
foo1->bar(); // while on the "-" character "2cl" followed by "." yields "foo.bar();"
foo2->bar(); // while on the "-" character the repeat command "." yield "foobar();" 
```

**In vim:**
```
foo1->bar(); // while on the "-" character "2cl" followed by "." yields "foo.bar();"
foo2->bar(); // while on the "-" character the repeat command "." yields "foo.bar();"
```
---
Thanks again to everyone involved in this project. I really appreciate vsvim.
If this example is inadequate let me know.

ricksladkey:
#### Changes

- Extend the text change tracker to support parallel tracking of the effective change
- When repeating commands use the effective change if applicable
- Add a context sensitive command flag for commands that should not use the effective change
- Upon leaving insert mode, move the caret after an effective change to account for auto-inserted characters (reverted after review feedback)
- Implement `<LeftMouse>` etc. for insert mode and `<LeftDrag>` for select mode and visual mode (moved to another PR)

#### Fixes

- Fixes #664
- Fixes #1353
- Fixes #2220

#### Discussion

The inability to reliably repeat change commands with ReSharper and Visual Assist is a major pain point for many VsVim users. Even users without those code assistants cannot repeat changes with auto-inserted characters like matching parentheses. This change attempts to fix the 90% case (the daily use case) of a change made on a single line where the user can plainly see what the previous version changed and wants the repeat to change it to that very same text.

VsVim already has an option called `:set vsvimatomicinsert` but it is flawed conceptually, violates the spirit of many vim principles, and sadly doesn't even always work. This change attempts to get the best of both worlds: reliable repeats and good fidelity with vim semantics.

Here is how it works. The existing text change tracker now also can track the effective change. The active insert region starts out empty where the caret is. As insert mode continues, changes to the active insert region cause it to grow or shrink. As long as the active insert region stays intact, and as long as the user didn't issue any context-sensitive insert commands (like tab or enter), the effective change consists precisely of the text contents of the active insert region.

As a result, code assistants can perform arbitrarily complex edits and caret movements behind the scenes and as long as they confine their changes and movements to the active insert region, we can use the buffer itself as the reference point to answer the question "What got inserted?" And because `<MouseLeft>` is now implemented, we can discriminate between the user using the mouse and a code assistant programmatically changing the caret position to decide whether to break the undo sequence.

I have tested this change with ReSharper and it fixes the "repeating changes" problem. From the bug reports, I am fairly confident it will fix the related problems with Visual Assist. It even fixes repeating inserts or changes that involve auto-inserted characters, a problem that can be demonstrated using only Visual Studio itself. I also included a detailed step-by-step ReSharper completion test showing exactly what kind of edits must handled for the repeat command to work correctly with ReSharper.

### DIFF
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -536,6 +536,15 @@ type internal InsertMode
         // edit data for the session
         _textChangeTracker.CompleteChange()
 
+        // If available, use the effective change for the edit.
+        match _textChangeTracker.EffectiveChange with
+        | Some textChange ->
+            InsertCommand.OfTextChange textChange
+            |> Some
+            |> x.ChangeCombinedEditCommand
+        | None ->
+            ()
+
         try
             match _sessionData.InsertKind, _sessionData.CombinedEditCommand with
             | InsertKind.Normal, _ -> ()
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -542,30 +542,29 @@ type internal InsertMode
             && _textChangeTracker.IsEffectiveChangeInsert
         then
             match _textChangeTracker.EffectiveChange with
-            | Some span ->
-                if span.Length <> 0 then
+            | Some span when span.Length <> 0 ->
 
-                    // Override the piecewise combined edit command with
-                    // the effective change which is a simple insertion.
-                    span
-                    |> SnapshotSpanUtil.GetText
-                    |> TextChange.Insert
-                    |> InsertCommand.OfTextChange
-                    |> Some
-                    |> x.ChangeCombinedEditCommand
+                // Override the piecewise combined edit command with
+                // the effective change which is a simple insertion.
+                span
+                |> SnapshotSpanUtil.GetText
+                |> TextChange.Insert
+                |> InsertCommand.OfTextChange
+                |> Some
+                |> x.ChangeCombinedEditCommand
 
-                    // If the effective change is confined to one line then
-                    // move the caret to the end of the effective change.
-                    // This can happen if the editor auto-inserted matching
-                    // parentheses, quotes, brackets, etc. and the user
-                    // didn't "finish" them manually. We want the caret
-                    // and the final edit point to appear as if the user
-                    // had entered them directly.
-                    let startingLineNumber = span.Start |> SnapshotPointUtil.GetLineNumber
-                    let endingLineNumber = span.End |> SnapshotPointUtil.GetLineNumber
-                    if startingLineNumber = endingLineNumber then
-                        _operations.MoveCaretToPoint span.End ViewFlags.None
-            | None ->
+                // If the effective change is confined to one line then
+                // move the caret to the end of the effective change.
+                // This can happen if the editor auto-inserted matching
+                // parentheses, quotes, brackets, etc. and the user
+                // didn't "finish" them manually. We want the caret
+                // and the final edit point to appear as if the user
+                // had entered them directly.
+                let startingLineNumber = span.Start |> SnapshotPointUtil.GetLineNumber
+                let endingLineNumber = span.End |> SnapshotPointUtil.GetLineNumber
+                if startingLineNumber = endingLineNumber then
+                    _operations.MoveCaretToPoint span.End ViewFlags.None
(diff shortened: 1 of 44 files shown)


Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -528,6 +534,37 @@ type internal InsertMode
...
+                |> Some
+                |> x.ChangeCombinedEditCommand
+
+                // If the effective change is confined to one line then
+                // move the caret to the end of the effective change.
+                // This can happen if the editor auto-inserted matching
+                // parentheses, quotes, brackets, etc. and the user
+                // didn't "finish" them manually. We want the caret
+                // and the final edit point to appear as if the user
+                // had entered them directly.
```
jaredpar: This part makes me a bit nervous. The IDE puts a lot of work into the caret position around auto-completion of parens and such. Moving the caret here can end up subverting expectations of users here.

ricksladkey: Mode-less editors like Visual Studio must necessarily assume the user might want to "add some more text". But VsVim has the luxury of the user giving us a definitive indication that insert mode is over by typing the escape key.

Let me pose a question. Without this change, if the user issues a change command and inserts `WriteLine("Hello, world!")` where the closing quote and closing parenthesis were auto-inserted, after typing escape the cursor would rest on the exclamation point. If that change command is then repeated, the cursor would be now positioned on the closing parenthesis, which seems counter-intuitive.

To put it another way, in vim, after any insertion that doesn't break the undo sequence, the cursor is *always* at the end of the text that got inserted.

But, to put this in context, I agree that the guiding principle should be *not* to subvert the user's expectations, if possible. It's hard to know how users will react, but I can see an argument for both points of view. One thing I am sure of is that when the change is repeated, the user *definitely* wants those auto-inserted characters to be repeated as if they had typed them.

TKumadaki:
I am sorry if it was a misunderstanding.

I tried using vsvim which compiled the code of ricksladkey.
The behavior when double-clicking a mouse against a word has become different from before.
In the past, when you double-click the mouse on a word, the word was selected.
However, in this version, it was only selected up to the position where the mouse was double-clicked.

Is this the intended behavior?

ricksladkey:
@TKumadaki Thanks for spotting this.

Sigh, no, that is not the intended behavior. I implemented a mouse handler for "left mouse down" to fix one problem (telling the difference between the user using the mouse and a code assistant programmatically positioning the caret) and it leads to a whole cascade of additional mouse behavior that needs to be emulated. Basically, once we handle "left mouse down" to move the cursor, left mouse dragging, left mouse double-clicking, and who knows what else also have to be implemented

I think it would be a good thing to do that work eventually, but I am not prepared to do it right now and so I am going to separate out the mouse changes from the "effective change" commits.

Src/VimCore/Modes_Insert_InsertMode.fs
ricksladkey: I reverted this as it is only a minor feature. Moving the caret in that case makes sense to *me*, but the important thing is the change that gets repeated.

PR #2232, opened 2018-06-30
Master

TKumadaki:
This is experimental code.
I do not want merge.

I wanted to know if C # scripts could be executed in VsVim.

I put the file "Hello.csx" in the user profile directory.

//**********************************************************

//Hello.csx

using System;

StatusUtil.OnError("Hello! VsVim");

//**********************************************************

After that, I typed "Hello" in VsVim's command mode

And confirmed that "Hello! VsVim" was displayed in the command margin.

Is it possible to incorporate C # script execution function into vsvim?

### DIFF
--- a/Src/VimCore/Interpreter_Expression.fs
+++ b/Src/VimCore/Interpreter_Expression.fs
@@ -635,6 +635,8 @@ and [<RequireQualifiedAccess>] LineCommand =
     ///  - new tabstop value
     | Retab of LineRangeSpecifier * bool * int option
 
+    | RunCSharpScript of string
+
     /// Process the 'set' command
     | Set of SetArgument list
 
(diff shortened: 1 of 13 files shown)


jaredpar:
> Is it possible to incorporate C # script execution function into vsvim?

Technically correct but it requires a lot of complexity. C# scripting didn't become a part of Visual Studio until VS2015 while VsVim is supported back to VS2012. Hence we can't use it directly in the lower levels of VsVim but would have to add it to one of the VsSpecific layers and indirectly call into it. That will require a bit of work.

Additionally we can't move the target framework to net461 yet. The required runtime for VS2012 is still net45 hence can't go above that without risking breaking customers. I am somewhat open to changing that though. The percentage of users in the wild who aren't on at least net46 is really low. Probably reasonable to ask people to upgrade to at least that.

> I wanted to know if C # scripts could be executed in VsVim.

Is there a reason you want to do this though? A lacking in the current experience you think could be solved with C# scripting integration?

TKumadaki:
I was not thinking about the VS 2012 user.

The full support of VIM script was that it was not the goal of the project.

I thought that the C # script execution environment would replace the VIM script.

but I do not understand how much C # script can handle VSVIM.

And I do not understand how to correctly incorporate the C # execution environment.

TKumadaki:
This is experimental code.
Close pull request

PR #2234, opened 2018-07-01
Fix g8 display to use hex

jaredpar:
Changes the g8 command to both:

- Display values in hex not decimal
- Display surrogate pairs properly

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -7,6 +7,7 @@ open Microsoft.VisualStudio.Text.Operations
 open Microsoft.VisualStudio.Text.Editor
 open Microsoft.VisualStudio.Text.Outlining
 open Microsoft.VisualStudio.Text.Formatting
+open System.Text
 open RegexPatternUtil
 open VimCoreExtensions
 open ITextEditExtensions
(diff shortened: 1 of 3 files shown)


Src/VimCore/UndoRedoOperations.fs
```diff
@@ -352,13 +352,9 @@ and UndoRedoOperations
     ///     - transactions we orphaned in ResetState because of detected errors
     ///     - transactions closed out of order 
     member x.LinkedUndoTransactionClosed (linkedUndoTransaction: LinkedUndoTransaction) = 
-        let canBeEmpty = Util.IsFlagSet linkedUndoTransaction.Flags LinkedUndoTransactionFlags.CanBeEmpty
-        let endsWithInsert = Util.IsFlagSet linkedUndoTransaction.Flags LinkedUndoTransactionFlags.EndsWithInsert
```
jaredpar: Whoops. Looks like I merged a change wrong here 😦

Src/VimCore/CommandUtil.fs
```diff
@@ -873,11 +874,18 @@ type internal CommandUtil
...
     member x.DisplayCharacterBytes () =
-        match SnapshotPointUtil.TryGetChar x.CaretPoint with
-        | None -> _commonOperations.Beep()
-        | Some c ->
-            let str = sprintf "%d" (int c)
-            _statusUtil.OnStatus str
+        if SnapshotPointUtil.IsEndPoint x.CaretPoint then 
+            _commonOperations.Beep()
+        else
+            let span = SnapshotCharacterSpan(x.CaretPoint)
```
ricksladkey: It makes me happy to see the character span type being put to good use!

jaredpar: Indeed. I've found a few other uses for it, and it's underlying logic, in other features like `ga`.

PR #2238, opened 2018-07-03
Mark glyphs, change/yank marks and change/yank notifications

### LINKED ISSUE #574: Feature Request: Visualize VsVim markers in a similar way like Visual Studio bookmarks
It would be really nice to have more visual feedback (like described in issue #408), but this time it's about having visual feedback not about the result of executed actions or commands, but about the position of markers.

While Visual Studio bookmarks are visualized by "blue" blocks left to the line numbers (if shown), it would be nice to have a similar visualization of the markers with e.g. "orange" blocks containing the marker "name" (a-z, A-Z, etc.).
Even the markers "<" and ">" and some of the other special markers (as supported by VsVim) should be shown.

### LINKED ISSUE #575: Feature Request: Extend markers functionality to be closer to VIM
Using in VIM `:help markers` tells a lot about the functionality of marks/markers in VIM.
Some of them I never used, some of them I didn't miss, because I already forgot them again, some of them I really miss in VsVim.

Some functionality worth porting IMHO:
1. Uppercase markers being persistent, causing, if used, the opening of the remembered file and a jump to the remembered location 
2. Special _markers_ like:
   - changes `'[`, `']`, `'.`, `'^`
   - positions `''` (needs a jump list recallable with :jumps), `'"`, `['`, `]'`

And all `'` things are useful with `, too!

### LINKED ISSUE #1151: Support `[ and `] to jump to beginning/end of previously changed or yanked text

### LINKED ISSUE #2083: Is possible to list marks when pressing ' key?
I was looking for somehow list marks in Vim when I found [this ](https://vi.stackexchange.com/questions/8451/is-it-possible-to-have-vim-show-the-list-of-available-marks-when-using-marks)on SO. I created a .vimrc file in my %USERPROFILE% folder with the following functions as described on the link above.

""""""""""""""""""""""""""""
" Ranger style marks command
"
""""""""""""""""""""""""""""
function! Marks()
    marks
    echo('Mark: ')

    " getchar() - prompts user for a single character and returns the chars
    " ascii representation
    " nr2char() - converts ASCII `NUMBER TO CHAR'

    let s:mark = nr2char(getchar())
    " remove the `press any key prompt'
    redraw

    " build a string which uses the `normal' command plus the var holding the
    " mark - then eval it.
    execute "normal! '" . s:mark
endfunction

nnoremap ' :call Marks()<CR>

So when I press ' I should get a list with all marks right?
Seems like some functionality isn't supported by VsVim as I get the following output:
:call to function Marks not supported

I am not a Vim guru by the way. Does anyone knows if it is possible to get this list? How?

### LINKED ISSUE #2182: Marks set on twice deleted lines don't reappear on undo
If you set a mark on a line, then delete that line, and then undo, the mark will disappear and then reappear. If you then delete it again and undo again, the mark will *not* reappear. This is because the reiterated snapshot version is the oldest version number of the snapshot corresponding to that exact state, and won't compare with equality to later snapshot versions (always monotonically increasing) with identical contents.

ricksladkey:
#### Changes

- Notate vim marks in the indicator margin
- Add local setting `hidemarks` to prevent certain marks from being notated
- Implement last change/yank marks ``` `[``` `'[` and ``` `]``` `']`
- Add status line notifications reporting the number of deleted/changed/yanked lines

#### Fixes

- Fixes #574
- Fixes #575
- Fixes #1151
- Fixes #2083
- Fixes #2182

#### Discussion

Visual Studio provides a gutter called the indicator margin, which is an ideal place for vim marks to be notated. This is the same place that breakpoint glyphs are displayed. Rather than display a fancy bookmark, this change uses a minimalist approach of simply displaying the mark character or characters on a line in the indicator margin. There is enough room to display up to three per line and the order chosen is ASCII sort order.

Last change/yank marks are convenient for motions and for reselecting recently put or yanked text. At a high level, insertion is a change and the marks reflect the beginning and end of the most recently inserted text. Yanking is also straightforward and uses the same marks. All other changes when atomic are easily tracked using the changed event for the text buffer. The main deficiency is when a sequence of changes are made and then the automatic tracking only records the last change.

Once we have centralized change/yank tracking, it then also straightforward to add vim-style "3 lines yanked" notifications. This can be very helpful to easily supply an accurate count to a subsequent edit command such as indenting, rather than reselecting the changed or yanked text or guessing how many lines were involved.

The original implementation of the mark glyph tagger attempted to use the `IBasicTaggerSource` infrastructure, but for reasons unknown, a glyph tagger apparently acts on the `VisualSnapshot` (elision buffer) vs. the `TextSnapshot` (normal text buffer). While this appeared to work, it caused large numbers of caught exceptions because the VsVim tagger infrastructure is not intended to work with the visual snapshot.

Several existing special marks previously worked correctly but visually didn't update as the user would expect during live editing. For example, the last insert point was not set until switching out of insert mode. These marks still work as they used to work, but are kept up-to-date as editing proceeds.


dhmgit:
I'm excited about this potential change, so I did a bit of testing in my local dog food branch.  It works very well overall with just a couple of issues that I noticed.  Note that these scenarios also affect other marks (letters, yank, etc.) but I used visual selection for the examples as it is (unsurprisingly) easiest to illustrate visually.

### Visual Selection Terminating on Blank Lines
The open and/or close mark for visual block/line selection will not be shown if the selection begins or ends on a blank line.  The `:marks` look mostly correct, so the problem appears to be with the drawing of the glyph and not the calculation of the mark positions.

**Line 1**
image
 <     24    0 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs
 &gt;     25   66 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs

**Line 2**
image
 <     24    0 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs
 &gt;     25   66 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs

**Block**
image
 <     20    0 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs
 &gt;     25   21 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs

**Character**
image
 <     25    8 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs
 &gt;     26    0 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs

### Mouse Selection
In this scenario, it looks like the glyphs and the `:marks` are both incorrect and inconsistent with one another.  However, using the keyboard to edit the visual selection will immediately correct both.

image
 <     25    8 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs
 &gt;     27    0 C:\git\dhm\VsVim\Src\VimWpf\Implementation\MarkGlyph\MarkGlyphTagger.cs

ricksladkey:
@dhmgit Thanks for giving this a test drive! I introduced the "no glyphs on empty lines" bug trying to fix another problem (stale marks), but then forgot to revert it after finding the true cause.

I took a crack at keeping selection marks up to date during mouse dragging and boy is it complicated. There are a lot of moving pieces! First, the type of selection I use for the mouse, "select mode" doesn't record selection marks at all, which is just a bug. Second, the selection mode, the selection change tracker and the selection tracker all have their hands in the selection. I got it to work, but broke the unit tests. I will give it another try tomorrow.

ricksladkey:
@dhmgit I moved the recording of the visual selection into the selection tracker and now selection marks update in real time for both keyboard and mouse in both visual and select modes.

dhmgit:
@ricksladkey, I retested with your latest changes and it looks like the issues I listed above have been resolved.  Very nice!

jaredpar:
This is awesome! I've wanted this for a long time. 

I just got back from vacation and am catching up on things. Will try and get this reviewed ASAP.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4532,8 +4534,19 @@ and SwitchModeEventArgs
...
+
+    member x.Mark = _mark
+    member x.TextBuffer = _textBuffer
+
+    override x.ToString() = _mark.ToString()
+
 and IMarkMap =
 
+    /// Raise the mark set event
+    abstract RaiseMarkSet: mark: Mark -> textBuffer: ITextBuffer -> unit
```
jaredpar: This method feels like a bit of a leaky abstraction on where we chose to store the various mark values. I'm wondering if we need to centralize mark storage into local marks (per buffer) and global. Then have the mark changed events on those locations. That would in theory remove the need for this function and let the owner of the data raise the event when it changes.

Src/VimCore/MefComponents.fs
```diff
@@ -72,7 +72,9 @@ type internal TrackingLineColumn
     /// Update the internal tracking information based on the new ITextSnapshot
     member x.OnBufferChanged (e: TextContentChangedEventArgs) =
         match _line with 
-        | Some snapshotLine -> x.AdjustForChange snapshotLine e
+        | Some snapshotLine ->
+            if e.AfterVersion <> snapshotLine.Snapshot.Version then
```
jaredpar: Is this optimizing for the case of avoiding calculations on an undo / redo?

Src/VimCore/Vim.fs
```diff
@@ -242,7 +242,7 @@ type internal VimBufferFactory
         let textBuffer = textView.TextBuffer
         let statusUtil = _statusUtilFactory.GetStatusUtilForView textView
         let localSettings = vimTextBuffer.LocalSettings
-        let jumpList = JumpList(textView, _bufferTrackingService) :> IJumpList
+        let jumpList = JumpList( textView, _bufferTrackingService, vim.MarkMap) :> IJumpList
```
jaredpar: Nit: extra space after `(`

Src/VimCore/VimSettingsInterface.fs
```diff
@@ -62,6 +62,7 @@ module LocalSettingNames =
     let QuoteEscapeName = "quoteescape"
     let EndOfLineName = "endofline"
     let FixEndOfLineName = "fixendofline"
+    let HideMarksName = "hidemarks"
```
jaredpar: Should use name `"vsvimhidemarks"` as this is a non-standard setting.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4565,6 +4581,14 @@ and IMarkMap =
     /// Delete all of the global marks 
     abstract Clear: unit -> unit
 
+    /// Raised when a mark is set
+    [<CLIEvent>]
+    abstract MarkSet: IDelegateEvent<System.EventHandler<MarkChangedEventArgs>>
```
jaredpar: Did you consider `MarkChanged` instead of `MarkSet`? The implementation in this PR faithfully fires on any set, even redundant ones, which I would expect based on the name. I've hit a few cases before where this can end up in stack overflows where inside a `MarkSet` event handler the value is set to the existing value, triggers another event and then it recurses until overflowing the stack.

I don't have a strong preference here though.

Src/VimCore/CommonOperations.fs
```diff
@@ -1586,6 +1586,34 @@ type internal CommonOperations
     member x.RaiseSearchResultMessage searchResult = 
         CommonUtil.RaiseSearchResultMessage _statusUtil searchResult
 
+    /// Record last change or yank start and end positions
+    member x.RecordLastChangeOrYank oldSpan newSpan =
```
jaredpar: I think we should consider breaking this into two methods: RecordLastChange and RecordLastYank. The meaning of the input differs significantly based on the scenario. 

One pattern that usually causes me concern is when there are multiple snapshot inputs to a method (`SnapshotSpan`, `SnapshotPoint`, etc ...) where sometimes they are on the same snapshot and other times they aren't. Can lead to cases where code works great in a few scenarios but then falls down where it hits the "other" case. Sometimes it's necessary to have this in an API but I try to avoid it if possible.

Splitting it up would seem to mean that:

- Yank always has one span 
- Delete has two and they're never on the same snapshot.

Src/VimWpf/Implementation/MarkGlyph/MarkGlyphTagger.cs
```diff
@@ -0,0 +1,318 @@
...
+using System.Linq;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Tagging;
+using Vim.Extensions;
+
+namespace Vim.UI.Wpf.Implementation.MarkGlyph
+{
+    internal sealed class MarkGlyphTagger : ITagger<MarkGlyphTag>, IDisposable
+    {
+        static ReadOnlyCollection<ITagSpan<MarkGlyphTag>> s_emptyTagList =
```
jaredpar: `private readonly`

Src/VimWpf/Implementation/MarkGlyph/MarkGlyphFactoryProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Vim.UI.Wpf.Implementation.MarkGlyph
+{
+    [Export(typeof(IGlyphFactoryProvider))]
+    [ContentType(VimConstants.AnyContentType)]
+    [TextViewRole(PredefinedTextViewRoles.Editable)]
+    [TagType(typeof(MarkGlyphTag))]
+    [Name("MarkGlyph")]
+    [Order(After = "VsTextMarker")]
+    class MarkGlyphFactoryProvider : IGlyphFactoryProvider
```
jaredpar: `internal sealed`

Src/VimWpf/Implementation/MarkGlyph/MarkGlyphTagger.cs
```diff
@@ -0,0 +1,318 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Tagging;
+using Vim.Extensions;
+
+namespace Vim.UI.Wpf.Implementation.MarkGlyph
+{
+    internal sealed class MarkGlyphTagger : ITagger<MarkGlyphTag>, IDisposable
```
jaredpar: Is there a reason you didn't use on of the tagger interface types here: in particular `BasicTaggerSource`? Those base types were written to handle all of the unwritten rules around taggers and handle most of them automatically. I've never written a glyph tagger though so it's possible those don't apply here.

jaredpar:
Made it through the `VimCore` layer today. Will try and pick up the glpyh layer tomorrow.

Src/VimCore/CoreInterfaces.fs
ricksladkey: My precedent for this was `VimBuffer.RaiseErrorMessage` which is used only by `StatusUtil` in a "friendly" fashion.

The only two external callers of the `RaiseMarkSet` function are the vim text buffer and the jump list and we would limit the callers who could call it I knew how!

It's true that the vim text buffer could raise its own separate event, which is arguably better anyway because the consumer is already tied to a vim buffer (one mark glyph tagger per buffer).

Then the jump list could delegate the storage of the "last jump location" to the vim text buffer.

Src/VimCore/MefComponents.fs
ricksladkey: When marks are set *during* the `OnBufferChanged` event, they will be already be correct for the current snapshot. When there are multiple listeners to that event, we don't know the order in which they will be called. So it can easily happen (and I saw it during debugging) that an listener who is called early and creates a mark (like the `TextChangeTracker`) can set a mark and then a later listener (like the `TrackingLineColumn`) can try to track that very mark.

So it's not just an optimization because if the snapshot line is already up-to-date with the after version we would corrupt the position of the mark by trying to adjust it.

Src/VimCore/VimSettingsInterface.fs
ricksladkey: I agree, but there are two conventions for the name: `vsvimsetting` (e.g. `vsvimatomicinsert`) and `vsvim_setting` (e.g. `vsvim_useeditordefaults`). I'm thinking the latter is better.

Src/VimCore/CommonOperations.fs
ricksladkey: Ha, ha, I tried to get clever! I used to have just one span (the new span) and a bool `isYank`. Then I realized I needed the old span to detect deletes. Then I thought I could delete the bool by using the spans themselves to represent yanks as oldSpan = newSpan.

I agree now that it is too obtuse.

Src/VimWpf/Implementation/MarkGlyph/MarkGlyphTagger.cs
ricksladkey: I discussed this in the "Notes" section of the PR description, which I will paste here:

> The original implementation of the mark glyph tagger attempted to use the `IBasicTaggerSource` infrastructure, but for reasons unknown, a glyph tagger apparently acts on the `VisualSnapshot` (elision buffer) vs. the `TextSnapshot` (normal text buffer). While this appeared to work, it caused large numbers of caught exceptions because the VsVim tagger infrastructure is not intended to work with the visual snapshot.

Src/VimCore/CoreInterfaces.fs
ricksladkey: Special marks are set quite often and so I optimized the mark glyph tagger to not force a tags changed event when a mark is set but doesn't change.

Src/VimCore/CoreInterfaces.fs
ricksladkey: It turns out that the jump list is scoped to a window so we have three different kind of marks: global, local and window.

Src/VimWpf/Implementation/MarkGlyph/MarkGlyphTagger.cs
ricksladkey: I'm glad you asked because my conclusion was flawed as it is indeed possible to write a glyph tagger that operates on the normal text buffer and so I have reverted with success to the original implementation.

ricksladkey:
For the record, I also tried to write the mark glyph tagger in F# but was unable to because we lack references to the WPF layer from VimCore.

jaredpar:
> For the record, I also tried to write the mark glyph tagger in F# but was unable to because we lack references to the WPF layer from VimCore.

That's awkward at times but deliberate. The reason is VS for Mac. They have started adopting the same editor implementation as VS minus the WPF layer. Keeping WPF out of VimCore means that, at least in theory, VsVim could run in VS for Mac as well if a Xamarin specific host were put on top of it. 

This is something I've sketched out doing with some Xamarin team members before. Just never had time to follow through with it yet 😦

jaredpar:
Can't wait to try this out.

Chris-E-J-Ellis:
My VsVim auto-updated last week and this feature appeared, love it, great work!

Issue #1099, opened 2013-05-29
Provide a way to copy word under cursor to build search and replace string


bradphelan:
Under VIM we can do

```
:nnoremap ;s :%s/\<<C-r><C-w>\>/
```

to enable search and replace of the word under the cursor by pressing `;` key when over a word. However this doesn't work in VSVIM.

sqlkoala:
Btw, it is already possible to do this: you yank into a register and then access this register in the commandline. Should be something like this.
`nnoremap ;s "ryiw:%s/<C-r>r/`

I even used this to build a mapping that searches the current word (in a regex) within my visual studio project :D

shmup:
> Btw, it is already possible to do this: you yank into a register and then access this register in the commandline. Should be something like this. `nnoremap ;s "ryiw:%s/<C-r>r/`
> 
> I even used this to build a mapping that searches the current word (in a regex) within my visual studio project :D

This is what my vimrc uses for similar things, and I will try to adjust to something working from this thread. But I wanted to point out the technique:

```
nnoremap <space>s :'{,'}s/\<<C-r>=expand('<cword>')<cr>\>/
nnoremap <space>% :%s/\<<C-r>=expand('<cword>')<cr>\>/
```

It is a lovely mapping that I use often, hoping a variant here satisfies.

PR #2241, opened 2018-07-05
Command margin edit fixes

dhmgit:
Fixes: #1099 

### Other Issues Fixed
This PR also addresses a number of minor command margin editing issues that I've run into.  I thought it would be preferable to include these in one PR since they are so closely related to one another.

- Repeated typing of CTRL-R will insert an equal number of double-quote characters.
- ESC in paste-wait mode cancels command mode instead of just canceling paste-wait.
- CTRL-U leaves the cursor before the initial character.
- CTRL-W does not delete the word before the cursor.
- BS over initial character does nothing if there are characters after the cursor.

### Issues Not Fixed
There are a few additional issues with keyboard focus, which anyone who has worked in this area  of the code is probably familiar with.  These issues generally occur when the keyboard focus moves from the command margin TextBox to the editor so that keystrokes are processed directly by the VimBuffer.  At this point, a number of keyboard operations cease to function as expected.  These are operations that require knowledge of or control over the command margin's caret index, of which VimBuffer has neither.  I didn't attempt to fix these issues, as any robust solution that I could envision would involve heavy changes to the command margin, and possibly abandoning the use of the WPF TextBox entirely.

**Example**
1. Type `:ab<left><c-r>`
2. Observe `:a"b` displayed in the command margin.
3. Mouse-click in the document editor so that the command margin loses keyboard focus.
4. Observe `:ab"` displayed in the command margin.

### DIFF
--- a/Src/VimCore/IncrementalSearch.fs
+++ b/Src/VimCore/IncrementalSearch.fs
@@ -5,6 +5,7 @@ open Microsoft.VisualStudio.Text.Operations
 open Microsoft.VisualStudio.Text.Editor
 open Microsoft.VisualStudio.Text.Outlining
 open NullableUtil
+open Vim.VimCoreExtensions
 
 type IncrementalSearchData = { 
 
@@ -132,8 +133,8 @@ type internal IncrementalSearch
                 member this.Completed (data: ITrackingPoint) _ = runActive (fun session -> x.RunCompleted session data) IncrementalSearchData.Default.SearchResult
                 member this.Cancelled (data: ITrackingPoint) = runActive (fun session -> x.RunCancelled session) ()
             }
-
-        let historySession = HistoryUtil.CreateHistorySession historyClient startPoint StringUtil.Empty None
+        let vimBuffer = _vimBufferData.Vim.GetVimBuffer _textView
+        let historySession = HistoryUtil.CreateHistorySession historyClient startPoint StringUtil.Empty vimBuffer
         _incrementalSearchSession <- Some (IncrementalSearchSession(key, historySession, incrementalSearchData))
 
         // Raise the event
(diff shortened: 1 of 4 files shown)


ricksladkey:
These changes look good!

> BS over initial character does nothing if there are characters after the cursor.

It's not particularly crucial to me, but I added this behavior because it is how vim behaves, at least in my testing with gvim 8.0.

Regarding the issue of whether the command margin has the focus or its vim buffer, it is a historical accident that the vim buffer can process command line keys at all. In the original implementation, that was the way that *all* command margin input was handled. When the command margin was changed so that it gets the focus (which I think was a necessary change), a lot of features had to be "thrown over the wall" and dealt with in command margin land.

I did not try this, but my assumption was that if the vim buffer was in command mode and it received the keyboard focus, it could "forward" the focus to the command margin to fix the focus problems. There might be some problem with the approach that I'm not taking into account.

dhmgit:
@ricksladkey , regarding the BS behavior, I guess I never checked on the gVim behavior before making this change.  You're right that it ignores the BS if there are characters after the initial char.  However, the reason I "fixed" this was that the BS keypress was not getting marked as handled if the text length was > 1, which caused the caret to be placed before the initial char.  Once this occurs, chaos ensues.  E.g. typing `:a<left><bs>b` will result in `:b:a` being displayed in the command margin.  I'm quite happy to change this back to match the gVim behavior, with the addition of always handling the BS keypress. 
 I don't like to stray from the gVim behavior (even in cases where it seems illogical) unless there is a good reason.

I did play around a bit with forcing focus back to the command margin.  Even more chaos.  In my experience messing around with focus in Windows needs to be done with the utmost care, so this should probably be part of a separate effort.

ricksladkey:
To clarify, besides the other fixes, this PR implements `<C-r><C-w>` in command mode, which is awesome!

PR #2243, opened 2018-07-07
Support for 'virtualedit' in block, insert and normal modes

### LINKED ISSUE #1165: 'virtualedit=' extend options support
I'll be very happy to have 'all' option for this setting, or at least 'block' option if it is possible, thank you.

ricksladkey:
#### Changes

- Add 'use virtual space' property to vim text buffer
- Support virtual caret motions: char left/right without wrap, char left/right with wrap, line up/down
- Support insert mode virtual caret char left/right, line up/down
- Support block selection with virtual active/anchor points
- Fill in virtual spaces when putting text from a virtual caret position
- Add dual virtual/non-virtual APIs for visual selection, block span
- Extend virtual snapshot point/line/span utilities

#### Fixes

- Fixes #1165

#### Discussion

This change adds the foundation necessary to fully implement the `virtualedit` setting. The focus is on `virtualedit=block`, which is the most requested feature. However, motions for all modes are supported and the bulk of the remaining work is to finish support for character-wise visual selection.

The goal of the change is to have no effect when virtual space is not applicable. However, incorporating virtual editing into the visual selection required many, many changes to core APIs and data structures to use `VirtualSnapshotPoint` instead of `SnapshotPoint` so that the virtual caret position is kept intact.

To ease the transition, some APIs have virtual and non-virtual flavors. Existing callers can continue to use the non-virtual API but revised callers can supply a virtual caret position. Virtual and non-virtual state is converted as necessary so that even though an API or data structure is using virtual points internally, as long as the caret position is non-virtual, there is no change in the behavior.

As an example, you can now use the `:set virtualedit=all` option in the following text buffer:

```
cat
dog bat
bear
```

and issue the follow command to select a 3x9 block: `<C-v>2j8l` with the caret in virtual space.

If you then yank the block and move the caret and put the yanked block with `y8lp` the buffer contains:
```
cat     cat
dog bat dog bat
bear    bear
```

There is more work to be done, but this is a good intermediate stopping point that implements virtual caret motions and virtual block editing without breaking non-virtual editing.


ricksladkey:
@jaredpar I have implemented virtual edit for visual character mode, but it basically doubles the size of the change. I can hold off on that feature and leave this PR as it is (quite big already) or include that feature in this PR and implement all of `virtualedit` in one go.

Unless you disagree, I'm assuming it would be better (easier to review) this PR as is and save visual character mode for another PR when and if this PR gets merged.

jaredpar:
> Unless you disagree, I'm assuming it would be better (easier to review) this PR as is and save visual character mode for another PR when and if this PR gets merged

I agree. I think splitting it up is better. Let me try and dive into this tonight.

Src/VimCore/CommandRunner.fsi
```diff
@@ -8,5 +8,5 @@ open Vim
 
 type internal CommandRunner =
     interface ICommandRunner
-    new: ITextView * IRegisterMap * IMotionCapture * IVimLocalSettings * ICommandUtil * IStatusUtil * VisualKind * KeyRemapMode -> CommandRunner
+    new: IVimBufferData * IMotionCapture * ICommandUtil * VisualKind * KeyRemapMode -> CommandRunner
```
jaredpar: I'm guessing there are a lot of small but useful simplifications like this we could make now. Most of these ctors were written before I'd worked out types like `IVimBufferData` and was far into mock based testing.

Src/VimCore/CommandUtil.fs
```diff
@@ -1836,31 +1842,36 @@ type internal CommandUtil
 
                 // Now run the put after command
                 let register = x.GetRegister (Some RegisterName.Unnamed)
-                x.PutAfterCaretCore register.RegisterValue 1 false
+                x.EditWithUndoTransaction "Put after mouse" (fun () ->
+                    x.PutAfterCaretCore register.RegisterValue 1 false)
```
jaredpar: Why did we need to switch to explicit transactions in this set of APIs?

jaredpar:
This PR is big but also really easy to follow. Had one small Q.

Src/VimCore/CommandUtil.fs
ricksladkey: This is because when you put from virtual space, we have to fill in virtual spaces first, otherwise the put gets jammed onto the end of the line it was on. So we we want both the real spaces added (if any) and the put text to be part of the same transaction.

Src/VimCore/CommandRunner.fsi
ricksladkey: I specifically needed a centralized place to put the `UseVirtualSpace` property and it went in `VimTextBuffer` which is the lowest level at which we can combine the settings with the mode to answer the question "Should I be using virtual space right now?" And I was sure that `VimTextBuffer` was OK to pass and many similar services use it. However, once I added `VimTextBuffer`, it became clear that other resources were now directly accessible without separate parameters.

I try to avoid gratuitous changes, but adding `VimTextBuffer` without removing `IVimLocalSettings` seemed like leaving the job half-finished!

jaredpar: I agree. This is definitely an improvement.

PR #2246, opened 2018-07-11
Prevent adjusting for 'scrolloff' during buffer construction or view layout

### LINKED ISSUE #2240: When non-zero 'scrolloff' is used, starting caret position of first window is offscreen
I have been trying to as an experiment to use `:set scrolloff=2` as my normal VsVim editor setting and trying to observe any misbehavior. One odd symptom that I finally associated with this setting is that upon starting Visual Studio, the caret of the restored active document was often offscreen. This only seems to affect the very first document. Other lazy loaded documents appear correctly, that is "how they last appeared the last time that Visual Studio was previously closed."

In other words, the active document's buffer window was properly scrolled when Visual Studio was closed (more than `scrolloff` away from the window's top and bottom), but upon restarting Visual Studio, the active document scroll position is "warped" from its remembered position to quite the opposite of what `scrolloff` is intended to do, scrolled so that the caret position is just offscreen.

ricksladkey:
Fixes #2240

This also incidentally fixes a problem where double-clicking on a test in an unopened document in Test Explorer would cause the caret to be moved to exactly `scrolloff` lines from the top or bottom of the window, whereas without VsVim, the Test Explorer would center the caret in the window of its own accord.

Because delaying the scroll offset adjustment uses the synchronization context, any test that uses a non-zero scroll offset and manually moves the caret needs to run the testable synchronization context queue in order to effect the actual adjustment.

### DIFF
--- a/Src/VimCore/SelectionChangeTracker.fs
+++ b/Src/VimCore/SelectionChangeTracker.fs
@@ -104,8 +104,15 @@ type internal SelectionChangeTracker
         // Don't update the caret if a selection is occuring.  This could be a user double clicking, 
         // dragging, etc ...  Don't want to update the caret in that case, let the mode change handle
         // that.
-        if _textView.HasAggregateFocus && not _vimBuffer.IsProcessingInput then
-            _commonOperations.EnsureAtCaret ViewFlags.ScrollOffset
+        if not _textView.InLayout && not _vimBuffer.IsProcessingInput then
+
+            let doUpdate () =
+                if not _textView.IsClosed then
+                    _commonOperations.EnsureAtCaret ViewFlags.ScrollOffset
+
+            let context = System.Threading.SynchronizationContext.Current
+            if context <> null then
+                context.Post((fun _ -> doUpdate()), null)
 
     member x.OnBufferClosed() = 
         _bag.DisposeAll()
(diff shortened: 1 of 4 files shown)


Src/VimCore/SelectionChangeTracker.fs
```diff
@@ -101,11 +101,38 @@ type internal SelectionChangeTracker
...
+            // Delay the update to give whoever changed the caret position the
+            // opportunity to scroll the view according to their own needs.
+            // If another extension moves the caret far offscreen and then
+            // centers it if the caret is not onscreen, then reacting too
+            // early to the caret position will defeat their offscreen
+            // handling. An example is double-clicking on a test in an
+            // unopened document in "Test Explorer".
+            let context = System.Threading.SynchronizationContext.Current
+            if context <> null then
+                context.Post((fun _ -> doUpdate()), null)
```
jaredpar: What happens if this is done as a part of say macro execution? In that case the macro would completely execute and then all of the posted operations here would execute at the conclusion of the macro.

ricksladkey: I believe that even if we were to "try to keep the screen up-to-date" during macro execution, WPF would not update the window until control returns back to the event loop. This can be verified by, say, recording a macro for `G1GG1GG1G` that repeatedly goes to the beginning and end of the buffer and observing that the screen doesn't "flash through" all the intermediate updates.

In addition, this operation is supposed to be "idempotent" which is to say that repeating it 'n' times has the same effect as executing it once. Alternatively, it only needs to run "before the next time we are waiting for input from the user."

PR #2251, opened 2018-07-13
Fix race condition on startup

### LINKED ISSUE #2250: Race condition on startup
There is a race condition on startup in the following case:

- VS solution is reopened with windows in the following configuration:
    - Tab well separate from the main VS tab well
    - At least two documents opened where one is visible and one is hidden
    - Hidden one should be on the right
- The tab well is closed without ever making the hidden document visible 

It's likely the bug is more general than this but it repros very consistently in this setup.

CC @333fred

jaredpar:
The behavior here is that we get an `ITextView.LayoutChanged` event but
by the time we can react the `ITextView` is already closed. The fix is to
move our contract check that the `ITextView` is in a proper state to
after we've checked to see if it has been closed.

closes #2250


jaredpar:
CC @333fred

PR #2264, opened 2018-07-18
Allow <C-u>/<C-d> to reach beginning/end of buffer

### LINKED ISSUE #2248: When 'scrolloff' is non-zero <C-u>/<C-d> cannot reach the beginning/end of the buffer
In vim, regardless of the `scrolloff` setting, pressing `<C-u>` repeatedly will eventually reach the beginning of the buffer. Likewise, pressing `<C-d>` repeatedly will eventually reach the end of the buffer. In VsVim, the caret stays `scrolloff` lines away from the beginning or end of the buffer.

ricksladkey:
Fixes #2248

### DIFF
--- a/Src/VimCore/CommonOperations.fs
+++ b/Src/VimCore/CommonOperations.fs
@@ -345,9 +345,15 @@ type internal CommonOperations
                     None
                 elif caretLineNumber < (topLineNumber + _globalSettings.ScrollOffset) || caretLineNumber > (bottomLineNumber - _globalSettings.ScrollOffset) then
                     let lineNumber = caretLineNumber
-                    let lineNumber = max lineNumber (topLineNumber + _globalSettings.ScrollOffset)
-                    let lineNumber = min lineNumber (bottomLineNumber - _globalSettings.ScrollOffset)
-                    Some lineNumber
+                    let topOffset = min caretLineNumber _globalSettings.ScrollOffset
+                    let lastVisualLineNumber = _textView.VisualSnapshot.LineCount - 1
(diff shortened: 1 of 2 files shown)


Src/VimCore/CommonOperations.fs
```diff
@@ -345,9 +345,15 @@ type internal CommonOperations
...
-                    Some lineNumber
+                    let topOffset = min caretLineNumber _globalSettings.ScrollOffset
+                    let lastVisualLineNumber = _textView.VisualSnapshot.LineCount - 1
+                    let bottomOffset = min (lastVisualLineNumber - caretLineNumber) _globalSettings.ScrollOffset
+                    let lineNumber = max lineNumber (topLineNumber + topOffset)
+                    let lineNumber = min lineNumber (bottomLineNumber - bottomOffset)
+                    if lineNumber <> caretLineNumber then
+                        Some lineNumber
+                    else
+                        None
```
jaredpar: Why did you choose to implement this change here vs. directly in `CommandUtil.ScrollLines`? The latter is where I would have thought to do this. Doing it here seems like it would affect `CommandUtil.ScrollPages` as well.

ricksladkey: `ScrollLines` was already putting the caret on the right line and then it calls this function which is named `AdjustCaretForScrollOffset`. So anybody should be able to call `AdjustCaretForScrollOffset` and expect that it will only adjust the caret when the caret is "too close to the top/bottom of the window, except when at the top/bottom of the buffer."

In fact,`ScrollPages` didn't have the bug because of the way that scrolling up leaves the caret at the bottom of the screen and vice versa. However, if it didn't it would also need to correctly "not move the caret away from the edge" when the edge is in fact the top/bottom of the buffer.

PR #2265, opened 2018-07-18
Let window determine working directory for external commands

### LINKED ISSUE #2252: Running an external command or filter doesn't respect the window-specific current directory
In VsVim, after running `:lcd <new-directory><CR>`, running `:!cd<CR>` will display the global current directory, not the window-specific current directory.

ricksladkey:
Fixes #2252

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -4311,7 +4311,7 @@ type IVimHost =
 
     /// Run the specified command with the given arguments and return the textual
     /// output
-    abstract RunCommand: file: string -> arguments: string -> input: string -> vimHost: IVimData -> RunCommandResults
+    abstract RunCommand: workingDirectory: string -> file: string -> arguments: string -> input: string -> RunCommandResults
 
     /// Run the Visual studio command in the context of the given ITextView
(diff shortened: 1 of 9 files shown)


Src/VimCore/CoreInterfaces.fs
```diff
@@ -4381,6 +4381,9 @@ and IVimBufferData =
     /// The current file path for this buffer, relative to CurrentDirectory
     abstract CurrentRelativeFilePath : string option
     
+    /// The current directory for this particular window
```
jaredpar: Shouldn't this comment, and the one on CurrentDirectory, reference "buffer" not "window"?

ricksladkey: I am constantly torn whether to use Visual Studio terminology or vim terminology. In this case, the other property used "window" so I followed that pattern.

The vim documentation for `:lcd` is:

> Like |:cd|, but only set the current directory for the current window.  The current directory for other windows is not changed.

But, yes, VsVim does not use the "window" terminology and so it should be "buffer" for consistency. Visual Studio uses buffer vs. view in this situation but VsVim calls the view a vim buffer and the buffer a text buffer. But if we say "buffer" we don't mean other vim buffers with the same underlying text buffer.

PR #2269, opened 2018-07-19
Support for 'virtualedit' for visual character mode and for marks

### LINKED ISSUE #1507: Can't select last character without line break
Let's say you have the following C# snippet:

```
public interface IWithStatus : IWithStatusGetter
{
    /// <remarks>
    /// Without "new" there is a warning. The alternative solution "IWithStatus :
    /// IWithStatusGetter, IWithStatusSetter" would cause compiler warnings because of ambigious
    /// type references. Disadvantage of this solution: If one *explicitly* implements
    /// IWithStatus.Status, one must also implicitly or explicitly implement
    /// IWithStatusGetter.Status.
    /// </remarks>
    new eStatus Status { get; set; }
}
```

When I visually select (with lower case v) the sentence "Disadvantage of this solution: If one .... IWithStatusGetter.Status.", VsVim (1.7.1.1) also selects the line break (CR LF). This is different from the behaviour of Vim (7.4).

In the example, deleting the visual selection in VsVim with d or x would also delete the line break after "IWithStatusGetter.Status" which is irritating if you only want to delete the sentence itself.

### LINKED ISSUE #2186: Using 'gv' to restore a characterwise exclusive visual selection selects one too many characters
In VsVim, if you use `:set selection=exclusive` and make a visual characterwise selection like `vl`, type `<Esc>` and then type `gv`, the restored visual selection will select an additional character at the end of the selection.

### LINKED ISSUE #2260: How do "xnoremap p pgvy" work correctly?
I want to paste the same words over and over again,
I entered "xnoremap p pgvy" in .vsvimrc.
Next, when pasting to a word shorter than the number of copied words,
When pasting next time, a part of the copied word disappears.
How will I be able to paste correctly?

ricksladkey:
#### Changes

- Extend `CharacterSpan` to support character selections beginning or ending in virtual space
- Extend setting marks, jumping to marks and the jump list to support marks in virtual space
- Support yanking a characterwise selection that ends in virtual space
- Fix problems with exclusive characterwise selections wrongly including the linebreak
- Fix last visual selection after putting over selection

#### Fixes

- Fixes #1507
- Fixes #2186
- Fixes #2260

#### Discussion

This is most of the rest of the 'virtualedit' setting. The main feature missing is restoring a characterwise visual selection that starts in virtual space and the last selection end point when the selection ends in virtual space. Both of these features require considerable work for special marks, which can be part of a subsequent PR.

To avoid a large number of changes, particularly to the unit tests, we support both virtual (e.g. `VirtualSnapshotPoint`) and non-virtual (e.g. `SnapshotPoint`) in the API to `CharacterSpan`. It would be possible to remove the non-virtual API and use the virtual one exclusively. However, changing all callers, especially when those callers know for a fact that the specified points are not in virtual space, would require many trivial changes and complicate the callers.

My current opinion is that the dual virtual/non-virtual API is, at least for now, beneficial because it ensures that most code continues to use the same code paths when virtual space is not in use while allowing the support for virtual space to evolve and mature.

I do not doubt that daily use by a user who uses `virtualedit` heavily will uncover further areas that need to be virtualized, but I think this change does most of the groundwork and that those kinds of issues can be best uncovered with field testing.

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -1515,18 +1515,22 @@ type CharacterSpan =
 
     val private _lastLineLength: int
 
+    val private _useVirtualSpace: bool
+
     new (start: SnapshotPoint, lineCount: int, lastLineLength: int) =
         let virtualStart = VirtualSnapshotPointUtil.OfPoint start
-        CharacterSpan(virtualStart, lineCount, lastLineLength)
+        CharacterSpan(virtualStart, lineCount, lastLineLength, false)
 
-    new (start: VirtualSnapshotPoint, lineCount: int, lastLineLength: int) =
+    new (start: VirtualSnapshotPoint, lineCount: int, lastLineLength: int, useVirtualSpace: bool) =
 
         // Don't let the last line of the CharacterSpan end partially into a line 
         // break.  Encompass the entire line break instead 
         let number = start.Position.GetContainingLine().LineNumber + (lineCount - 1)
         let line = SnapshotUtil.GetLineOrLast start.Position.Snapshot number
         let lastLineLength = 
-            if line.Length = 0 then
+            if useVirtualSpace then
+                lastLineLength
+            elif line.Length = 0 then
                 line.LengthIncludingLineBreak
             elif lastLineLength > line.Length then
                 line.LengthIncludingLineBreak
@@ -1535,7 +1539,8 @@ type CharacterSpan =
         { 
             _start = start
             _lineCount = lineCount
-            _lastLineLength = lastLineLength }
+            _lastLineLength = lastLineLength
+            _useVirtualSpace = useVirtualSpace }
 
     new (span: SnapshotSpan) = 
         let lineCount = SnapshotSpanUtil.GetLineCount span
@@ -1548,7 +1553,7 @@ type CharacterSpan =
                 max 0 diff
         CharacterSpan(span.Start, lineCount, lastLineLength)
 
-    new (span: VirtualSnapshotSpan) =
+    new (span: VirtualSnapshotSpan, useVirtualSpace: bool) =
         let lineCount = VirtualSnapshotSpanUtil.GetLineCount span
         let lastLine = VirtualSnapshotSpanUtil.GetLastLine span
         let endColumnNumber = VirtualSnapshotPointUtil.GetColumnNumber span.End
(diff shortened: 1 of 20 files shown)


Src/VimCore/CoreInterfaces.fs
```diff
@@ -1509,61 +1509,82 @@ type KeyMappingResult =
...
 type CharacterSpan = 
 
-    val private _start: SnapshotPoint
+    val private _start: VirtualSnapshotPoint
 
     val private _lineCount: int
 
     val private _lastLineLength: int
 
+    val private _useVirtualSpace: bool
```
jaredpar: Having trouble understanding of this means

1. Consumer should consider the virtual space when consuming this type 
2. This type was created by considering virtual space of the inputs

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1509,61 +1509,82 @@ type KeyMappingResult =
 [<DebuggerDisplay("{ToString()}")>]
 type CharacterSpan = 
 
-    val private _start: SnapshotPoint
+    val private _start: VirtualSnapshotPoint
```
jaredpar: Did you consider breaking this into multiple types: CharacterSpan and VirtualCharacterSpan? That would give it symmetry with similar types. Possible that’s overkill here.

Src/VimCore/CoreInterfaces.fs
ricksladkey: I will add a explanatory comment.

It means "This character span was created when virtual space was in effect and should/will use the virtual interpretation of start point and length."

If you have a character span that selects four characters starting at the `c` in a line consisting only of `cat`, then the meaning of the character span depends on whether you are using virtual space. If you are using virtual space, the span ends in a space. Otherwise the span ends in a newline.

In other words, when using virtual space, an end point in virtual space always adds spaces whereas when not using virtual space, an end point in virtual space (which can only be just past the end of the line) includes the linebreak of that line.

Last way to explain it, although the selection looks identical in vim, the `virtualedit` setting can affect what gets yanked.

Src/VimCore/CoreInterfaces.fs
ricksladkey: There are many consumers of `CharacterSpan` that don't care if it's virtual or non-virtual; they just want to know the non-virtual interpretation of the span. This is analogous to caret position which is always encoded as a virtual point even when virtual space is not in use.

Likewise every non-virtual character span can be encoded as a virtual one with no loss of fidelity. The only disadvantage of using a single type is the same disadvantage as encoding the caret position as a virtual point: it is technically possible to encapsulate a span that is "more virtual" for the current `virtualedit` setting.

jaredpar:
>     member x.LastLineLength = x._lastLineLength

Consider having two properties here:

- LastLineLength: the length of the last line (non-virtual) 
- VirtualLastLineLength: the property as is 

Most of the other properties have virtual and non-virtual pairs here. This seems like another place where that would be useful. 

---
Refers to: Src/VimCore/CoreInterfaces.fs:1591 in 84cd45c. [](commit_id = 84cd45c05e3954660d13816033e75a6829180867, deletion_comment = False)

jaredpar:
Had one more comment. If you agree with the suggestion I think we can just pick it up in a future PR. This looks great!

ricksladkey:
@jaredpar  I will take a look at `LastLineVirtualLength`.

In other news, if you could merge hotfix #2284 it will fix the build. Thanks!

PR #2274, opened 2018-07-24
Format text lines

### LINKED ISSUE #333: gq doesn't format text
If you select a set of rows and press 'gq' the text isn't formatted into properly sized lines. In vim the width for the command is based on the 'textwidth' variable.

ricksladkey:
#### Changes

- Rename functions associated with `=` formatting to be named `FormatCodeLines`
- Add a new common operation to format text lines called `FormatTextLines`
- Add `gq` and `gw` operations to format text lines in visual mode
- Add `gqgq`, `gqq`, `gwgw` and `gww` operations to format text lines in normal mode
- Add `textwidth` setting defaulting to 0
- Add `comments` setting defaulting to line comments in C, C++, C#, F#, Java, Javascript, Python, Lisp, Shell

#### Fixes

- Fixes #333

#### Discussion

This is a "bare-bones" implementation of formatting text lines that omits many of the more esoteric vim features but adds the crucially missing capability to intuitively reformat comments in the most used programming languages. Subsequent more ambitious PRs can address the missing functionality.

#### Limitations

- The `formatoptions` variable is not implemented
- None of the comment specification flags in `comments` are implemented
- Tabs in the leader are properly accounted for but tabs in the text are not
- All the lines being formatted in one operation must have the same leader
- `gw` is currently the same as `gq`, i.e. it doesn't preserve the caret position

### DIFF
--- a/Src/VimCore/CommonOperations.fs
+++ b/Src/VimCore/CommonOperations.fs
@@ -248,6 +248,7 @@ type internal CommonOperations
             else
                 _localSettings.TextWidth
         let comments = _localSettings.Comments
+        let tabStop = _localSettings.TabStop
 
         // Extract the lines to be formatted and the first line.
         let lines = range.Lines |> Seq.map SnapshotLineUtil.GetText
@@ -331,13 +332,18 @@ type internal CommonOperations
             else
                 concatWords words :: lines
 
+        // Calculate the length of the leader with tabs expanded.
+        let leaderLength =
+            StringUtil.ExpandTabsForColumn leader 0 tabStop
+            |> StringUtil.GetLength
+
         // Aggregrate individual words into lines of limited length.
         let takeWord ((column: int), (words: string list), (lines: string list)) (word: string) =
 
             // Calculate the working limit for line length.
             let limit =
                 if lines.IsEmpty || useLeaderForAllLines then
-                    textWidth - leader.Length
+                    textWidth - leaderLength
                 else
                     textWidth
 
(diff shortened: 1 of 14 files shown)


ricksladkey:
I'm already using this every day! So nice...

Src/VimCore/CommonOperations.fs
```diff
@@ -228,10 +228,156 @@ type internal CommonOperations
...
+        _editorOperations.MoveToStartOfLineAfterWhiteSpace(false)
+
+    /// Format the text lines in the specified line range
+    member x.FormatTextLines (range: SnapshotLineRange) preserveCaretPosition =
+
+        // Get formatting configuration values.
+        let autoIndent = _localSettings.AutoIndent
+        let textWidth =
+            if _localSettings.TextWidth = 0 then
+                79
```
jaredpar: Why 79?

ricksladkey: That's what the vim documentation says it uses if `textwidth` is zero!

> If the 'textwidth' option is 0, the formatted length is the screen width (with a maximum width of 79).

I guess the idea is that it wraps at 80 columns but makes a concession for terminals that "double-wrap" the line if it is exactly 80.

Also, from this documentation you can see that I did not limit the width to the screen width, which doesn't seem to be a reasonable setting to use with Visual Studio where the use has virtually no control over the width of the window.

ricksladkey:
Moved the magic number to `VimConstants` with a comment.

PR #2279, opened 2018-07-26
Next match

### LINKED ISSUE #1449: Feature request: gn/gN motions
I recently learned about a pair of motions that Vim [added in patch 7.3.610](http://ftp.vim.org/vim/patches/7.3/7.3.610) (gn and gN) which reference the last used search pattern. This allows you do do things like "dgn" to delete the next match, or cgn to change the next match. If used on their own, they visually select the next match.

Since many of these types of things can technically be done with ":s", they may not be strictly necessary, but it does seem like it could make some of my editing patterns more efficient.

ricksladkey:
#### Changes

- Add the capability for the search service to include the start point in the search
- Add the `gn` and `gN` family of key bindings
- Implement 'next match' motion for operators in normal mode
- Implement 'select next match' operation from normal mode
- Implement 'extend selection to next match' from visual mode

#### Issues

- Fixes #1449

### DIFF
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -2937,7 +2937,7 @@ type internal MotionUtil
             let searchData = SearchData(last.Pattern, last.Offset, searchKind, options)
 
             // All search operations update the jump list.
-            _jumpList.Add x.CaretPoint
+            _jumpList.Add x.CaretVirtualPoint
 
             let searchResult = _search.FindNextPattern x.CaretPoint searchData _wordNavigator count
 
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -2925,8 +2925,8 @@ type internal MotionUtil
             else
                 x.MarkLine mark
 
-    /// Operate on the next occurrence of last pattern searched for
-    member x.NextSearch path count =
+    /// Operate on the next match for last pattern searched for
+    member x.NextMatch path count =
         let last = _vimData.LastSearchData
         if StringUtil.IsNullOrEmpty last.Pattern then
             _statusUtil.OnError Resources.NormalMode_NoPreviousSearch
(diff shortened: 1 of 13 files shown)


Src/VimCore/MotionUtil.fs
```diff
@@ -2925,6 +2925,38 @@ type internal MotionUtil
...
+                match searchResult with
+                | SearchResult.Error _ -> None
+                | SearchResult.NotFound _ -> None
+                | SearchResult.Found (_, span, _, _) ->
+
+                    let motionKind = MotionKind.CharacterWiseExclusive
+                    let isForward = path = SearchPath.Forward
+                    MotionResult.Create(span, motionKind, isForward) |> Some
+
+            _vimData.ResumeDisplayPattern()
```
jaredpar: Why is this needed here? I didn't see anything obvious in this function that would kill the display pattern here.

Src/VimCore/SearchService.fs
```diff
@@ -199,6 +193,20 @@ type internal EditorSearchService
 
         result
 
+    /// Get the match, if any, at the specified position
+    member x.GetMatchAtPosition (findData: FindData) (position: int) =
+        let snapshot = findData.TextSnapshotToSearch
+        let checkOptions = findData.FindOptions &&& (~~~FindOptions.SearchReverse)
+        let checkFindData = FindData(findData.SearchString, snapshot, checkOptions, findData.TextStructureNavigator)
+        match x.FindNextRaw checkFindData position with
+        | Some matchSpan ->
```
jaredpar: Consider using `when` here. I think it makes it a bit more concise. I've actually becoming a bigger fan of `when` after reviewing your PRs.

jaredpar:
Really excited about this motion. It's something I've wanted for a while now.

Src/VimCore/MotionUtil.fs
ricksladkey: I was unable to leverage `x.SearchCore` because the semantics were so different, but I did use it as a reference as if I were using it.  In this case, `ResumeDisplayPattern` is the second to last line of `x.SearchCore`.

My assumption is that if the user has manually disabled `hlsearch` that any kind of search continuation should reenable it. In other words, `gn` has the same semantics relative to `hlsearch` as `n` does.

Src/VimCore/SearchService.fs
ricksladkey: Part of my goal for contributing to VsVim is to get more real-world experience programming F#. One of the side effects of writing a lot of code is gradually learning about language features and formulating opinions about them.

The nuance I now understand about `when` with `match is that it subverts "exhaustive matching". That is fine when you are trying to do a match for one thing with an additional condition. But it always has to be weighed against the security of exhaustive matching.

In this case, `Some`/`None` matching will never be expanded to include a third option and so it is a good candidate for `match`/`when`.

PR #2293, opened 2018-08-05
C # Script Demo

TKumadaki:
This PR confirms what VsVim can do with C # script.

### Preparation

Only Visual Studio 2017 works.
Place the VsVimScripts folder in the user profile folder.
Add the following to .vsvimrc.

```
nnoremap yss :csx SimpleSurround<Enter>
nnoremap S :csx Scroll<Enter>
vnoremap S :csx SelectionMove<Enter>

```
### Methods.csx

Open the C # source and input the following in command mode:
`csx methods`
The method list and line number are displayed.
It was created as an aid to the `G` command.

### SimpleSurround.csx

Open the text file and enter the following.
`Hello !!`
Enter the following in normal mode.
`yss`
In addition enter the following.
`t`
The following is displayed in the command margin.
`<`
In addition enter the following.
`p width = "100"`
Then press the Enter key.
Then it is entered as follows.
`<p width = "100"> Hello !! </ p>`

I mimited a small part of vim surround.

### Scroll.csx

Enter `S` in normal mode.
Scroll with `j, k, d, u` command.
Entering another key returns to normal mode.

### SelectionMove.csx

Enter `S` in visual mode.
The selection range moves with` j, k` command.
Entering another key returns to normal mode.

### Solution.csx

Enter the following in command mode.
`csx solution`
The solution explorer can be operated with `j, k, enter` keys.

### Problem

User need to upgrade the framework version.
I do not understand Bad consequences of scripts operating Vim


TKumadaki:
@jaredpar You said in [# 2232](https://github.com/jaredpar/VsVim/pull/2232):

> Is there a reason you want to do this though? A lacking in the current experience you think could be solved with C# scripting integration?

I'd like to do something like this using C # script.

I think that there are various other problems in this PR,
I understand that it is difficult to upgrade the framework version

ricksladkey:
This doesn't seem quite complete because versions prior to VS2017 need to be able to resolve references, but it does use the VsSpecific mechanism so it should be easy to finish.

Since vimscript is not on the roadmap, and since this is fairly non-invasive to the rest of VsVim, it seems like a valuable mechanism to add functionality to a released version of VsVim.

Here are some reasons I can think of to support scripting:

- There are certain things you just can't do without modifying VsVim itself
- PRs to VsVim without general applicability are unlikely to be merged
- It has precedent in the `:vsc` command, which is not part of vim, but is widely used

TKumadaki:
@ricksladkey Thanks for the comment.

> This doesn't seem quite complete because versions prior to VS2017 need to be able to resolve references, but it does use the VsSpecific mechanism so it should be easy to finish.


Should I do the following work?

- Upgrade the version of the project framework of Vs 2013, Vs 2015.
- Add references such as Microsoft.CodeAnalysis .. to the project of Vs 2013, Vs 2015.
- Change CSharpScriptRunner, CSharpScriptParam directive.
- Display some message to the user of vs2012.

it may be better to use CSharpCodeProvider.
but I have never used CSharpCodeProvider...

ricksladkey:
@TKumadaki I'm just expressing my opinion! What really matters is what @jaredpar thinks so we need to wait until he weighs in on the subject.

TKumadaki:
@ricksladkey Thanks for the reply.
I will wait for opinions from @jaredpar.

ricksladkey:
It also seem relevant that my gvim comes with both Python and Perl scripting enabled so C# scripting for VsVim would be very much in the spirit of vim.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -5096,6 +5107,12 @@ and IVimBuffer =
     /// Whether the buffer is readonly
     abstract IsReadOnly: bool with get
 
+    /// Raised Call C# script
+    abstract RaiseCallCSharpScript : CallCSharpScriptEventArgs -> unit
+
+    /// VimBuffer Key Event Intercept
+    abstract KeyIntercept: bool with get,set
```
jaredpar: Why do we believe that we need this? The existing API has a lot of hook for consuming key strokes. I don't understand why a new API is needed here? Especially just for running scripts.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -5157,6 +5174,14 @@ and IVimBuffer =
     [<CLIEvent>]
     abstract Closed: IDelegateEvent<System.EventHandler>
 
+    /// Raised when Call CSharpScript
+    [<CLIEvent>]
+    abstract CallCSharpScript: IDelegateEvent<System.EventHandler<CallCSharpScriptEventArgs>>
```
jaredpar: I'm a bit skeptical that we need this. Consider that there is no such event for running a vim script. Feels like we should have an event for both or neither.

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -469,6 +469,10 @@ type VimInterpreter
     member x.RunCall (callInfo: CallInfo) = 
         _statusUtil.OnError (Resources.Interpreter_CallNotSupported callInfo.Name)
 
+    member x.RunCallCSharpScript (callInfo: CallInfo) = 
+        let args = CallCSharpScriptEventArgs(callInfo)
+        _vimBuffer.RaiseCallCSharpScript args
```
jaredpar: I think the best way to approach this is through the host. Essentially have a method on `IVimHost` that is used to invoke the script file.

Src/VsSpecific/Vs2017/Vs2017.csproj
```diff
@@ -4,10 +4,16 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Vim.VisualStudio.Vs2017</RootNamespace>
     <AssemblyName>Vim.VisualStudio.Vs2017</AssemblyName>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
```
jaredpar: Don't think this should be necessary. The Microsoft.CodeAnalysis.CSharp.Scripting library targets netstandard1.3 which is compatible with net46.

Src/VsSpecific/Vs2017/Vs2017.csproj
```diff
@@ -4,10 +4,16 @@
...
     <DefineConstants>$(DefineConstants);VS_SPECIFIC_2017</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.8.2" />
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="EnvDTE80">
+      <HintPath>..\..\..\References\Vs2010\envdte80.dll</HintPath>
+    </Reference>
```
jaredpar: This needs to be changed to 

```
 <Reference Include="envdte80, Version=8.0.0.0, Culture=neutral,
PublicKeyToken=b03f5f7f11d50a3a" />
```

Src/VsSpecific/VsSpecific/CSharpScriptRunner.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.CodeAnalysis.Scripting;
+using System.IO;
+using Microsoft.CodeAnalysis.CSharp.Scripting;
+using System.Reflection;
+using EnvDTE;
+
+namespace Vim.VisualStudio.Specific
+{
+    [Export(typeof(IVimBufferCreationListener))]
+    internal sealed class CSharpScriptRunner : IVimBufferCreationListener, IDisposable
```
jaredpar: I think the structure for running scripts should be the following:

1. Define a method on the host for running scripts `IVimHost.RunCSharpScript` 
1. Define a MEF interface for running scripts: `ICSharpScriptExecutor`
1. Export that interface here instead of `IVimBufferCreationListener` 
1. Optionally import `ICSharpScriptExecutor` inside `IVimHost`.

That pattern is much closer to how other features like this have been done. I feel like it's the best approach here.

TKumadaki:
@jaredpar Thank you for review.
My PR seems to have been wrong.
I am thinking that I would like to work on it again by referring to your review.
But I do not know if I can complete the work.
So I am going to close PR.

May I ask a question?

I added a command that is not in Vim, and I ran a C # script
What do you think about it?

Is the C # script a feature worth updating the framework version?

jaredpar:
> I added a command that is not in Vim, and I ran a C # script. What do you think about it?

I'm not inherently against it. I can see how it would be useful. If the integration was a bit more low impact I would likely be okay with it. 

> Is the C # script a feature worth updating the framework version?

If it's absolutely necessary. At a glance though it shouldn't be. Why did you think it required on here?

TKumadaki:
>I'm not inherently against it. I can see how it would be useful. 
>If the integration was a bit more low impact I would likely be okay with it.

I was pleased to hear that.
I follow your advice and I am creating a new branch.

>If it's absolutely necessary. At a glance though it shouldn't be. Why did you think it required on here?

I misunderstood at this time.

in Visual Studio 2017 VsVim do not need to update the version of the framework.
your advide is correct

PR #2296, opened 2018-08-06
More surrogate pair progress

jaredpar:
This adds `VirtualSnapshotColumn` which I think is the final API piece necessary to port the rest of VsVim to the new types.


ricksladkey:
I know I started this treasure hunt by implementing basic caret motions for emojis but I am very glad that you are the one taking seriously the ramifications of the frightening truth that "a text buffer uses an encoding." In my mind, the whole reason Windows and .NET used 16-bit characters was to get the benefit of *not* using an encoding.

The reality is that the benefit is just using a *single* encoding, and that encoding should have been, in hindsight, UTF-8. Nevertheless, UTF-16 is what we are stuck with. An alternative to snapshot column would be a snapshot line variant that uses 32-bit character for truly unencoded text, but snapshot column serves other purposes like handling double-width code points, so this solution seems like the best one.

jaredpar:
> The reality is that the benefit is just using a single encoding, and that encoding should have been, in hindsight, UTF-8.

Yep. I don't think I truly understood the implications of UTF-16 until I started going down this rabbit hole. Now that I'm at the bottom I realize that UTF-16 is just mostly hiding problems instead of solving them. 😦

jaredpar:
At least ... I hope I'm at the bottom of the rabbit hole 😄

PR #2298, opened 2018-08-06
Fix sources of excess allocations

jaredpar:
Profiling VsVim recently I discovered a couple of sources of significant and unneeded allocations:

1. Normal mode was using `ICommandRunners.Commands` to see if commands had been populated yet. Under the hood that was doing a transform on a large map and represented nearly half the allocations on unit testing and a significant amount of CPU time. 
1. `KeyInputSet.CompareTo` was allocating `KeyInput list` instances at a very high volume.


jaredpar:
Looks like it shaved off ~10 seconds of test run time in AppVeyor.

PR #2316, opened 2018-08-12
Small delete register fix

### LINKED ISSUE #2315: Small delete register is not updated as expected
In gVim 8.0, the small delete register "- is always updated when less than one line is deleted, unless a register (including the unnamed register "") is specified in the command.

In VsVim:
1. Specifying the unnamed register causes the small delete register to be updated.
2. With the option clipboard="unnamed", the small delete register is not updated when no register is specified in the command.

**Examples**
The following should hold regardless of the clipboard setting.

Command  |  Updates Small Delete
------- | -------------------------
`dw` | Y
`""dw` | N 
`"xdw` | N

dhmgit:
Fixes #2315


jaredpar:
Thanks!

PR #2321, opened 2018-08-13
Implement line command flags and corresponding commands

### LINKED ISSUE #2320: Support the ex command-line flags 'l', '#' and 'p'
VsVim parses but ignores the optional flags to, e.g., the `print` command.

ricksladkey:
Fixes #2320


ricksladkey:
I started to implement this feature just to get `:print` to print out the line number like vim. As usual with vim, it turned out to be way more complicated than one might expect!

PR #2325, opened 2018-08-14
Solution explorer key binding conflicts

### LINKED ISSUE #2283: Cannot paste in solution explorer
I can no longer use <kbd>Ctrl-V</kbd> to paste files into solution explorer. I suspect our changes to remove the copy / paste key bindings by default may be causing this. Haven't dug into it yet.

dhmgit:
This is intended to fix #2283.  I suspect that I am missing some detail since this seems a bit too simple; and if it were that simple, then someone more knowledgeable than myself would have already fixed the issue.

On the other hand, it seems to solve the specific problem that I have encountered:  Although I was previously able to assign a VS key binding such as CTRL+V to the Edit.Paste command in the "Solution Explorer" scope, configuring CTRL+V to be handled in VsVim would eventually remove the VS key binding, generally the next time VS was restarted.  After this change, VS key bindings scoped to "Solution Explorer" are left alone by VsVim, regardless of the "handled in" settings.

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/KeyBindingService.cs
+++ b/Src/VsVimShared/Implementation/Misc/KeyBindingService.cs
@@ -206,7 +206,7 @@ namespace Vim.VisualStudio.Implementation.Misc
         internal bool ShouldSkip(CommandKeyBinding binding)
         {
             var scope = binding.KeyBinding.Scope;
-            if (!_includeAllScopes && _scopeData.GetScopeKind(scope) == ScopeKind.Unknown)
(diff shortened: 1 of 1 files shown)


jaredpar:
This is actually a lot simpler than I was anticipating. I was worried we'd need to do some awful hackery here to get this working. I didn't realize there was a specific version of the copy / paste for this scope.

ricksladkey:
This is like a real-word example of the [Two Dollar Chalk Mark](https://colterreed.com/the-parable-of-the-two-dollar-chalk-mark/) parable. Bravo!

PR #2334, opened 2018-08-15
Fix duplicate ESC key processing when ReSharper is installed

### LINKED ISSUE #1373: Highlighting in Insert Mode goes to Visual Mode, but goes back to Command Mode
Steps to Reproduce:
1) Go to Insert Mode
2) Highlight text
3) See the cursor change to block
4) Hit 'Esc'
5) See mode is command mode

Performing same steps in Vim takes you back to Insert Mode. I am using VS 2012 with Resharper and Productivity Power Tools

dhmgit:
Fixes #1373 

This works for me (in all scenarios that I can conceive of) with ReSharper and VS2017.  

**Approach**
Although we can't directly detect when an ESC **was** missed by the buffer, it's quite easy to detect when it **wasn't** missed by the buffer.  Since ReSharperKeyUtil.PreviewKeyUp always (AFAIK) sees the ESC key, we can reset the "seen" flag every time ESC is released.  Thus we can indirectly detect an ESC key that the buffer didn't see.

**Suggestion 1**
I'm assuming that the original rationale behind exiting insert mode when an intellisense popup is dismissed via ESC was to match Vim's suggestion popup behavior. This makes sense when the suggestion popup must be invoked through a special key combination such as `<C-n>` or `<C-p>`.  However, intellisense providers in VS can generally be configured to automatically popup suggestions based on characters being inserted into the buffer.  I personally have ReSharper configured to disable this behavior -- otherwise the intellisense is constantly popping up with unwanted suggestions while typing in insert mode.  These popups must be dismissed with ESC, which reverts to normal mode, forcing insert mode to be re-entered manually in order to continue typing.  I think it is worth considering making this "ESC passthrough" behavior optional to accommodate those who want to keep the automatic intellisense popup behavior enabled.

**Suggestion 2**
To improve consistency, we might consider applying the ESC passthrough behavior to all modes and not just insert mode. My reasoning here is that you can invoke the various ReSharper popups from just about any mode using key bindings that are not handled by VsVim.  Again, this is in contrast to Vim, in which (AFAIK) you can only get suggestions in insert mode.  For instance, from any VsVim mode, I can bring up a "type search" popup with `<C-.>` which I have configured to be handled by VS.  Hitting ESC to dismiss the popup returns me to normal mode when I am in insert mode, but does not return to normal mode if I'm in, say, visual mode.

### DIFF
--- a/Src/VsVimShared/Implementation/ReSharper/ReSharperKeyUtil.cs
+++ b/Src/VsVimShared/Implementation/ReSharper/ReSharperKeyUtil.cs
@@ -17,11 +17,19 @@ namespace Vim.VisualStudio.Implementation.ReSharper
 
         private readonly IVimBuffer _vimBuffer;
         private readonly IVimBufferCoordinator _vimBufferCoordinator;
+        private bool _bufferSawEscape;
 
         private ReSharperKeyUtil(IVimBufferCoordinator vimBufferCoordinator)
         {
             _vimBufferCoordinator = vimBufferCoordinator;
             _vimBuffer = vimBufferCoordinator.VimBuffer;
+            _vimBuffer.KeyInputStart += OnBufferKeyInputStart;
+        }
+
+        private void OnBufferKeyInputStart(object sender, KeyInputStartEventArgs e)
+        {
+            if (e.KeyInput.Key == VimKey.Escape)
+                _bufferSawEscape = true;
         }
 
         internal static ReSharperKeyUtil GetOrCreate(IVimBufferCoordinator vimBufferCoordinator)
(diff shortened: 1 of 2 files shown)


jaredpar:
> Since ReSharperKeyUtil.PreviewKeyUp always (AFAIK) sees the ESC key

Just want to affirm this statement since you listed it as AFAIK. 

> This works for me (in all scenarios that I can conceive of) with ReSharper and VS2017.

Glad there is someone on this repo who frequently uses ReSharper. Too often I feel like I'm just tossing fixes over the wall with no concrete understanding of how impactful they are. I just don't use ReSharper enough to have a feel for parts of the experience.

> I'm assuming that the original rationale behind exiting insert mode when an intellisense popup is dismissed via ESC was to match Vim's suggestion popup behavior.

Indeed. It also matched the expectations of most of the user base. The original VsVim release had <kbd>Esc</kbd> only dismissing Intellisense but leaving it in insert mode. Got lots of negative feedback about this. Hence now at the behavior we have today. 

> To improve consistency, we might consider applying the ESC passthrough behavior to all modes and not just insert mode

Perhaps what we need to do here is define the pass through behavior of <kbd>Escape</kbd>

- Default: no pass through. The behavior we have today.
- Insert: pass through only when in insert mode and there is a popup active. 
- All: pass through in all modes when a popup is active. 

I considered adding this in the past for ReSharper but the problem I keep getting stuck on is detecting when a popup is active. I haven't found a reliable way to do this.

ricksladkey:
@jaredpar Although it might your life easier if I not only reviewed straightforward changes like this but also merged them, I am going on the assumption that you would prefer to either review or merge every change that goes into master. So you can either merge this at your convenience (or whatever) or let me know that you want me to both review and merge PRs like this one.

PR #2335, opened 2018-08-16
Add vertical scrolling to command margin

### LINKED ISSUE #1808: :registers doesn't show named registers
When I type :registers I get the basic set of numbered registers, but not the named a-z registers, that I would expect to see in normal vim. Am I missing something or is this not supported yet?

image

dhmgit:
Fixes #1808

### DIFF
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginControl.xaml
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginControl.xaml
@@ -43,45 +43,42 @@
             CaretBrush="{Binding Path=TextForeground}"
             />
 
-        <ScrollViewer 
-            Grid.Row="1" 
-            Grid.Column="1"
-            HorizontalScrollBarVisibility="Auto"
-            VerticalScrollBarVisibility="Disabled">
-            <Border BorderThickness="1" BorderBrush="{Binding ElementName=_commandLineInput, Path=BorderBrush}">
-                <Grid>
-                    <Grid.ColumnDefinitions>
-                        <ColumnDefinition Width="*" />
-                        <ColumnDefinition Width="Auto" />
-                    </Grid.ColumnDefinitions>
-                    <TextBox
-                        x:Name="_commandLineInput"
-                        MaxHeight="200"
-                        BorderThickness="0"
-                        VerticalAlignment="Center"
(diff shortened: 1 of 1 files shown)


Src/VimWpf/Implementation/CommandMargin/CommandMarginControl.xaml
```diff
@@ -43,45 +43,42 @@
...
-                        VerticalAlignment="Center"
-                        FontFamily="{Binding Path=TextFontFamily}"
-                        FontSize="{Binding Path=TextFontSize}"
-                        Foreground="{Binding Path=TextForeground}"
-                        Background="{Binding Path=TextBackground}"
-                        CaretBrush="{Binding Path=TextForeground}" 
-                        />
-                </Grid>
-            </Border>
-        </ScrollViewer>
```
jaredpar: Do you understand why the `ScrollViewer` didn't work in the old XAML code? Asking because I am curious what was wrong here.

dhmgit: I think my choice to move the scrolling to the text box (which has a scrollviewer embedded) has obscured the underlying issue.  I'll try to explain both.

The reason the scroll viewer wasn't showing its vertical scrollbar (even after being set to "Auto") was that its height was unconstrained.  An unconstrained scroll viewer will happily resize itself to accommodate its contents.  Since scroll bars are only enabled when the scroll viewer's content area is smaller than its contents, you would never have seen the vertical scrollbar.

Moving `MaxHeight="200"` from `_commandLineInput` to the `ScrollViewer` tag would have fixed this particular problem. On the other hand, the scrollbar is now **outside** of the text box and the thin border around it.  If the text box were tall enough to cause the vertical scrollbar to show, then the bottom border would have disappeared "off screen", which I thought looked weird (see image below).  The simplest solution was to hand over scrolling to the scroll viewer embedded within the text box itself.

image

dhmgit: To contrast with the above image, here's the scroll bar **inside** the text box:
image

jaredpar: Thanks for the explanation!

PR #2337, opened 2018-08-16
Digraphs

### LINKED ISSUE #1759: Implement digraphs and Ctrl-K
http://vimdoc.sourceforge.net/htmldoc/digraph.html

ricksladkey:
Fixes #1759


Test/VimCoreTest/InsertModeIntegrationTest.cs
```diff
@@ -371,6 +371,108 @@ public void NormalizeLineEndings()
...
+
+            [WpfFact]
+            public void InlineSwapped()
+            {
+                Create("", "");
+                _globalSettings.Digraph = true;
+                _vimBuffer.ProcessNotation(":<BS>e");
+                Assert.Equal("ë", _textBuffer.GetLine(0).GetText());
+                Assert.Equal(1, _textView.GetCaretPoint().Position);
+            }
```
jaredpar: Another test to add. Start with the word `CAT` in register and in a blank line in insert mode: `<Ctrl-R>c<BS>H`. That should cause the digraph for TH to be entered. It should excercise a slightly different code path than normal typing.

jaredpar:
Can't wait to add this to my vimrc
:dig ET 128125

Src/VimCore/Vim.fs
```diff
@@ -664,6 +670,10 @@ type internal Vim
 
     member x.MaybeLoadFiles() =
 
+        if x.AutoLoadDigraphs && not _digraphsAutoLoaded then
+            DigraphUtil.AddToMap _keyMap DigraphUtil.DefaultDigraphs
```
jaredpar: I think we should create a separate map for digraphs vs. storing them into the standard key map. The aren't a mode and don't have some of the odd states that keyremapmodes have for key resolution

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -644,6 +644,43 @@ type VimInterpreter
...
+            System.Char.ConvertToUtf32(getChars keyInputSet, 0)
+
+        if List.isEmpty digraphList then
+            let columns = 4
+            _keyMap.GetKeyMappingsForMode KeyRemapMode.Digraph
+            |> Seq.map (fun keyMapping ->
+                (
+                    getChars keyMapping.Left,
+                    getDisplay keyMapping.Right,
+                    getCode keyMapping.Right
```
jaredpar: This feels like we are working against `KeyInputSet` here vs. it being of value. I think it would be a bit cleaner if we just held diagraphs in a separate map with a type that has the concrete behaviors we need: two characters and a code point.

Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -865,6 +876,27 @@ type internal InsertMode
...
+                        // and if so, insert it.
+                        let firstKeyInput =
+                            text.[text.Length - 1]
+                            |> KeyInputUtil.CharToKeyInput
+                        let secondKeyInput = keyInput
+                        match x.TryInsertDigraph firstKeyInput secondKeyInput with
+                        | Some processResult ->
+                            Some processResult
+                        | None ->
+                            x.TryInsertDigraph secondKeyInput firstKeyInput
```
jaredpar: Good catch. Didn't see any documentation about it trying the reverse diagraph here.

Src/VimCore/Interpreter_Interpreter.fs
ricksladkey: Yeah, kinda like vim storing a macro in a register.  :-)

When I was prototyping, it wasn't clear to me that digraphs *wouldn't* be used as real mappings in insert mode when `digraph` is set. And of course I was trying to leverage existing infrastructure to complete the job quickly. But I agree that a digraph-specific data structure that is a thin wrapper around a string -> string dictionary would be a better fit.

Src/VimCore/Modes_Insert_InsertMode.fs
ricksladkey: Yes, it's in there!

> If a digraph with {char1}{char2} does not exist, Vim searches for a digraph
> {char2}{char1}.  This helps when you don't remember which character comes
> first.

Src/VimCore/Interpreter_Parser.fs
```diff
@@ -2231,6 +2233,39 @@ type Parser
...
+                _tokenizer.MoveNextChar()
+                if _tokenizer.IsAtEndOfLine then
+                    result <- LineCommand.ParseError Resources.CommandMode_InvalidCommand |> Some
+                else
+                    let secondChar = _tokenizer.CurrentChar
+                    _tokenizer.MoveNextChar()
+                    x.SkipBlanks()
+                    match x.ParseNumber() with
+                    | Some number ->
+                        let digraph = (string(firstChar) + string(secondChar), number)
```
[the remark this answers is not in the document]
ricksladkey: The documentation only seems to support 16-bit unicode, and in my testing with gvim 8.0 I could not get a 32-bit code point to work, but I might not have had my buffer encoding set up correctly.

In any case, it was straightforward to allow it (knowing what we know now), and I did test adding our old friend the extraterrestrial alien emoji as a digraph and that worked just fine.  Probably should add a test for that!

Test/VimCoreTest/InsertModeIntegrationTest.cs
ricksladkey: Interesting! Although, vim does "support" that, it seems more a like a unfortunate consequence. Think about the analogous operation (which also works in vim): putting `0` into a register, pasting it and then typing `<C-d>` removes all indent.

Anyway, it does work also in "VsVim" because a text change records a paste as though the user typed it. I can add a test.

ricksladkey:
I'm out of town this weekend but I will fit making these changes into my spare time.

jaredpar:
No need to rush.

ricksladkey:
I added a special-purpose digraph map and added a test for "as if typed" register paste forming part of an inline digraph sequence.

PR #2343, opened 2018-08-21
Include empty lines in block insert

### LINKED ISSUE #2342: Column insert in visual block ignores empty lines
> Note: In this issue I use the term 'empty line' rather than 'blank line' to differentiate between a line with truly nothing (an empty line) and a line that contains only one or more whitespace characters (a blank line). This issue pertains *only* to empty lines

When inserting text across multiple lines in visual block mode, empty lines do not get the inserted text.

EXAMPLE

From the top of the code fragment below I follow this sequence..

1.     Ctrl-Q
2.     (reposition cursor to the end of the fragment)
3.     "I# " _(without quotes)_

Original code fragment..

```
for (x=1; x<=10; ++x) {

  do something
  do something else

  do another thing
  print results
}
```

becomes..

```
# for (x=1; x<=10; ++x) {

#   do something
#   do something else

#   do another thing
#   print results

# }
```

What I expect (and how vim 8.1 behaves) is this..

```
# for (x=1; x<=10; ++x) {
#
#   do something
#   do something else
#
#   do another thing
#   print results
#
# }
```

Reasons this issue matters are..

1. Behavior is different than gvim (v8.1 on windows)

2. Prevents subsequent navigations to the start or end of the block using '{' and '}' from working efficiently (each press of { or } only navigates as far as the next empty line)

3. The resultant code it is less visually appealing.

jaredpar:
The check for valid column position in block insert were excluding empty
lines.

closes #2342

### DIFF
--- a/Test/VimCoreTest/VisualModeIntegrationTest.cs
+++ b/Test/VimCoreTest/VisualModeIntegrationTest.cs
@@ -1217,6 +1217,18 @@ namespace Vim.UnitTest
                     _vimBuffer.ProcessNotation(@"<C-q>jjI#<Esc>");
                     Assert.Equal(new[] { "#dog", "#", "#tree" }, _textBuffer.GetLines());
                 }
+
+                /// <summary>
+                /// A block insertion should add the text when the insertion is at the end column
+                /// of the line. 
+                /// </summary>
+                [WpfFact]
+                public void EndOfLine()
+                {
+                    Create("dog", "x", "tree");
+                    _vimBuffer.ProcessNotation(@"l<C-q>jjI#<Esc>");
+                    Assert.Equal(new[] { "d#og", "x#", "t#ree" }, _textBuffer.GetLines());
+                }
             }
 
             public sealed class InsertTabTest : BlockInsertTest
--- a/Test/VimCoreTest/VisualModeIntegrationTest.cs
+++ b/Test/VimCoreTest/VisualModeIntegrationTest.cs
@@ -1204,8 +1204,20 @@ namespace Vim.UnitTest
                     Assert.Equal("og", _textView.GetLine(0).GetText());
                     Assert.Equal("cat", _textView.GetLine(1).GetText());
                 }
-            }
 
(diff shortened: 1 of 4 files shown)


Test/VimCoreTest/VisualModeIntegrationTest.cs
```diff
@@ -1204,8 +1204,20 @@ public void DontRepeatDelete()
...
+                /// Issue 2342
+                /// </summary>
+                [WpfFact]
+                public void EmptyLineBlockAtStartOfLine()
+                {
+                    Create("dog", "", "tree");
+                    _vimBuffer.ProcessNotation(@"<C-q>jjI#<Esc>");
+                    Assert.Equal(new[] { "#dog", "#", "#tree" }, _textBuffer.GetLines());
+                }
+            }
```
ricksladkey: A similar test from vim column 2 with the buffer:

```
dog
x
tree
```

should result in:

```
d#og
x#
t#ree
```

but with the current code change, the middle line doesn't get an insertion:

```
d#og
x
t#ree
```

ricksladkey: It would also be good to have to a test for when we *correctly* don't perform an insertion, that is when the block is truly *past* the end-of-line. An example is the buffer:

```
dog

tree
```

from vim column 2.

jaredpar: > but with the current code change, the middle line doesn't get an insertion:

Good catch. Let me add that. 

> It would also be good to have to a test for when we correctly don't perform an insertion, that is when the block is truly past the end-of-line

That test is already present: `BlockInsertTest.MiscTest.EmptyColumn`

jaredpar:
Fixed the other scenario around end of line + added the tests.

Also I started to add some testing for surrogate pairs and block insert but ran into some rather surprising behavior. Need to dig into it a bit more to understand how Vim handles surrogate pairs more deeply. Filed #2344 to track this bigger item.

PR #2347, opened 2018-08-23
fixes for one-commmand mode switching and display

### LINKED ISSUE #2346: One-command mode issues
A number of mode switch sequences involving "one-command" operations do not match gVim 8.0 behavior.  Some modes are simply misreported in the status area (e.g. (insert) VISUAL might be shown as VISUAL).  A number of these sequences actually enter an unexpected or invalid state.  

_Note: the following test cases all start in normal mode_

Sequence | Expected | Actual
--- | --- | ---
`gh<C-o>:<C-u>pwd<CR>` | NORMAL | SELECT
`gh<C-o>:<Esc>` | NORMAL | SELECT
`i<C-o>:<Esc>` | INSERT | (insert)
`i<C-o>R<Esc>` | NORMAL | (insert)
`i<C-o>gh<C-o>:<Esc>` | INSERT | SELECT
`i<C-o>gh<C-o>:<C-u>pwd<CR>` | INSERT | SELECT
`i<C-o>gh<C-o><Esc>` | INSERT | SELECT
`i<C-o>gh<C-o>w` | (insert) SELECT | SELECT
`i<C-o>ghx` | INSERT | ? (need 3 `<Esc>` to return to NORMAL)
`i<C-o>i<Esc>` | NORMAL | (insert)
`i<C-o>v:<Esc>` | INSERT | (insert)

dhmgit:
Fixes #2346 

This is the most "invasive" change I've made so far in this codebase, so I expect a fair amount of scrutiny.  The biggest challenge is to deal with the very large number of code paths that can lead to a mode switch without resorting to adding a lot of special case logic, or modifying the mode-specific code.  The more I experimented with gVim, the more I was surprised by features (or "features") that I didn't know existed.  Hopefully, I've hit all or nearly all of them.

### DIFF
--- a/Src/VimCore/VimBuffer.fs
+++ b/Src/VimCore/VimBuffer.fs
@@ -190,7 +190,6 @@ type internal VimBuffer
     let _keyMap = _vim.KeyMap
     let mutable _processingInputCount = 0
     let mutable _isClosed = false
-    let mutable _inSelectModeOneCommand = false
 
     /// This is the buffered input when a remap request needs more than one 
     /// element
@@ -488,10 +487,10 @@ type internal VimBuffer
                     // Certain types of commands can always cause the current mode to be exited for
                     // the previous one time command mode.  Handle them here
                     let maybeLeaveOneCommand() = 
-                        if _inSelectModeOneCommand then
+                        if _vimTextBuffer.InSelectModeOneTimeCommand then
                             // completing any command (even cursor movements) in a one-command mode initiated from a 
                             // select mode will revert to the analogous select mode
-                            _inSelectModeOneCommand <- false
+                            _vimTextBuffer.InSelectModeOneTimeCommand <- false
                             match VisualKind.OfModeKind x.Mode.ModeKind with
                             | Some visualModeKind -> 
                                 x.SwitchMode visualModeKind.SelectModeKind ModeArgument.None |> ignore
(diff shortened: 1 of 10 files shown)


Src/VimCore/VimBuffer.fs
```diff
@@ -190,6 +190,7 @@ type internal VimBuffer
     let _keyMap = _vim.KeyMap
     let mutable _processingInputCount = 0
     let mutable _isClosed = false
+    let mutable _inSelectModeOneCommand = false
```
jaredpar: Believe this state should be in `IVimTextBuffer` now. This would line up better with recent changes.

Src/VimCore/VimBuffer.fs
```diff
@@ -682,12 +688,43 @@ type internal VimBuffer
...
+                | Some oneTimeModeKind ->
+                    _inSelectModeOneCommand <- false
+                    _vimTextBuffer.InOneTimeCommand <- None
+                    oneTimeModeKind
+                | None ->
+                    modeKind
+            else
+                modeKind
+        
+        _modeMap.SwitchMode overrideModeKind modeArgument
```
jaredpar: This logic feels like it's in the wrong place. The implementation of `SwitchMode` should be pretty much what it is today: switch to the mode that is provided. Logic around taking a desired mode and changing it based on the current state should be at a higher layer.

dhmgit: At some point, I think I became overly concerned about missing a case without centralizing the switching override logic.  It turned out to be quite easy to move most of the logic back to VimBuffer.ProcessOneKeyInput where it originally resided, especially with the added test net.  Only one special case was needed in SelectionChangeTracker.  This seems like the best compromise because moving all of the mode switch logic to mode-specific layers would involve touching somewhere north of 200 locations that return ModeSwitch results.

How do you feel about the one-time command logic added to SwitchPreviousMode?  Removing this logic would mean duplicating it in both VimBuffer as well as VimInterpreter.RunNormal.  My testing indicates that running a "norm" command from one-command mode switches back to the saved mode (and not to command mode, which is technically the previous mode).

PR #2348, opened 2018-08-24
<C-c> in the command margin copies if there is a selection

### LINKED ISSUE #2338: <C-c> in the command line should copy when there is a selection
Although vim does not support selections in command mode, VsVim does, and it would be very helpful if when characters were selected in the command margin, `<C-c>` performed the copy operation instead of functioning like the `<Esc>` key. In VsVim 2.5.0, `<C-c>` *does* copy in the command margin, but VsVim master currently always treats `<C-c>` like `<Esc>`, even if there is a selection.

ricksladkey:
Fixes #2338

### DIFF
--- a/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
+++ b/Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
@@ -417,7 +417,7 @@ namespace Vim.UI.Wpf.Implementation.CommandMargin
                     {
                         if (_margin.CommandLineTextBox.SelectionLength != 0)
                         {
-                            // Let the text box handle <C-c> if there is a selection.
+                            // Copy if there is a selection.
                             // Reported in issue #2338.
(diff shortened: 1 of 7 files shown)


Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -74,6 +74,7 @@ internal void Clear()
         private readonly IEditorFormatMap _editorFormatMap;
         private readonly IClassificationFormatMap _classificationFormatMap;
         private readonly ICommonOperations _commonOperations;
+        private readonly IClipboardDevice _clipboardDevice;
```
ricksladkey: In order to make this change testable, we needed to supply the command margin controller with a clipboard service. When used as an extension, the clipboard device is the real clipboard; When running unit tests, the clipboard device is a testable mock of the real clipboard.

Issue #2344, opened 2018-08-21
Surrogate pair spaces


jaredpar:
The spacing logic for surrogate pairs appears to be incorrect. For symbols like `\U0001F47D` (👽) vim appears to treat them as accounting for two spaces, not one. Consider the following buffer (make sure to `:set enc=utf-8`):

```
dog
👽👽👽👽👽👽
cat
```

Now starting at the beginning and type `llj` this will put you on the second alien not the third. That indicates that vim believes the first alien occupies two spaces. If it believes it occupied one then it would have tracked to the first alien. The same is true if you use `lllj`. 

To fix this need to research a bit how Vim assigns space counts to these characters. It could be as simple as surrogate pairs occupy two spaces. Doubt it's that easy. It doesn't seem to be due to some unicode category that I can tell. 

There is another twist though to consider: block insert. Take the same repro sample above, starting at the beginning and type: `l<C-Q>jjI#<Esc>`. That will leave you with the following

```
d#ogs
#👽👽👽👽👽👽
t#rees
```

That behavior suggests that when a spaces count in block insert "splits" a surrogate pair the insert should go before the pair, not after. This is different that wide characters in the BMP plane. These also count for two spaces but when a block insert "splits" a wide character it's converted to spaces then the spaces are split. This is another scenario that I need to dig into a bit more to understand the full implications.

This came from digging into PR feedback on #2343

ricksladkey:
These are the best references I could find:

https://news.ycombinator.com/item?id=11963781

For example, the extraterrestrial alien emoji is listed in this official Unicode list as "East Asian Wide" (W):

https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt

jaredpar:
Thanks. I think that should be enough to get this going. Really two issues here: 

1. Update the width logic to account for Unicode 9
1. Update the block insert logic to handle splitting of surrogate pairs to not convert into spaces. 

For (2) I still don't have a concrete understanding of how Vim makes this decision. I may need to go dig through their source a bit better to understand what the distinction is.

jaredpar:
The logic seems to be simply that a block insert never starts in the middle of a multibyte character. Instead it always moves to the front for that given row.

PR #2349, opened 2018-08-26
Move to Unicode 11.0

jaredpar:
This moves us to the Unicode 11.0 for purposes of determining if characters are wide or normal. 

The new file, UnicodeUtil.fs, is quite large because I embedded the Unicode 11.0 tables inside it. That does hurt the readability of the file a bit (I've been editing it with the table minimized). I considered a couple of other approaches:

1. Move the table to a separate file entirely. That forced me to make a few types visible that I didn't want to, mostly `UnicodeRangeEntry`.
1. Embed the unicode standard as a text resource and parse it at startup. That would make startup a bit slower as I'd have to parse then create the record where as the current approach just inlines the record creation. I'm still debating a bit if that is worth it. 

Still debating a bit between the current approach and (2). 

Part one of fixing #2344. Will handle the block insert issue in a separate PR.

### DIFF
--- a/Src/VimCore/Unicode.fs
+++ b/Src/VimCore/Unicode.fs
@@ -3,7 +3,7 @@
 namespace Vim
 
 [<RequireQualifiedAccess>]
-type EastAsianWidth = 
+type internal EastAsianWidth = 
     | FullWidth
     | HalfWidth
     | Wide
@@ -22,7 +22,6 @@ type EastAsianWidth =
         | "A" -> EastAsianWidth.Ambiguous
         | _ -> invalidArg "text" "Not a valid value"
 
-[<Struct>]
 type UnicodeRangeEntry = {
     Start: int
     Last: int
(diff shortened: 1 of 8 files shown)


Src/VimCore/Unicode.fs
```diff
@@ -0,0 +1,2583 @@
...
+        | None -> 0
+        | Some node -> node.Height
+
+    static member private GetCount (node: IntervalTreeNode option) = 
+        match node with 
+        | None -> 0
+        | Some node -> node.Count
+
+/// A balanced internal tree for efficient lookup of unicode information
+type internal IntervalTree
```
ricksladkey: As a data structure enthusiast, I gotta appreciate using the right tool for the right job.

jaredpar:
> I'm worried that we are dependent on Visual Studio using the same logic and it isn't using it. 

I have similar worries. I considered threading through an interface everywhere we thread through `tabStop` today and let it decide character width:

``` fsharp
type IVimCharacterWidth = 
    abstract GetSpaces: codePoint: int -> int 
```

Eventually though decide to continue on the current course for now. Based on what I gathered both vim / gVim use this approach of picking a "spaces" count based on unicode data vs. actual displayed size.

PR #2350, opened 2018-08-26
Use explicit names in discriminated unions

jaredpar:
Recently was having a bit of trouble following a few of our tests due to the heavy use of `Item1`, `Item2`, etc ... in the code. This is the default naming scheme for discriminated union values and the only option when I originally started VsVim. Now there is the option of giving Discriminated Union values explicit names and they map into the generated members. 

I did an experiment where I forced the naming of all values and saw how it impacted the test code. Overall I like the effects. But it's a bit change wanted to get some feedback about the approach.

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -3174,17 +3174,17 @@ type NormalCommand =
 type VisualCommand = 
 
     /// Add count to the word in each line of the selection, optionally progressively
-    | AddToSelection of bool
+    | AddToSelection of IsProgressive: bool
 
     /// Change the case of the selected text in the specified manner
-    | ChangeCase of ChangeCharacterKind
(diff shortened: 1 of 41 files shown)


dhmgit:
Since you requested feedback, I'll add my two cents.

Besides some low-risk churn, I don't see a downside.  It's hard to argue with more descriptive naming.  It's also nice to be able to comprehend code by reading it as text, as opposed to needing an IDE to grasp the type and purpose of members and variables.  Furthermore, it's well known that naming is hard, but if you really can't think of an appropriate name then it's generally wise to rethink your implementation.

I've had a similar rule (forbidding non-private unnamed tuples) in place for quite some time in a large C# project that I work on professionally.  Other than the occasional raised eyebrow, I haven't had any real complaints from developers. Now that named tuples are a feature in the latest language versions, the last excuse for exposing unnamed tuples has been eliminated.

jaredpar:
Thanks for the feedback! Glad you all like the change as well.

PR #2355, opened 2018-08-28
Fix surrogate pair block insert

### LINKED ISSUE #2344: Surrogate pair spaces

jaredpar:
This addresses block insert around surrogate pairs. Particularly when
the block insert column is in the middle of a wide surrogate pair. This
should be done by pretending the column was the start of the surrogate
pair. This is different than wide characters which are not surrogate
pairs; those are converted to spaces first and then block insert occurs
in those spaces.

closes #2344

### DIFF
--- a/Src/VimCore/EditorUtil.fs
+++ b/Src/VimCore/EditorUtil.fs
@@ -317,6 +317,8 @@ type SnapshotCodePoint =
 
     member x.IsInsideLineBreakOrEnd = x.IsEndPoint || x.IsInsideLineBreak
 
+    member x.IsSurrogatePair = x.CodePointInfo = CodePointInfo.SurrogatePairHighCharacter
+
     /// Is the unicode character at this point represented by the specified value? This will only match
     /// for characters in the BMP plane.
     member x.IsCharacter(c: char) = 
@@ -896,6 +898,8 @@ and [<Struct>] [<StructuralEquality>] [<NoComparison>] [<DebuggerDisplay("{ToStr
 
     /// The SnapshotColumn in which this overlap occurs
     member x.Column: SnapshotColumn = x._column
+
+    member x.CodePoint = x.Column.CodePoint
 
     member x.Line = x.Column.Line
 
--- a/Src/VimCore/InsertUtil.fs
+++ b/Src/VimCore/InsertUtil.fs
@@ -203,7 +203,7 @@ type internal InsertUtil
                 let lineNumber = startLineNumber + i
                 let line = SnapshotUtil.GetLine currentSnapshot lineNumber
                 let column = SnapshotOverlapColumn.GetColumnForSpacesOrEnd(line, spaces, _localSettings.TabStop)
-                if column.SpacesBefore > 0 && not column.CodePoint.IsSurrogatePair then
+                if column.SpacesBefore > 0 && column.Column.IsCharacter '\t' then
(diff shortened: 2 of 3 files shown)


ricksladkey:
I cannot find a test for splitting and replacing with spaces a wide east asian codepoint in the BMP. And I am confused by the logic that the splitting and replacing with spaces should apply to any codepoints. For comparison, I cannot find any such characters that make my gvim 8.0 behave that way.

To be specific, I tested the following buffer with both gvim 8.0 and VsVim with this PR and got different results:

```
dog
𝕥xxx <- (U+1D565) MATHEMATICAL DOUBLE-STRUCK SMALL T (> 0xffff && normal width)
👽xxx <- extraterrestrial alien (> 0xffff && double width)
𠂊xxx <- (U+2008A) Han Character (> 0xffff && double width, non-emoji)
是xxx <- CJK Ideograph "是" (U+662F) (< 0xffff && double width)
bear
```

With VsVim and this PR, this results in:

```
d#og
𝕥#xxx
#👽xxx
#𠂊xxx
 # xxx
b#ear
```

But with my gvim 8.0 this results in:

```
d#og
𝕥#xxx
#👽xxx
#𠂊xxx
#是xxx
b#ear
```

So as far as I can tell, east asian wide codepoints *always* block insert before and *never* replace with spaces, which makes more sense to me with my admittedly limited understanding.

jaredpar:
I believe you're right. I ran some other tests and they line up with the experiment you did. At this point I'm confused why I believed they did this at all. I tried going back through the history of my changes here to see if I could find a reason and nothing jumped out. I'm going to blame faulty memory here. 

The one case where it does convert a character to spaces before inserting is when it crosses a non-expanded tab. Example (make sure to have `:set note` and `set ts=4`)

```
biggest
a<Tab>dog
smallest
```

Now if you execute `ll<C-q>jjI#<Esc>` you end up with 

```
bi#ggest
a #  dog
sm#allest
```

The tab here is deleted, replaced with equivalent number of spaces (three) and then the insert proceeds on that text. 

Will clean up the faulty behavior here and make sure that behavior is tested properly.

jaredpar:
Updated to only special case tabs here. Ensured there are tests for that in `PartialTabEditTest`

PR #2359, opened 2018-08-30
Fix macro record

### LINKED ISSUE #2356: Closing a buffer with a keyboard command disables macro recording
Closing a buffer using a keyboard command such as `ZZ` or `:q<cr>` causes subsequent attempts to record a macro to fail.

The root cause appears to be the MacroRecorder's `_nestingDepth` variable, which fails to decrement to zero when the buffer is closed in this manner.  This is because the buffer's Closed event fires before the KeyInputEnd event.  The MacroRecorder disposes its event handlers on Closed and so does not see the final KeyInputEnd event.  Subsequent attempts to record a macro will fail because `_nestingDepth` is always positive.

dhmgit:
Fixes #2356

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -5228,6 +5228,12 @@ and IVimBuffer =
     [<CLIEvent>]
     abstract Closed: IDelegateEvent<System.EventHandler>
 
+    /// Raised after the buffer is closed AND any pending KeyInputEnd has been raised.
+    /// A buffer can be closed internally (i.e. while processing its own keyboard input), which will raise Closed -> KeyInputEnd -> InputClosed.
+    /// It can also be closed externally, which will raise Closed -> InputClosed.
+    [<CLIEvent>]
+    abstract InputClosed: IDelegateEvent<System.EventHandler>
+    
     inherit IPropertyOwner
 
 /// Interface for a given Mode of Vim.  For example normal, insert, etc ...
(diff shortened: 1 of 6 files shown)


Src/VimCore/VimBuffer.fs
```diff
@@ -639,6 +645,8 @@ type internal VimBuffer
 
         finally 
             _keyInputEndEvent.Trigger x args
+            if _isClosed && not x.IsProcessingInput then
+                _inputClosedEvent.Trigger x
```
jaredpar: This same logic needs to be added to `ProcessBufferedKeyInputs` as it can also drive key strokes.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -5228,6 +5228,12 @@ and IVimBuffer =
     [<CLIEvent>]
     abstract Closed: IDelegateEvent<System.EventHandler>
 
+    /// Raised after the buffer is closed AND any pending KeyInputEnd has been raised.
+    /// A buffer can be closed internally (i.e. while processing its own keyboard input), which will raise Closed -> KeyInputEnd -> InputClosed.
+    /// It can also be closed externally, which will raise Closed -> InputClosed.
+    [<CLIEvent>]
+    abstract InputClosed: IDelegateEvent<System.EventHandler>
```
jaredpar: Instead of `InputClosed` I think we should call this `PostClosed`. This matches the naming convention `ITextBuffer` used for their events: `Changing`, `Changed` and `PostChanged`. That has symetry with the closing events here and the three variants are needed for the same reason.

PR #2361, opened 2018-09-03
Simple tooltip for the marks

### LINKED ISSUE #2354: Mark glyph tooltips
Think we should add tooltips to the marks glyphs we are displaying. Believe this will help with two issues:

1. The glyph currently caps the display at three marks. This is practical given there is limited space here. The tooltip though can contain all of the marks. 
1. Helps explain what the glyphs are. Had a couple of coworkers ask what all of the `.[]` were showing up in the well after updating VsVim. Once I pointed it it was the marks they had they were pretty happy about the feature.

jaredpar:
Adds a simple tooltip display for our marks. Helps in a couple of cases:

1. The glyph currently caps the display at three marks. This is practical given there is limited space here. The tooltip though can contain all of the marks. 
1. Helps explain what the glyphs are. Had a couple of coworkers ask what all of the `.[]` were showing up in the well after updating VsVim. Once I pointed it it was the marks they had they were pretty happy about the feature. 

closes #2354


zstarkebaumcalico:
I'm stoked about this feature. Thanks for the hard work.

PR #2368, opened 2018-09-14
Left mouse button

### LINKED ISSUE #1419: Virtualedit inconsistencies with mouse handling
I found a couple of issues with virtualedit mouse handling that don't seem to match gvim's behaviour, although I might be misunderstanding how it should work.

If virtualedit is set to "" (I only found out about the option reading through the VsVim source and finding AdjustCaretForVitualEdit(), so I hadn't set it to my liking yet), you can still get the behaviour of "onemore" by clicking past the end of the line with the mouse in normal mode. However if you're in visual mode, clicking past the end of a line - say to lose the selection - will place the caret on the last character as it does in gvim.

This behaviour also pointed out a different bug to me, although once I set virtualedit="onemore" I think this bug is valid also with respect to that. In gvim using the `%` command while past the end of the line in the "onemore" column will still match the last character of a line, however in VsVim the search for a matching item goes strictly forward.

A potential fix for the `%` command that seems to work for me is as below, but there might be a better way of doing this!

``` diff
diff --git a/Src/VimCore/MotionUtil.fs b/Src/VimCore/MotionUtil.fs
index 8b60bab..ab37580 100644
--- a/Src/VimCore/MotionUtil.fs
+++ b/Src/VimCore/MotionUtil.fs
@@ -691,7 +691,11 @@ type MatchingTokenUtil() =

         let line = SnapshotPointUtil.GetContainingLine point
         let lineText = SnapshotLineUtil.GetText line
-        let column = point.Position - line.Start.Position
+        let column =
+            if point.Position = line.End.Position then
+                max 0 (line.End.Position - line.Start.Position - 1)
+            else
+                point.Position - line.Start.Position
         let found =
             match x.FindMatchingTokenKindCore lineText column with
             | None -> None
```

Thank you kindly for a great extension, if you need any more information to reproduce this let me know :smile:.

### LINKED ISSUE #1682: multiple-lines are selected when I drag the mouse on the same line
Hi, I really appreciate for this wonderful Vim plugin. I am quite amazed by such a good compatibility with gVim editor. However, I found the following issues that are quite annoying in my daily job:
(1) When I drag the mouse to select a portion of text, multiple lines are automatically selected even though the mouse movement is on the same line.
(2) Vim's selection in command line mode is consistently inclusive, but VsVim doesn't look so. Below is an example with one line of sample text:  
     xyz.abc

When I double-click xyz:
(gVim) xyz is marked and the block cursor is at z. yank and paste results in xyz
(VsVim) xyz is marked and the block cursor is at the period. yank and paste results in xyz though.
--> block cursor is NOT inclusive

When I type vw at x, both gvim and VsVim show the block cursor at z, and yank catches xyz
--> block cursor is now inclusive.

### LINKED ISSUE #2366: visual selection with mouse doesn't include last character
The behavior of selecting text with the mouse in VsVim 2.6.1189 differs a bit from gVim 8.1.1.  

In gVim, when highlighting some text with the mouse, the last character that the cursor hovers over is included in the selection.  In VsVim, the last character is not included.  

Here is an example of what I mean (selection, yank, put):

**gVim:**
gvim2

**VsVim:**
vsvim2

Thanks!

### LINKED ISSUE #2367: Double-click on word incorrectly places block cursor
_Update: corrected problem description_

With `:behave xterm`

Upon double-clicking a word with the mouse:
* gVim selects the word with the block cursor over the last character
image

* VsVim selects the word and places the block cursor over the first character following the word
image

Possibly related to #2366

ricksladkey:
#### Changes

- Implement left mouse button down (move caret)
- Implement left mouse button drag (select text)
- Implement left mouse button double-click (select word or matching token)
- Implement left mouse button triple-click (select line)
- Implement left mouse button quadruple-click (select block)

#### Issues

- Fixes #1419
- Fixes #1682
- Fixes #2366
- Fixes #2367


ricksladkey:
Ha, ha, so much work for so little gain! I will add more tests as I recover my enthusiasm...

ricksladkey:
This issue is full of corner cases because of all the vim settings and non-deterministic demons because of the complexities of real-life mouse events. I've tried to capture all the desired functionality in tests, but if anyone could test it out by hand, that would be very much appreciated!

dhmgit:
@ricksladkey, I merged your changes into my dogfood branch and will be testing as I work my day job.  I'll post comments with any issues I discover.

jaredpar:
Done with review pass (iteration 17)

Src/VimCore/CommandUtil.fs
```diff
@@ -102,6 +102,9 @@ type internal CommandUtil
 
     let mutable _inRepeatLastChange = false
 
+    /// The last mouse down position before it was adjusted for virtual edit
+    let mutable (_leftMouseDownPoint: SnapshotPoint option) = None
```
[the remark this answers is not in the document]
ricksladkey: Heh, I know F# syntax intuitively, not rigorously!

Src/VimCore/KeyNotationUtil.fs
```diff
@@ -183,6 +195,9 @@ module KeyNotationUtil =
                         | 'a' -> VimKeyModifiers.Alt ||| modifier |> Some
                         | 'm' -> VimKeyModifiers.Alt ||| modifier |> Some
                         | 'd' -> VimKeyModifiers.Command ||| modifier |> Some
+                        | '2' -> VimKeyModifiers.Double ||| modifier |> Some
+                        | '3' -> VimKeyModifiers.Triple ||| modifier |> Some
+                        | '4' -> VimKeyModifiers.Quadruple ||| modifier |> Some
```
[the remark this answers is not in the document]
ricksladkey: Not that I could find, but it certainly is implied by the existence of documented bindings like `<2-LeftMouse>. To test whether `2` was a modifier or part of a "key" name, I tried to create a mapping in gvim for `<2-A>` (double-click letter `A`) and gvim accepts it and reports it that way in the `:map` command.

The command `:help double-click` describes double/triple/quadrule click.

ricksladkey:
The mouse is an "all or nothing" deal: once we implement part of it, we have to implement all of it.

Strangely, this almost adds nothing for me because I use `:behave mswin` and the Visual Studio behavior we were relying on was "good enough" to be fairly faithful to vim.

However, even for me, this does add several distinct new features:

- Double-click on open or close parens, etc.
- Double-click on the `#` of an `#ifdef`
- Triple-click enters line mode
- The mouse can no longer place the cursor after then end of the line without `:set virtualedit=onemore`
- The mouse can no longer be placed on the phantom line

It also allows us to write tests for all of this functionality.

And finally, I learned that you can double-click on help items in vim help to navigate to them!

dhmgit:
Overall, these changes are working very well in my testing.  I did, however, find a couple of behaviors that do not match gVim.

**Mouse extension in line mode**

Starting in VISUAL LINE mode, `<S-LeftMouse>` does not expand the selection over full lines.  In the following screenshot, I started in VISUAL LINE mode with line 3 selected, then shift+left-clicked the middle of the 4th line.
image

**Mouse drag in line and block modes**

Triple- or quadruple-click without the final button release (i.e. dudu[du]d) followed by mouse movement does not extend the selection.  The correct VISUAL mode is entered, but mouse pointer movement is not matched by the caret.

ricksladkey:
@dhmgit Thank you very much. That is just the kind of feedback I was hoping for.

Src/VimCore/CommandUtil.fs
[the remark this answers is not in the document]
jaredpar: Yeah I feel the same. I've made a few efforts at being more rigorous here and I'm getting a bit more crisp. Although I feel implicit line endings will always trip me up. Paren's aren't needed here though.

Src/VimCore/KeyNotationUtil.fs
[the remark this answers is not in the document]
jaredpar: Gotcha. At least I wasn't missing the documentation then.

jaredpar:
> The mouse is an "all or nothing" deal: once we implement part of it, we have to implement all of it.

Curious about this. I get why we need to handle double, triple click as all or nothing but unsure why we need to bring in drag to that as well.

ricksladkey:
@jaredpar Once we mark the left mouse button down event as handled, Visual Studio's "select text" state machine no longer gets entered. I had originally only wanted to implement moving the caret with the mouse, but that broke drag-to-select, double-click, etc.

It turns out that every one of the various mouse events had vim-specific behavior that needed to be implemented anyway, so all the work that was done had to be done one way or another. The irony is that it was "mostly working" before with very little work on part due to Visual Studio infrastructure.

dhmgit:
Something I've noticed during testing (something which is consistent with gVim, but nevertheless a potential user complaint) is that dragging a selection is not possible.  Instead of the selection being moved by a left-click+drag, a new selection is started from the clicked position.

I rarely drag selections, so I'm not 100% sure that this function was available with VsVim enabled before this PR, but I seem to recall it working before.  That said, I wouldn't miss the drag capability, and it's inconsistent with gVim (and thus contravenes the design philosophy of this project), so it's probably not an issue.  I just wanted to make sure this was a known and deliberate behavior change rather than an accident.

ricksladkey:
@dhmgit Frankly, I was unaware of the cut/paste via drag feature of Visual Studio. Vim seems to have no such equivalent with any combination of the applicable settings. My understanding is also that fidelity with vim is the overarching project principle and so I'm satisfied that being vim-compatible justifies the new behavior.

ricksladkey:
Sorry, I keep trying to improve these changes!

Src/VimCore/CommandUtil.fs
```diff
@@ -1084,6 +1086,22 @@ type internal CommandUtil
         let span = action snapshotData
         BufferGraphUtil.MapSpanDownToSingle _bufferGraph span x.CurrentSnapshot
 
+    /// Extend the selection for a mouse click
+    member x.ExtendSelectionForMouseDrag (visualSpan: VisualSpan) =
```
jaredpar: >(visualSpan: VisualSpan) [](start = 41, length = 24)

Why does this take a VisualSpan and not use it?

Src/VimCore/CommandUtil.fs
```diff
@@ -1084,6 +1086,22 @@ type internal CommandUtil
...
+        // Double-clicking creates a problem because the caret is moved to the
+        // end of what was selected rather than directly under the mouse
+        // pointer.  Prevent accidentally dragging after a double-click by
+        // moving the caret only if the mouse position is over a different
+        // snapshot point than it was when the mouse was previously clicked.
+        x.MoveCaretToMouseIfChanged() |> ignore
+        CommandResult.Completed ModeSwitch.NoSwitch
+
+    /// Extend the selection for a mouse drag
+    member x.ExtendSelectionForMouseClick (visualSpan: VisualSpan) =
```
jaredpar: >(visualSpan: VisualSpan) [](start = 42, length = 24)

Why does this take a VisualSpan and not use it?

Src/VimCore/CommandUtil.fs
```diff
@@ -102,6 +101,9 @@ type internal CommandUtil
 
     let mutable _inRepeatLastChange = false
 
+    /// The last mouse down position before it was adjusted for virtual edit
```
jaredpar: This will be kept in sync for all of the moves that we do. But what about mouse moves that happen externally? For instance another extension doing the move or the caret moving due to an insert? Does it need to be cleared for those?

Src/VimCore/CommonOperations.fs
```diff
@@ -83,6 +84,60 @@ type internal CommonOperations
 
     member x.CaretLine = TextViewUtil.GetCaretLine _textView
 
+    /// The snapshot point in the buffer under the mouse cursor
+    member x.MousePoint =
```
jaredpar: How does this differ from the typical way of looking at the caret point? Is it just applying all of the vim logic on top of it?

jaredpar:
Done with review pass (iteration 32)

Src/VimCore/CommandUtil.fs
ricksladkey: At one point it manipulated the visual span, but eventually I realized it is like any other caret motion that simply affects what is selected. Will remove.

Src/VimCore/CommandUtil.fs
ricksladkey: Will remove.

Src/VimCore/CommandUtil.fs
ricksladkey: Our only requirement is that a mouse drag (which can only occur when the mouse button is down) is preceded by a mouse click, and the mouse device assures us of that. The time to clear it is on mouse up but that contract just mentioned makes that unnecessary.

Other extensions or insert mode can only move the caret, not the mouse and this variable is unrelated to the caret.

If not clearing it just feels wrong, I can add `<LeftUp>` bindings for all modes. In fact, I'll do that because it is remotely conceivable that the mouse up event occurs at a position for which a drag event was not generated and the drag is only to provide feedback between down and up.

Src/VimCore/CommonOperations.fs
ricksladkey: It's not a caret point; it's a buffer point. What this API does is effectively translate the mouse position as supplied by the mouse device in pixel coordinates into a setting-sensitive virtual buffer point.

ricksladkey: I see now that I had done all of the work that made this API available to both command util and insert util, but then changed it so that insert util no longer needs it. So I could now move `x.MousePoint` back into command util where it was before.

ricksladkey:
@jarepar Was I supposed to merge this (and any left mouse fixups) into staging?

jaredpar:
Let’s do master for now. Given it was fairly new I wanted to give it a little more dogfooding time.

PR #2370, opened 2018-09-15
Fix tag block selection edge cases

### LINKED ISSUE #2369: Tag block selection edge cases
VsVim behavior does not match gVim in the following cases:

### Case 1
Starting with fully selected inner span, `at` selects grandparent tag instead of parent.
Key sequence: `vitat`

| Before | After |
| --- | --- |
| image | image |
| image | image|

### Case 2
Starting with caret in empty tag, `vit` does not expand selection.
Key sequence: `vit`

| Before | After |
| --- | --- |
| image | image |
| image | image |

### Case 3
Starting with a partially selected tag and `:set selection=inclusive`, `vit` selects the parent tag instead of the current tag.
Key sequence: `vit`

| Before | After |
| --- | --- |
| image | image |
| image | image |

dhmgit:
Fixes #2369

### DIFF
--- a/Test/VimCoreTest/VisualModeIntegrationTest.cs
+++ b/Test/VimCoreTest/VisualModeIntegrationTest.cs
@@ -1680,8 +1680,21 @@ namespace Vim.UnitTest
                     Assert.Equal(span, _textView.GetSelectionSpan());
                 }
                 
+                [WpfTheory]
+                [InlineData("inclusive")]
+                [InlineData("exclusive")]
+                public void SelectPartialInnerExpandAll(string selectionSetting)
+                {
+                    Create("<a><b><c/></b></a>");
+                    _globalSettings.Selection = selectionSetting;
+
+                    _textView.MoveCaretTo("<a><b><c".Length);
+                    _vimBuffer.Process("vlit");
+                    Assert.Equal("<c/>", _textView.GetSelectionSpan().GetText());
+                }
+
                 [WpfFact]
-                public void SelectInnerExpandAll()
+                public void SelectFullInnerExpandAll()
                 {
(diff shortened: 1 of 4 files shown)


Test/VimCoreTest/VisualModeIntegrationTest.cs
```diff
@@ -1679,6 +1679,31 @@ public void AllSimpleSingleLine()
...
+                    Create("<parent>", "<child>", "<grandchild />", "</child>", "</parent>");
+
+                    var initialPosition = _textBuffer.GetLineFromLineNumber(2).Start;
+                    _textView.MoveCaretTo(initialPosition);
+                    _vimBuffer.Process("vitat");
+                    Assert.Equal(_textBuffer.GetLineRange(1, 3).Extent, _textView.GetSelectionSpan());
+                }
+
+                [WpfFact]
+                public void EmptyTag_SelectInnerExpandInner()
```
ricksladkey: Can you add a `:set selection=exclusive` flavor of this test? In my testing, that results in zero-width selection after the first step.

dhmgit: I've expanded the test case to cover both selection modes.  Thanks for the suggestion -- I always run "inclusive" mode and didn't realize a zero-width selection was even possible.

ricksladkey:
@jaredpar I've reviewed, approved and merged this and labeled the issue for 2.6.0. I feel comfortable doing that for small/straightforward PRs, if it makes your life easier. If you want to be a reviewer on every PR, in the future I can just review but not merge.

jaredpar:
@ricksladkey that's fine by me. I sent you an email a few weeks ago about this idea. Did I send it to the wrong address (the one listed on your blog)?

PR #2371, opened 2018-09-17
C # script execution function with VsVim.

TKumadaki:
[VsVimScripts.zip](https://github.com/jaredpar/VsVim/files/2387352/VsVimScripts.zip)

Add C # script execution function with VsVim.
I fixed #2293

**Changes**
Define a method on the host for running scripts IVimHost.RunCSharpScript
Define a MEF interface for running scripts: ICSharpScriptExecutor
Define CSharpScriptExecutor within VsSpecific and execute C # script in this class.
Overall rewrite.

**Preparation**
I do not include script files in PR.
Please use attached file.
I also publish it in my repository.
[TKumadaki/VsVimScripts](https://github.com/TKumadaki/VsVimScripts)

**"csx" command**
Run the C # script.
this command reuse the object.
So script can save the last state with a static variable.

**"csxe" command**
Run the C # script.
This command creates an object every time.
It is used for debugging.

The difference between the two commands can be confirmed by `csx calledcount` and `csxe calledcount`.

**script file**
`util.csx` is a library script.
`simplesurround.csx,selectionmove.csx,methods.csx `is almost the same as last time.
`solution.csx` has added some functions.

**Metadata reference problem**
I intended to specify the installation path of VsVim in the metadata search path.
As below

```
var smr = ScriptMetadataResolver.Default
    .WithBaseDirectory(scriptPath)
    .WithSearchPaths(searchPath);

var so = ScriptOptions.Default
      .WithSourceResolver(ssr)
      .WithMetadataResolver(smr);

```
Because it is not known beforehand which dll to use in the script.

However, when writing the following code in the script, a CS 1705 error occurred.

```
#r "Microsoft.VisualStudio.Text.UI.Wpf.dll"
public bool TryGetWpfTextView(IVimBuffer vimBuffer, out IWpfTextView textView)
{
    textView = vimBuffer?.TextView as IWpfTextView;
    if (textView == null)
    {
        return false;
    }
    return true;
}

```
The cause is Microsoft.VisualStudio.Text.UI.Wpf.dll in the installation path of VsVim
It was because the version of Vs2017 project is different from what is referenced.

So I dynamically acquire the dll used by the project of Vs 2017
I decided to add a reference.
in this way.

```
var asm = AppDomain.CurrentDomain.GetAssemblies().Where(assembly =>   
!assembly.IsDynamic &&  
!string.IsNullOrWhiteSpace(assembly.Location));

```
This method imports more than 200 dlls.
Also, if Microsoft.VisualStudil.Shell.XX.X.dll
Multiple imported, CS0433 occurs in script.

I judged that dynamic import is unsafe.

Currently it is as follows.

```
var asm = new List<Assembly>();

asm.Add(typeof(EnvDTE.AddIn).Assembly); //EnvDTE.dll
asm.Add(typeof(EnvDTE80.Breakpoint2).Assembly); //EnvDTE80.dll
asm.Add(typeof(EnvDTE90.Module).Assembly); //EnvDTE90.dll
asm.Add(typeof(EnvDTE100.Solution4).Assembly); //EnvDTE100.dll

//asm.Add(typeof(Microsoft.VisualStudio.Shell.Package).Assembly); //Microsoft.VisualStudil.Shell.15.0 is imported. 
//CS0433 error occurs when multiple versions of Microsoft.VisualStudil.Shell are imported.
//So commented out.

asm.Add(typeof(Microsoft.VisualStudio.Text.ITextSnapshot).Assembly); //Microsoft.VisualStudio.Text.Data.dll
asm.Add(typeof(Microsoft.VisualStudio.Text.Document.ChangeTag).Assembly); //Microsoft.VisualStudio.Text.Logic.dll
asm.Add(typeof(Microsoft.VisualStudio.Text.Editor.IWpfTextView).Assembly); //Microsoft.VisualStudio.Text.UI.Wpf.dll

asm.Add(typeof(Vim.IVim).Assembly); //VimCore.dll
asm.Add(typeof(Vim.UI.Wpf.IBlockCaret).Assembly); //VimWpf.dll
asm.Add(typeof(Vim.VisualStudio.ISharedService).Assembly); //Vim.VisualStudio.VsInterfaces.dll,Microsoft.VisualStudil.Shell.11.0 is imported.
asm.Add(typeof(Vim.VisualStudio.Extensions).Assembly); //Vim.VisualStudio.Shared.dll,Microsoft.VisualStudil.Shell.11.0 is imported.

var so = ScriptOptions.Default
      .WithSourceResolver(ssr)
      .WithMetadataResolver(smr)
      .WithReferences(asm);
```

I do not like this part.
However, I can not think of any other method.
Please tell me if there are other methods.


TKumadaki:
I made the following changes.

In previous PR I did not understand the mechanism of VsSpecific.
I modified CSharpScriptExecutor to be included in SharedService to correct this bug.
But this is a different change from the review of @jaredpar.
I may not understand VsSpecific yet.

I included the 'PublicAssemblies' and 'PrivateAssemblies' folders in the dll's search path.
This made it possible to minimize the assembly given to the method of WithReferences.

I decided to load 'Microsoft.CodeAnalysis.CSharp.Scripting.dll' from Visual Studio's directory dynamically.
This is due to the following reasons.

- It is not necessary to include dll's in Vsix.
- Even if there is a problem with CSharpScriptExecutor, there is little influence on VsVim.

As a result, CSharpScriptExecutor strongly depends on the structure of VisualStudio's directory.

TKumadaki:
Hi @jaredpar.
I think that it became ready for review with the above change.
Please let me know if there is any part that you can not review this PullRequest.
I would like to improve it.

If this PR is reviewable,and if you review it at your leisure
I'm glad.
:-)

Src/VimCore/VimBuffer.fs
```diff
@@ -702,7 +702,23 @@ type internal VimBuffer
  
             if _isClosed && not x.IsProcessingInput then
                 _postClosedEvent.Trigger x   
-    
+
+    /// Process from C# script
+    member x.ProcessFromScript (keyInput: KeyInput) =
```
jaredpar: Why is this still needed?

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
```diff
@@ -0,0 +1,172 @@
...
+                if (_cSharpScriptCreate == null)
+                {
+                    assemblyName = "Microsoft.CodeAnalysis.CSharp.Scripting.dll";
+                    assemblyPath = Path.Combine(baseDirectory, assemblyName);
+                    if (!File.Exists(assemblyPath))
+                    {
+                        vim.ActiveStatusUtil.OnError($"{assemblyName} not found.");
+                        return;
+                    }
+                    var asm = Assembly.LoadFile(assemblyPath);
```
jaredpar: Rather than looking for this on disk have you tried just using `Assembly.Load(AssemblyName)`? That is the "better" way to load assemblies (avoids load context issues). In this particular case the path should be on the probing path of Visual Studio so this should work I believe.

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
```diff
@@ -0,0 +1,172 @@
...
+                                _cSharpScriptCreate = mi;
+                                break;
+                            }
+                        }
+                    }
+                    if(_cSharpScriptCreate == null)
+                    {
+                        vim.ActiveStatusUtil.OnError($"method 'CSharpScript.Create' not found.");
+                        return;
+                    }
```
jaredpar: Rather than having all this reflection why not add references to the C# scripting API directly. If we make sure to reference the 1.0 version, which is what you're logically doing here, those will load fine in later versions of Visual Studio. Because it's a supported API set the have the correct binding redirects in the box.

TKumadaki: As you advised, I did not need to do something like this.
After deciding what to do with ProcessFromScript, I would like to fix it.

Src/VimCore/VimBuffer.fs
jaredpar: Why does `Process` not work here? I've thought about this for a bit and can't see what's wrong here. Is it the recursive event handling that you're worried aboutal

TKumadaki: I may be wrong
Here I think that it is as follows.

1. Type j.
2. An OnKeyInputStart event occurs.
3. Since "e.Handled = true", Vsvim does nothing.
4. "vimBuffer.Process ('k')" is executed in the OnKeyInputStart event.
5. An OnKeyInputStart event occurs.
6. Since "e.Handled = true" is set, for the key of "k" Vsvim does nothing.

TKumadaki: If I make the following changes, the code will be executed.

```
#r "Microsoft.VisualStudio.CoreUtility.dll"

using Vim;
using Vim.Extensions;

var activeVimBuffer = Vim.ActiveBuffer;
if (activeVimBuffer.IsNone())
{
    Vim.ActiveStatusUtil.OnError(":Can not get VimBuffer");
    return;
}
IVimBuffer vimBuffer = activeVimBuffer.Value;
vimBuffer.KeyInputStart += OnKeyInputStart;

public void OnKeyInputStart(object sender, KeyInputStartEventArgs e)
{
    if (e.KeyInput.Char == 'k')
    {
        return;
    }

    e.Handled = true;
    if (e.KeyInput.Char == 'j')
    {
        var ki = KeyInputUtil.CharToKeyInput('k');
        //vimBuffer.ProcessFromScript(ki);
        vimBuffer.Process(ki); //Work.
    }
    else if (e.KeyInput.Key == VimKey.Escape)
    {
        vimBuffer.KeyInputStart -= OnKeyInputStart;
    }
}

```

TKumadaki: I think that there is no problem unless `Process` is used in `OnKeyInputStart` event.
However, I think that some users would like to use `Process` in the `OnKeyInputStart` event.

TKumadaki: I thought for a moment.
And if I modify the code as follows,
I noticed that it works.
So, I remove ProcessFromScript.

```
public void OnKeyInputStart(object sender, KeyInputStartEventArgs e)
{

    e.Handled = true;
    if (e.KeyInput.Char == 'j')
    {
        vimBuffer.KeyInputStart -= OnKeyInputStart;

        var ki = KeyInputUtil.CharToKeyInput('k');
        vimBuffer.Process(ki); //Work.

        vimBuffer.KeyInputStart += OnKeyInputStart;
    }
    else if (e.KeyInput.Key == VimKey.Escape)
    {
        vimBuffer.KeyInputStart -= OnKeyInputStart;
    }
}
```

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
```diff
@@ -0,0 +1,115 @@
...
+        private void Execute(IVim vim, CallInfo callInfo, bool createEachTime)
+        {
+            try
+            {
+                Script<object> script;
+                if (!TryGetScript(vim, callInfo.Name, createEachTime, out script))
+                    return;
+
+                var globals = new CSharpScriptGlobals(callInfo, vim);
+                script.RunAsync(globals).Wait();
```
jaredpar: Calling `.Wait` can introduce some interesting deadlock cases. Did you consider having the `Execute` method return a `Task` instead so the caller could potentially wait this in a safe fashion? I'm not 100% sure that's possible here but want to make sure it was considered.

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                vim.ActiveStatusUtil.OnError(string.Join(Environment.NewLine, ex.Diagnostics));
+            }
+            catch (Exception ex)
+            {
+                vim.ActiveStatusUtil.OnError(ex.Message);
+            }
+        }
+
+        private ScriptOptions GetScriptOptions(string scriptPath)
```
jaredpar: Believe this could be `static`

jaredpar:
It looks like you uploaded the same Microsoft.CodeAnalysis DLLs for both VS2017 and VS2019. Based on the version number I believe those are actually the DLLs for VS2015. The 1.1 version is what was shipped there. 

If that's the case we should just be able to put those in the VS2015 directory and reference them from all the environments. The DLLs are back compat and a part of the supported VS API set so the have appropriate app.config entries

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
TKumadaki: I did not recognize the possibility of deadlock.
I will verify this problem.

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
TKumadaki: I would like to fix this.
But I had never minded that private methods should be static.
Why should I do this?
Is it because memory can be saved by this change?

TKumadaki:
I'm sorry. This got the oldest one from NuGet Gallery.
My Visual Studio 2015 dll version is 1.3.1.0.

Should I do the following work?
Delete the dll from the directory of Vs2017 and Vs2019.
Place the dll of version 1.3.1.0 in the directory of Vs2015.
Refer to this from the project.

TKumadaki:
Hi @jaredpar
I have made a few changes, not wait for your reply.
Please let me know if it is a wrong change.

And as you pointed out
The following script will cause a deadlock.

```
using System;
using System.Runtime.InteropServices;
using System.Windows;
using System.IO;
using System.Threading.Tasks;

[DllImport("kernel32.dll")]
public static extern bool AllocConsole();

[DllImport("kernel32.dll")]
public static extern bool FreeConsole();

var count = await Task.Run(DeadLockMethod);
MessageBox.Show(count.ToString());

private int DeadLockMethod()
{
    int count = 0;
    try
    {
        AllocConsole();

        using (var st = Console.OpenStandardOutput())
        using (var sw = new StreamWriter(st))
        {
            for (count = 0; count < 100000; count++)
            {
                sw.WriteLine(count);
                sw.Flush();
            }
        }

        FreeConsole();
    }
    catch (Exception ex)
    {
        MessageBox.Show(ex.Message);
    }
    return count;
}
```

Deadlock does not occur with this fix.
However, the following scripts can be operated by the user during execution.
Is this movement accepted?

```
#r "Microsoft.VisualStudio.CoreUtility.dll"

using System.Threading.Tasks;
using Vim;
using Vim.Extensions;

var activeVimBuffer = Vim.ActiveBuffer;
if (activeVimBuffer.IsNone())
{
    Vim.ActiveStatusUtil.OnError("Can not get VimBuffer");
    return;
}
IVimBuffer vimBuffer = activeVimBuffer.Value;

vimBuffer.SwitchMode(ModeKind.Normal, ModeArgument.None);
for (var count = 0; count < 100; count++)
{
    var ki = KeyInputUtil.CharToKeyInput('j');
    vimBuffer.Process(ki);
    await Task.Delay(1000);
}
```

TKumadaki:
As System.Collections.Immutable.dll in VS2019 is new
When I compile my PR, MSB3277 came to occur.

In order to solve this problem, I did the following.

- Place C # API dll in VS2017, VS2019.

The version of System.Collections.Immutable.dll in my Visual Studio 2017 is 1.2.3.0.
Since C # API dll refers to this dll, it replaced it with a new version.

jaredpar:
@TKumadaki 

>  As System.Collections.Immutable.dll in VS2019 is new

Yes that will usually be the case. The version will change with every release of Visual Studio. We pretty much always need a different copy in every VS2015, VS2017, etc ... folder. If you sync to the latest in master you'll see I recently updated it for VS2019 preview 3.

Src/VsSpecific/VsSpecific/CSharpScriptExecutor.cs
```diff
@@ -0,0 +1,116 @@
...
+                  .WithSourceResolver(ssr)
+                  .WithMetadataResolver(smr)
+                  .WithReferences(asm);
+
+            return so;
+        }
+
+        private bool TryGetScript(IVim vim, string scriptName, bool createEachTime, out Script<object> script)
+        {
+            if (!createEachTime && _scripts.ContainsKey(scriptName))
```
jaredpar: Can be simplified to 

```
!createEachTime && _scripts.TryGetValue(scriptName, out script)
```

The version today uses double lookup with `Contains` and the indexer

Test/VsVimTest/ReferenceTest.cs
```diff
@@ -141,6 +141,11 @@ internal static ReferenceData GetReferenceData(AssemblyName assemblyName)
                 switch (name)
                 {
                     case "System":
+                    case "System.Collections.Immutable":
+                    case "Microsoft.CodeAnalysis":
+                    case "Microsoft.CodeAnalysis.CSharp":
+                    case "Microsoft.CodeAnalysis.CSharp.Scripting":
+                    case "Microsoft.CodeAnalysis.Scripting":
```
jaredpar: These are more like `ReferenceKind.Editor` than `ReferenceKind.Framework`. The `Framework` notation means it's a part of the .NET Framework and hence the version doesn't really matter. The .NET Framework takes care of all the compat concerns here as we're careful to target the correct minimum target framework in the project files. The C# scripting assemblies though have different versions on different versions of Visual Studio and we need to be careful to reference the right ones.

One big difference though is we can't make the major version of the assembly to a VS version like we can for editor assemblies. I think we need to create a new `ReferenceKind` value here and lets call it `ComponentVersioned`. The mapping from version number to VS version though is likely going to be pretty specific. It can't be generalized. The best way can do is map major versions of Microsoft.CodeAnalysis (2 -> 2017, 3 -> 2019). For System.Collections.Immutable we'll likely just need to just map the specific versions.

jaredpar: This is a weird area. If you get stuck on this let me know. It may be faster for me to just do the change locally and push an update to your PR.

---
In reply to: [259923068](https://github.com/jaredpar/VsVim/pull/2371#discussion_r259923068) [](ancestors = 259923068)

TKumadaki: I added a reference test for C# API.

Test/VsVimTest/ReferenceTest.cs
```diff
@@ -192,7 +199,19 @@ internal static ReferenceData GetReferenceData(AssemblyName assemblyName)
                     case "Microsoft.VisualStudio.Platform.WindowManagement":
                         return new ReferenceData(assemblyName, ReferenceKind.ShellNonVersioned, getVersionFromVersionNumber());
 
+                    case MicrosoftCodeAnalysis:
+                    case MicrosoftCodeAnalysisCSharp:
+                    case MicrosoftCodeAnalysisCSharpScripting:
+                    case MicrosoftCodeAnalysisScripting:
+                    case SystemCollectionsImmutable:
+
+                        return new ReferenceData(assemblyName, ReferenceKind.ComponentVersioned, vsVersion: null);
```
jaredpar: Don't think we should be taking this approach. Rather we should be mapping the versions of these DLLs to the VS version that they ship in. For instance for MS.CA 2.* ships in 2017, 3.* ships in 2019, etc ... That will make the rest of the validation remain unchanged.

TKumadaki: I made a correction, but I am not confident.
When I return from my company, I would like to continue working.

TKumadaki: I noticed that the comment of `ComponentVersioned` is not appropriate.

jaredpar: That's what I was looking for here. Think this is good now. 

---
In reply to: [260988572](https://github.com/jaredpar/VsVim/pull/2371#discussion_r260988572) [](ancestors = 260988572)

jaredpar: Yeah we should update that comment but overall this is good now. 

---
In reply to: [260989748](https://github.com/jaredpar/VsVim/pull/2371#discussion_r260989748) [](ancestors = 260989748)

Src/VsSpecific/Vs2017/Vs2017.csproj
```diff
@@ -8,7 +8,12 @@
     <DefineConstants>$(DefineConstants);VS_SPECIFIC_2017</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.CodeAnalysis.Scripting, Version=2.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
+    <Reference Include="Microsoft.CodeAnalysis, Version=2.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
+    <Reference Include="Microsoft.CodeAnalysis.CSharp.Scripting, Version=2.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
+    <Reference Include="Microsoft.CodeAnalysis.CSharp, Version=2.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
```
jaredpar: By using the 2.10 assemblies this code will only work on VS2017 update 9. Customers on VS2017 RTM will not be able to use this as they reference 2.0.0.0. That's a very small number of customers though. For a given version of Visual Studio only the RTM and latest update are supported. Overwelming majority of customers though are on the latest update. If anyone is one RTM and hits this we can re-evaluate. 

Commenting mostly in case anyone runs into this. Don't think it's an actual problem. We should leave as is.

TKumadaki:
I seem to have modified the code yesterday while you were reviewing.
If the modification is not appropriate, I will undo it.
And I added two comments.

TKumadaki:
I plan on doing business trips several times for two weeks from tomorrow.
Reply and correction may be delayed.

I believe that documentation on this function is necessary,
I made the following.

[C# scripting](https://github.com/TKumadaki/VsVim/blob/added-documentation-on-csharp-script/Documentation/C%23%20scripting.md)

My English is poor, I do not think that it explains well.
However, I think that this is better than nothing.

When this PR is merged, I would like to PR the documentation.

jaredpar:
Thanks for the contribution here! 

> My English is poor, I do not think that it explains well.
However, I think that this is better than nothing.

More than happy to help out with the wording here. I think it's a good tutorial on how to use this.

ricksladkey:
Congratulations contributor @TKumadaki!

TKumadaki:
@jaredpar
I am glad that I can make the function I want by myself.
Of course, that is because of your help.
Thank you.

@ricksladkey 
I want to thank you.
The positive opinion from you has been encouraging to me.

PR #2384, opened 2018-10-12
Remove telemetry

jaredpar:
The telemetry collected in VsVim is really only useful if it's on by
default. Given we're now in an environment where all telemetry must be
opt-in (GDPR) there really isn't much point in keeping this around
anymore.


ricksladkey:
Well, now I know what GDPR is!

PR #2386, opened 2018-10-15
Fix MEF composition error

### LINKED ISSUE #2249: CompositionFailedException: This part (Vim.Vim) cannot be instantiated
Recently when starting VS 2015, whether I create a new project or open an existing one I get a popup like this:

image

And in the [ActivityLog.xml](https://github.com/jaredpar/VsVim/files/2188392/ActivityLog.zip) is the exception in the subject.

Strangely, as far as I can tell nothing is broken or not working after that.

I can't tell you exactly when it started happening unfortunately, I was running a local build with a temp fix for #1777 for a long time, so I only recently jumped onto updated builds and started seeing it then. If it's possible to run older builds to bisect the problem I'm happy to do that I'm just not sure how from the VS extension gallery.

I was running with all extensions disabled and it still happens, so I don't think it's a conflict with something else.

jaredpar:
This fixes a MEF composition error that occurs more frequently on older
versions of Visual Studio. There appears to be a timing issue with
accessing the output window on the UI thread during composition.

Unfortunately I'm not able to see what is going wrong as it's happening
within the DTE layer. The fix though is pretty straight forward:

1. Delay the initialization of the Output Window to idle
1. Catch exceptions coming from that code and raise errors vs. letting
them be unhandled

closes #2249

### DIFF
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -21,6 +21,7 @@ using Microsoft.VisualStudio;
 using Microsoft.FSharp.Core;
 using Microsoft.VisualStudio.OLE.Interop;
 using EnvDTE80;
+using System.Windows.Threading;
 
 namespace Vim.VisualStudio
 {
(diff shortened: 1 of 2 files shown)


ricksladkey:
It sounds like you were able to repro the issue. That's great!

jaredpar:
Yeah finally reproduced this. I'm still not entirely sure what is needed to make it happen. I have a number of machines with VS2107 and VS2015 installed but only one reproduced this issue. Think it may have something to do with the solution state loading. Essentially whether or not the output window was visible when you last closed the solution or something along those lines. Either way though fix is basically  the same.

PR #2391, opened 2018-10-17
Option page support for marks

### LINKED ISSUE #2372: Option page support for marks display
Given the marks display is present in the UI think we should add an option page that helps control it's display. Mostly just surfacing the underlying option, `vsvim_hidemarks`, just as we do for a lot of other options.

One snag is that all of the settings we surface are global marks. Hence the effect of controlling them from a global UI is pretty straight forward: change in the global options UI and it affects every buffer. 

Controlling a local mark is a bit different. I think the user expectation is still that if I change the setting through the options UI then it should update my display everywhere. Yet it's a local mark so do we change only the active buffer, all of them, etc ...

Think about this I think there are a couple of options:

1. Make `vsvim_hidemarks` a global setting. 
1. Keep `vsvim_hidemarks` as a local setting but when it's changed in the UI manually update every buffer.

My intuition says to go with (1). My reasoning is that for me at least I can't think of when I would want two buffers two display a different set of marks. But I'm a UI minimalist and I could just be missing some key scenarios here. Wanted to see what others thought (particularly because I didn't implement this feature).

jaredpar:
Adds support for mark display in the options page. 

closes #2372

### DIFF
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -223,6 +223,7 @@ namespace Vim.VisualStudio
 
             _vsMonitorSelection.AdviseSelectionEvents(this, out uint selectionCookie);
             _runningDocumentTable.AdviseRunningDocTableEvents(this, out uint runningDocumentTableCookie);
+            _vimApplicationSettings.SettingsChanged += OnApplicationSettingsChanged;
 
             InitOutputPane();
         }
@@ -257,6 +258,11 @@ namespace Vim.VisualStudio
                     }
                 };
             } 
+        }
+
+        private void OnApplicationSettingsChanged()
+        {
+            _display
         }
 
         public override void EnsurePackageLoaded()
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -241,6 +241,7 @@ namespace Vim.VisualStudio
         private readonly IVsRunningDocumentTable _runningDocumentTable;
         private readonly IVsShell _vsShell;
         private readonly IProtectedOperations _protectedOperations;
+        private readonly SettingsSync _settingsSync;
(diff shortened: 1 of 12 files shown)


jaredpar:
@ricksladkey know this issue was assigned to you. Had some insomnia the other night and wanted to see how much work this was (ended up going down a bit of a rabbit hole). If you had an implementation you were working on would be happy to take that over this.

Src/VsVimShared/VsVimHost.cs
```diff
@@ -48,7 +49,8 @@ internal sealed class SettingsSource : IVimCustomSettingSource
             private const string UseEditorDefaultsName = "vsvim_useeditordefaults";
             private const string UseEditorTabAndBackspaceName = "vsvim_useeditortab";
             private const string UseEditorCommandMarginName = "vsvim_useeditorcommandmargin";
-            private const string CleanMacros = "vsvim_cleanmacros";
```
ricksladkey: Ha, ha, I noticed this too.

Src/VsVimShared/VsVimHost.cs
```diff
@@ -64,76 +66,165 @@ internal static void Initialize(IVimGlobalSettings globalSettings, IVimApplicati
...
+                    case UseEditorTabAndBackspaceName:
+                    case UseEditorCommandMarginName:
+                    case CleanMacrosName:
+                        return SettingValue.NewToggle(false);
+                    case HideMarksName:
+                        return SettingValue.NewString("");
+                    default:
+                        Debug.Assert(false);
+                        return SettingValue.NewToggle(false);
+               }
```
ricksladkey: Yeah, one of the areas I was getting bogged down in was how best to handle a new non-toggle custom setting.

jaredpar:
> Nervously looking forward to the release.

I feel like that before every release :smile:

PR #2394, opened 2018-10-19
Drag by word

### LINKED ISSUE #2383: Double-click to select word then drag doesn't continue to select whole words
In vim if you double-click the mouse, then hold the second click and drag, it will continue to select whole words (this is a typical behaviour in windows at least, not sure about other OSs). At the moment on VsVim the drag will select character-wise.

I think this is probably related to #2368 since before as I understand it was using the VS mouse handling.

Also additionally - I don't think this is vim behaviour, but previously on VS you'd get the same if you control-click and then drag (word-wise selection). It would be nice to get that too since at the moment it just doesn't drag at all. I can understand though if that's out of scope as that's not original vim behaviour, I just have some muscle memory to overcome :smile:.

ricksladkey:
#### Issues

- Fixes #2383

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -1092,9 +1092,9 @@ type internal CommandUtil
     /// Extend the selection for a mouse click
     member x.ExtendSelectionForMouseDrag (visualSpan: VisualSpan) =
 
-        /// Reset the anchor point by switching to the appropriate visual mode
+        /// Change the anchor point by switching to the appropriate visual mode
         /// with a modified visual selection
-        let resetAnchorPoint (anchorPoint: SnapshotPoint) =
+        let changeAnchorPoint (anchorPoint: SnapshotPoint) =
             let visualKind = VisualKind.Character
             let caretPoint = x.CaretPoint
             let tabStop = _localSettings.TabStop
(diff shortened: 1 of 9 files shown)


Src/VimCore/CommandUtil.fs
```diff
@@ -1087,20 +1090,134 @@ type internal CommandUtil
...
+                    |> SnapshotPointUtil.GetChar
+                let previousPointChar =
+                    point
+                    |> SnapshotPointUtil.SubtractOne
+                    |> SnapshotPointUtil.GetChar
+                let isWhite = CharUtil.IsWhiteSpace pointChar
+                let wasWhite = CharUtil.IsWhiteSpace previousPointChar
+                let isWord = TextUtil.IsWordChar WordKind.NormalWord pointChar
+                let wasWord = TextUtil.IsWordChar WordKind.NormalWord previousPointChar
+                let isEmptyLine =
```
jaredpar: Consider adding a helper to `SnapshotPointUtil` that detects if the point is on an empty line.  I'm guessing that would be helpful elsewhere too.

ricksladkey:
I found a bug in this. This whole mouse handling business is the trickiest part of VsVim I have yet encountered...

I'll fix it and resubmit for review.

Issue #2400, opened 2018-10-24
Cannot change block caret color


farzadmf:
Hi,

Today, I'm facing an issue with VsVim where my block caret is now black. I tried to set the color in **Item Colors** in settings, but the weird thing is that when I change it and click OK, the color is still black, and when I come back to the settings window, the setting has been reverted back to its previous value.

Example:
1. I open VS, the caret in VS is black
2. I go to the settings and change `Block Caret Background` (which is now `Black`) to `Yellow` (just to test)
3. I click OK, the caret is still black
4. I come back to the settings and `Block Caret Background` is set to `Black` again!

One thing that might be important is that the first time I opened VS today, I saw an exception saying something like "the error may be because of an exception"; I'm used to exceptions appearing from time to time, but maybe this one caused something to change and now VsVim cannot save its settings?!

I appreciate if you help me

ricksladkey:
I'm sorry to hear about that. Can you perform a couple of experiments for me to narrow down where the problem is?

First, can you successfully change other VsVim colors from the settings window, like the command margin foreground/background?

![image](https://user-images.githubusercontent.com/2680524/47442652-ae02b980-d780-11e8-873b-1e10d493aa97.png)

Second, does trying to change the block caret background color through the Visual Studio settings work or behave any differently?

Tools -> Options -> Environment -> Fonts and Colors -> Display Items -> VsVim Block Caret -> Item Background

![image](https://user-images.githubusercontent.com/2680524/47442853-1b164f00-d781-11e8-9a33-17fd8e752a04.png)

adamskt:
I ran into the same situation today, and the VsVim options do not appear in the Fonts and Colors "Display Items" listing in Tools -> Options -> Environment:
![screenshot from 2018-10-24 09-30-04](https://user-images.githubusercontent.com/105415/47446148-717a9200-d76f-11e8-94b1-61647d886c25.png)

I did see the plugin exception prior and restarted Visual Studio.  The relevant lines from ActivityLog.xml are:
```
[...]

PR #2403, opened 2018-10-29
Make assembly version same as product version

jaredpar:
This change will enforce the assembly version information matches up
with the product version.

Having the assembly version match the product version can potentially
help track down installation / MEF issues. The assembly version is
included in most log files even on load failures. Seeing if the versions
match the product version can help identify what type of problem is
happening.

For instance if they don't match then there is some type of installation
or MEF corruption issue at hand.

Related to #2400

### DIFF
--- a/Src/VimCore/Constants.fs
+++ b/Src/VimCore/Constants.fs
@@ -36,12 +36,7 @@ module VimConstants =
     [<Literal>]
     let MainKeyProcessorName = "VsVim";
 
-#if DEBUG
-    [<Literal>]
-    let VersionNumber = "2.6.99.99 Debug"
-#else
     [<Literal>]
     let VersionNumber = "2.6.0.0"
-#endif
 
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,6 +12,10 @@
     <OutputPath>$(BinariesPath)$(Configuration)\$(MSBuildProjectName)</OutputPath>
     <BaseIntermediateOutputPath>$(BinariesPath)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
 
+    <AssemblyVersion>2.6.0.0</AssemblyVersion>
+    <FileVersion>$(AssemblyVersion)</FileVersion>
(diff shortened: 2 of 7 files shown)


Src/VimCore/Constants.fs
```diff
@@ -36,12 +36,7 @@ module VimConstants =
     [<Literal>]
     let MainKeyProcessorName = "VsVim";
 
-#if DEBUG
-    [<Literal>]
-    let VersionNumber = "2.6.99.99 Debug"
-#else
```
ricksladkey: I have been grateful for this feature, which prevents Visual Studio from constantly badgering me that my VsVim extension has a newer version. How does that conflict with the desire for versioned assemblies?

jaredpar: Mostly it just added extra pain in the infrastructure to set it up properly. I didn't think anyone was still using this (originally designed it for a different purpose). Let me restore this.

jaredpar: How do you typically install VsVim? After looking deeper at this I don't think this is related to updates. This should only really control the value displayed in Help About. The actual version of the VSIX is fixed in source.extension.manifest and that is what VS should be triggering update requests on. 

If you install from vsixgallery.com the versioning story is a bit different.

ricksladkey: I have my own master in my fork which I periodically reset to your upstream master, merge my various feature branches, build, and then xcopy deploy into VS2017. My experience with the nagging is maybe from other extensions. If the the manifest is upgraded to 2.7.0.0, that should stop the nagging unrelated to this version (which I see from the `:ver` command).

Yeah, so don't worry about it.

jaredpar: Gotcha. As long as we keep the version in master ahead of release then your scenario should continue to work. I have a similar process for updating which is why I haven't seen this in a while either.

jaredpar: That reminds me. Need to get the version in master moved to 2.7.

PR #2405, opened 2018-10-30
Move to 2.7.0 in main development

jaredpar:


jaredpar:
> This looks good. And I just realized the pattern: master is set to require approvals, but staging isn't!

That's an oversight on my part. Will fix shortly.

PR #2411, opened 2018-11-05
Use external edit mode when there are multiple carets

### LINKED ISSUE #2362: Interact with multi-caret selections
Visual Studio 15.8 introduces a new way to create multi-caret selections with `shift + alt + .` to select the next occurrence of the current word, which I find very useful.

This puts VsVim in VISUAL mode, which means `shift + i` and `shift + a` don't work. 

Can this be changed somewhere so that VISUAL BLOCK mode is used instead? Or is it possible to enter block mode while keeping the current selection? (`ctrl + q` on my machine enters block mode but only the original word remains selected)

_(This may be dumb question since I'm not very familiar with vim's visual mode)_

ricksladkey:
#### Issues

- Fixes #2362

#### Discussion

Note: I believe multiple carets were introduced partway into Visual Studio 2017 and that means the reference assembly API surface changed since it was released. I think that means that VsVim is only compatible if the user's Visual Studio is properly updated.

If this is unacceptable, it would be possible to extend the shared service assembly to differentiate sub-versions of VS2017.


ricksladkey:
This is even worse than I thought: we have to upgrade potentially ALL the VS2017 reference assemblies to get the unit tests to run. I suppose we call the function `ITextView2.MultiSelectionBroker.HasMultipleSelections` using reflection instead.

jaredpar:
>  I think that means that VsVim is only compatible if the user's Visual Studio is properly updated. If this is unacceptable, it would be possible to extend the shared service assembly to differentiate sub-versions of VS2017.

This is fine. Visual Studio itself doesn't really support older updates for a major version. Essentially if you filed a bug on 15.3 they'd fix it but put it in 15.9 (at the time of this comment). They won't patch 15.3 this point. The ask is for users to get to the latest. Hence it's reasonable to adopt the same stance here. 

> This is even worse than I thought: we have to upgrade potentially ALL the VS2017 reference assemblies to get the unit tests to run. I suppose we could call the function 

Indeed. These assemblies really have to be updated as a unit. Mixing and matching versions will lead to some errors. 

In the short term I think we should use reflection. I'm going to be working on a change next week that removes 2012 and 2013 support. At the same time I'm going to be updating all of the reference assemblies. Once that is done we can swap this to use non-reflection.

Src/VimCore/SelectionChangeTracker.fs
```diff
@@ -34,6 +34,9 @@ type internal SelectionChangeTracker
     /// key input and not in Visual Mode 
     let mutable _selectionDirty = false
 
+    /// Whether the text view has multiple carets
+    let mutable _hasMultipleCarets = false
```
jaredpar: Did you consider implementing this as an `IExternalEditAdapter` implementation? That is where my first attempt would've been to fix this. If it didn't work I'd be interested to know why.

ricksladkey: I did look at and try to comprehend the external edit adapter/monitor infrastructure. In my admittedly cursory review, I concluded that it was designed to use events for triggering entry to the mode. Furthermore, that core concept being probed was a specific kind of tag.

In the case of multiple carets, we are hinging off of selection changes. But because VsVim also reacts to external selection changes, we would have an ordering dilemma because it is critical that the selection changed event not be allowed to propagate to the selection change tracker otherwise it will switch to some visual or selection mode, clobber the selection and remove the multiple carets. As a result, I concluded that the selection change tracker needs to be aware of multiple carets and incorporate their presence into its overall (very careful) selection change handling.

This is not to say that the external edit adapter API could not be extended to handle this use case but that out-of-the-box it did not seem like good fit. However, if you now know that we need to react to selection changed events, and probably need the cooperation of the selection changed handler to prevent its interference, you might be able to imagine an approach that would be more general, less invasive, more encapsulated, etc.

jaredpar:
I stumbled across a trick the other day that should allow us to work around the need for new API DLLs if we're willing to use reflection: 

https://twitter.com/jcansdale/status/1059906364166848513

This is essentially taking advantage of the fact that MEF is contract name based, not type identity based. We should be able to import the multi caret component in the same fashion (leaving the import as optional) and then use reflection to grab the member in question.

Src/VimCore/SelectionChangeTracker.fs
jaredpar: I should clean up the API so it's clearer that it's not purely tag based. It definitely can be implemented on the basis of tags but that's optional. It can also be implemented with a simple on / off type of implementation (the later ReSharper interfaces are done that way). 

But I agree the ordering is problematic here based on the current implementation. I think I could refactor this a bit too work but that may be a bit overkill. The main purpose of that interface was to allow external components to easily plug into this system. This is pretty core so makes sense that it should go directly in this case.

ricksladkey:
@jaredpar That's a neat trick and we might have needed it, but the new `ITextView2` interface gives us the property `MultiSelectionBroker` so we don't need MEF to import it for us. I just pushed my new code that reverts the assembly reference update and replaces the shared service code with reflection.

However, during manual testing, I found an issue that surprised me. With the debugger I noticed that the switch from external edit mode back to normal/insert mode occurs when you type `<Esc>` , not when it returns to a single caret (which produces a selection changed event). That subverts the mutable state we were keeping to know that this particular instance of external edit mode was caused by multiple carets.

And then I found these other problems:

- The external edit banner message says it will return to insert mode
- It in fact goes back to normal mode no matter whether it was initiated from normal or insert mode

To summarize, I have already converted the PR to use reflection, but I wanted to fix these three problems before submitting it for review.

ricksladkey:
Closing this PR. See #2362.

Issue #2406, opened 2018-10-30
statusline no longer expanding properly?


philipmat:
I think this may be a defect introduced with 2.6.0, for I will admit I don't remember how this looked like earlier, however I want to believe that I would've been able to remember if it looked like the below (VS on top, Vim at the bottom)

![chrome_2018-10-30_12-51-19](https://user-images.githubusercontent.com/709928/47739393-45658200-dc43-11e8-9863-dd7cd1fa2796.png)

Upon further research, support for `statusline` and `laststatus` seemed to have been introduced with PR #1361 back in 2014 so now I'm really questioning my sanity. 

Am I correct in assuming that `:set statusline` supports no expansion markers (`%F` , etc)?  
Has anything changed recently that would cause the status line to show up where it didn't use to before?

jaredpar:
It's possible this changed, there was some work around expansion IIRC. What string did you use to use for expansion here?

Blackclaws:
I can confirm this bug appeared just now after updating to the latest version of visual studio. So something might have changed on their end? I now only see %F%m and so on in the statusline.

ricksladkey:
I am guessing that somehow a new behavior was introduced in PR #2244. Perhaps the status line is now visible when it previously wasn't, etc. However, the current behavior (as far as I can tell) is correct for the limited amount of what is implemented, i.e. the `laststatus` and `statusline` variables but no expansion.

In any case, nothing surrounding the status line or its expansion has changed in a long time and PR #2244 is the only change since VsVim 2.5.0 that affects the status line in any way. The explanation that seems to make the most sense is that expansion was never supported, but it wasn't obvious it wasn't being displayed at all.

philipmat:
I agree with @ricksladkey findings and that #2244 likely made the status line visible and thus emphasized the lack of expansion in `statusline`. Added a blurb to the SupportedFeatures.md to document this behavior.

PR #2414, opened 2018-11-07
Documenting statusline and laststatus settings

philipmat:
Documents the findings of #2406


jaredpar:
Thanks for documenting that behavior!

PR #2418, opened 2018-11-14
Support <C-[> to exit from command mode

### LINKED ISSUE #2417: Escaping from command mode via C-{
While entering anything during command mode (e.g from `/` or `?` or `:w` etc), pressing `<C-{>` does not exit the mode, requiring instead an explicit `<ESC>`.

Is there a workaround for this?

ricksladkey:
#### Issues

- Fixes #2417


Src/VimWpf/Implementation/CommandMargin/CommandMarginController.cs
```diff
@@ -554,6 +554,11 @@ internal void HandleCharEvent(TextCompositionEventArgs e)
                 var textChar = e.ControlText[0];
                 switch (textChar)
                 {
+                    case (char)0x1B: // <C-[>
```
jaredpar: Can this key be mapped though to a different combination?

PR #2426, opened 2018-11-19
Handle multiple changes in text change events

### LINKED ISSUE #2422: Parameter name: span was out of range of valid values [vs2015/2.6.0.0]
I started getting daily error messages after updating to 2.6.0.0.
I just triggered it by pressing ESC
------
[Microsoft Visual Studio]
Specified argument was out of the range of valid values.
Parameter name: span
----------
I believe it usually pops up when pressing ESC and it's always 'span'.

Unfortunately I can't reliably reproduce it. Any hints on how I can provide better information? (log files, etc)

-- john

ricksladkey:
#### Issues

- Fixes #2422

### DIFF
--- a/Src/VimCore/Modes_Insert_Components.fs
+++ b/Src/VimCore/Modes_Insert_Components.fs
@@ -129,11 +129,12 @@ type internal TextChangeTracker
 
     /// Start tracking the effective change
     member x.StartTrackingEffectiveChange() =
+        let snapshot = _textView.TextBuffer.CurrentSnapshot
         let caretPosition = _textView.Caret.Position
         let position = caretPosition.BufferPosition.Position
         _effectiveChangeData <-
             Some {
-                Snapshot = _textView.TextSnapshot
+                Snapshot = snapshot
                 LeftEdge = position
                 RightEdge = position
                 LeftDeletions = 0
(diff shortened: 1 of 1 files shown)


Src/VimCore/Modes_Insert_Components.fs
```diff
@@ -157,6 +152,7 @@ type internal TextChangeTracker
     member x.EffectiveChange =
         match _effectiveChangeData with
         | Some data ->
+            Contract.Assert(data.Snapshot = _textView.TextSnapshot)
```
jaredpar: Believe this will be false in some extreme corner cases. The implementation of `ITextView.TextSnapshot` doesn't directly call into `ITextBuffer.CurrentSnapshot` but rather uses a cached field. This cached field is only updated after `PerformLayout` occurs. In cases where `PerformLayout` is delayed, or this is called on an `ITextBuffer` that maps to multiple `ITextView` I'm pretty sure you could observe this failing. 

This should be safer: `_textView.TextBuffer.CurrentSnapshot`

jaredpar: Note: I'm not sure if what I wrote is true for every version of VS. For 2017 at least though this is true (just looked at the implementation in ILSpy)

---
In reply to: [235755886](https://github.com/jaredpar/VsVim/pull/2426#discussion_r235755886) [](ancestors = 235755886)

PR #2433, opened 2018-11-24
Use exclusive selection caret in insert mode

### LINKED ISSUE #2424: Delay before showing insertion caret after 'o' command
For the past few months, mostly using the head of VsVim master or related branches, I have noticed a significant delay after typing the `o` (open new line) command before the insertion caret is displayed. The delay is a least one second. My experience has mostly been with Visual Studio 2017.

Now that VsVim 2.6.0 has been released, I've tested 2.6.0 against Visual Studio 2015 and it does not have the same problem. There isn't anything obvious that VsVim could be doing to cause the delay and using the simple `i` or `a` command only sometimes causes the delay with VS2017.

Trying to understand the constellation of symptoms, I am sometimes able to induce the delay using `<C-S-F12>` from normal mode. When it happens the Visual Studio insertion caret takes over a second to appear even though VsVim (for the most part) is entirely out of the loop. As a result, I am suspecting some change in Visual Studio 2017 itself is the root cause of the problem.

Anyway, I was wondering if anyone else can reproduce this delay. It's just irritating enough to be distracting!

ricksladkey:
#### Issues

- Fixes #2424

### DIFF
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaretController.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaretController.cs
@@ -120,6 +120,8 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
                     kind = CaretDisplay.Invisible;
                     break;
                 case ModeKind.Insert:
+                    kind = CaretDisplay.Select;
+                    break;
                 case ModeKind.ExternalEdit:
                     kind = CaretDisplay.NormalCaret;
                     break;


Src/VimWpf/Implementation/BlockCaret/BlockCaretController.cs
```diff
@@ -120,6 +120,8 @@ private void UpdateCaretDisplay()
                     kind = CaretDisplay.Invisible;
                     break;
                 case ModeKind.Insert:
+                    kind = CaretDisplay.Select;
```
jaredpar: This will break users who have Visual Assist installed. They use `ITextCaret.IsHidden` to determine if they should / shouldn't be adding experiences like custom Intellisense. 

Maybe the fix here is to condition this behavior in some way and then disable it only for the times Visual Assist is installed. That information is already available in `IVisualAssistUtil.IsInstalled`. We could put something like `UseDefalutCaret` on `VimHost` and then have it return true only when Visual Assist is installed.

ricksladkey: Added this logic pretty much as you described.

PR #2443, opened 2018-12-12
Add settings cache to avoid querying mounted registry hive

oxilumin:
VsVim was pretty slow on my work PC and when I did analysis - it showed a lot of requests to VS's private registry hive every time I type a character. After adding cache VsVim is much more responsive in insert mode.


Src/VsVimShared/Implementation/Settings/InMemorySettingsCache.cs
```diff
@@ -0,0 +1,71 @@
...
+        [ImportingConstructor]
+        internal InMemorySettingsCache()
+        {
+            _stringsCache = new Dictionary<string, string>();
+            _booleansCache = new Dictionary<string, bool>();
+        }
+
+        void ISettingsCache.UpdateCache(string propertyName, string value)
+        {
+            if (!_stringsCache.ContainsKey(propertyName))
```
jaredpar: This could be simplified to `_stringsCache[propertyName] = value`

Src/VsVimShared/Implementation/Settings/InMemorySettingsCache.cs
```diff
@@ -0,0 +1,71 @@
...
+        void ISettingsCache.UpdateCache(string propertyName, bool value)
+        {
+            if (!_booleansCache.ContainsKey(propertyName))
+            {
+                _booleansCache.Add(propertyName, value);
+            }
+            else
+            {
+                _booleansCache[propertyName] = value;
+            }
```
jaredpar: This could be simplified to `_booleansCache[propertyName] = value`

Src/VsVimShared/Implementation/Settings/InMemorySettingsCache.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (!_booleansCache.ContainsKey(propertyName))
+            {
+                result = default;
+
+                return false;
+            }
+
+            result = _booleansCache[propertyName];
+
+            return true;
```
jaredpar: This is doing a double lookup at the moment: `Contains` then the indexer. It can be simplified to a single lookup with the following pattern: `return _booleansCache.TryGetValue(propertyName, out result);`

Src/VsVimShared/Implementation/Settings/InMemorySettingsCache.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (!_stringsCache.ContainsKey(propertyName))
+            {
+                result = default;
+
+                return false;
+            }
+
+            result = _stringsCache[propertyName];
+
+            return true;
```
jaredpar: This is doing a double lookup at the moment: `Contains` then the indexer. It can be simplified to a single lookup with the following pattern: `return _stringsCache.TryGetValue(propertyName, out result);`

Src/VsVimShared/Implementation/Settings/VimApplicationSettings.cs
```diff
@@ -38,6 +38,7 @@ internal sealed class VimApplicationSettings : IVimApplicationSettings
 
         private readonly WritableSettingsStore _settingsStore;
         private readonly IProtectedOperations _protectedOperations;
+        private readonly ISettingsCache _settingsCache;
```
jaredpar: Rather than having consumers deal with both `ISettingsCache` and `WritableSettingsStore` did you just consider adding an abstraction over `WritableSettingsStore`? Essentially say `CachableSettingsStore` that has the same interface as `WritableSettingsStore` but does the in memory caching you wrote?

jaredpar:
Thanks for taking the time to run a profile here and do the analysis. What setting did you see getting queried all the time. It's possible that should also just be locally cached vs queried all the time.

oxilumin:
Hi @jaredpar . Most of the settings in settings store were queried regularly (50к queries to registry within ~10 seconds of typing, about 15 queries for each time the setting is checked).

After the change there is only one check per variable at start and procmon shows no registry activity while typing.

I've got rid of code duplication in settings file for different types and changed it a bit so now all the settings are queried through the cache the same way and all the types including binding collection are cached (I think ideally that class could be autogenerated in runtime from interface, name attributes and default config).
Also I've added some tests that test cache in isolation.

I hope you don't mind adding ability to compile with VS2019 Preview (until it's released the settings are equal to VS2017). With this PR It builds well in both VS2017 and VS2019.

Directory.Build.props
```diff
@@ -47,13 +47,20 @@
     <DefineConstants>$(DefineConstants);VS2017</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup>
+  <PropertyGroup Condition=" '$(VsVimTargetVersion)' == '16.0' ">
```
jaredpar: Please move the VS2019 changes into a separate PR. Want to keep this focused on the perf problem. Makes it easier to back port if it becomse necessary.

jaredpar:
> Most of the settings in settings store were queried regularly (50к queries to registry within ~10 seconds of typing, about 15 queries for each time the setting is checked).

Interesting. That's definitely not what I intended when I originally added this. I'm going to experiment a bit to see if there are settings which are getting queried in a way that's improper. The caching is still likely a good fix here though overall. 

> I hope you don't mind adding ability to compile with VS2019 Preview (until it's released the settings are equal to VS2017

I'd prefer we move this to a separate PR. Partly to keep PRs single purpose. The other part is to make it easier to back port this fix to earlier versions if it's needed.

jaredpar:
Yeah there are a fwe settings we query quite regularly. Particularly whether or not tracing is enabled (happens on every single log call). Didn't realize this is implemented by directly querying the registry. Assumed there was some level of caching going on here in the VS layer but appears not.

Directory.Build.props
oxilumin: Agree

oxilumin:
I personally think this has to be done on Visual Studio level, not in plugin, I'll try to touch base with them. They probably won't implement cache of actual objects (like in case of VsVim - removed bindings collection) and rather do it on string/bool level though.

Anyway - most likely it'd take some time so we could have this fix in the plugin for now.

Src/VsVimShared/Implementation/Settings/CachedSettingsStore.cs
```diff
@@ -0,0 +1,93 @@
...
+            return cache.TryGetValue(key, out value) ||
+                   TryUpdateFromUnderlyingStore(key, out value, defaultValue);
+        }
+
+        private bool TryUpdateFromUnderlyingStore<T>(string key, out T value, T defaultValue)
+        {
+            var foundInUnderlyingStore = _underlyingStore.Check(key, out value, defaultValue);
+
+            if (foundInUnderlyingStore)
+            {
```
jaredpar: Consider adding `Debug.Assert(!_notFoundKeys.Contains(key))`

Src/VsVimShared/Implementation/Settings/IVsStoreSpecializedAccessorsMap.cs
```diff
@@ -0,0 +1,22 @@
...
+    public interface IVsStoreSpecializedAccessorsMap
+    {
+        bool Initialized<T>();
+
+        void Map<T>(VsSettingsStoreAccessorsProxy<T> accessors);
+
+        VsSettingGetter<T> GetGetter<T>(WritableSettingsStore store);
+
+        VsSettingSetter<T> GetSetter<T>(WritableSettingsStore store);
+    }
```
jaredpar: Why did you take this approach, which seems to have a couple of layers, vs. just wrapping `WritableSettingsStore` in an interface? The design seems to be to allow for betting testing around `WritableSettingsStore` (which is indeed hard to test). But it would seem having a single interface on top of it an a small adapter class would accomplish much the same end here. Is there a reason you preferred this approach?

jaredpar:
Done with review pass (iteration 4). Think we're pretty close now. Really like how `VimApplicationSettings` looks after this change. Had some questions about the abstraction layer benefits.

Src/VsVimShared/Implementation/Settings/IVsStoreSpecializedAccessorsMap.cs
oxilumin: HI @jaredpar . I thought about 2 approaches - using runtime mapping with typeof(), GetType() and Type -> Accessor map and the 2nd one - using this emulated dictionary specialization through static type. This approach is intrinsicly typesafe and faster as it is resolved at jit-compilation time. But this approach requires static generic class. I usually don't like this kind of dependencies as they are harder to test and add complexity so I try to encapsulate it.
The idea is to hide this static class as a private class under a minimalistic facade. This is what VsStoreSpecializedAccessorsMap and StaticMapSpecializationProvider do. They are tricky to test but they are very simple.
The main storage implementations (Inheritors of ISettingsStorage) just take them as a dependency, adding interface here decouples this dependency so the classes are simple and testable.
Now there is an additional part for VS storage, since it has methods like GetBool/String - we need a map between these accessor methods and types (through our specializationMap).
This logic is subject to change as application settings change (when we add new setting types), so I didn't want to put it in VsSettingsStore, which will reduce the need to change this class unless the actual underlying WritableSettingsStore changes in future versions of VS.

This way our main extension point when we add new types is VsStoreStaticSpecializationProvider which doesn't contain any logic beyond accessors mapping.

Single layer of abstraction wouldn't allow to decouple logic from hard static class dependency.

jaredpar: I think I'm missing something here. Why do you feel a `static` class was an issue here? There wasn't one before and I'm not sure how adding a simple caching layer necessitated one being added.

oxilumin: The interface of ISettingsStore now is Get<T> and Set<T>, but underlying WritableSettingsStore from VS SDK has GetBoolean/GetString/GetInt/GetUint/etc.

Generic interface is nicer in terms of code readability so I went with it. To convert generic interface to this set of type-named methods we could either go with chain of if(typeof(T) == typeof(...)) for each type, which doesn't look very nice, a map from Type to delegate or we could leverage the fact that static generic classes have different instances of fields for different generic types and it's resolved at compile-time (so no getting/comparing types at runtime).

The latter requires static class, and I wanted to have a very lightweight wrapper for that static class so the logic related to static class is minimal and the rest could be easily tested/modified. So the cache doesn't depend on static class at compile time.

If we have only 1 layer of indirection (only ISettingsStore) - we'd either have to use typeof(T) -> Func<string, T> map and compare types in runtime or our ISettingsStorage inheritor would incapsulate both settings-related logic and a static class (if we don't want to have runtime type-checks).

Actually if you don't mind using Castle.Core - after adding relative line numbers support - I could simplify all the code to dynamically generated proxy.

jaredpar: Sorry been a busy week. Just getting back to this. 

> Generic interface is nicer in terms of code readability so I went with it

In some cases yes but I'm not convinced in this case. The problem is the underlying store here is not generic. It's fixed to a constrained set of types: `int`, `bool`, `string` and `uint`. Wrapping this with a generic interface implies that the storage can handle any type. That means either you must do one of the following:

1. Throw when a type is specified that isn't represented in the underlying storage. 
1. Implement a serialization layer by which any non-natively supported type is serialized, and deserialized, to one of the supported ones.

This PR takes both approaches here. This means while the interface reads as generic and flexible the implementation is actually quite constrained. There are only a small subset of types available that work here. That's why I think the non-generic one is a bit cleaner here. It may have a few more methods but support declaration is clear.

oxilumin: I gave it a second thought - What do you think about using some config library (like Config.net)?
This way we'll only need to implement a class that reads/writes strings from/to the default VsStore, Wrapper for CommandKeyBinding serializer and a small wrapper for notifications.
Overall amount of code should be much less and we get serialization support for a lot of types for free.

jaredpar: What types do we need to serialize though? Most all of the settings are just the types natively supported and CommandKeyBinding should already round trip to `string` pretty easily.

oxilumin: Serialization is just a benefit, main idea here was to reuse existing tools for settings storage/caching. 
I could prepare a separate PR with config.net as a settings layer (it's a small change, and it'd be easier to compare it when it's visual) and you could decide whether you prefer that one or want to continue with modifying the current one.
If you prefer to continue with the current one - I can change interface to Get/Set Bool/Int/String/etc + generic Get/Set that have IFormatter as an extra argument.

jaredpar: Lets try that that approach of separating it out.

oxilumin: I did some changes so vs settings store is one class now.

jaredpar:
Wanted to give an update here. This PR had stalled out a bit and I wanted to get a release out soon hence I implemented a minimal solution here #2484. Don't think this means we need to abandon this change, just wanted to get the underlying issue fixed before the next release.

ricksladkey:
An alternative version of this excellent idea was merged in PR #2484.

PR #2447, opened 2018-12-26
Add build support for VS2019

oxilumin:


jaredpar:
What version of VS2019 did you get these binaries from? Assuming it was preview1 but wanted to make sure.

PR #2451, opened 2019-01-06
Add support for binary operators other than '+'

oxilumin:


Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -195,34 +195,44 @@ type ExpressionInterpreter
...
+            if leftValue.VariableType = VariableType.List || rightValue.VariableType = VariableType.List then
+                notSupported()
+            else
+                let leftNumber = _getValue.ConvertToNumber leftValue
+                let rightNumber = _getValue.ConvertToNumber rightValue
+                match leftNumber, rightNumber, binaryKind with
+                | Some left, Some right, BinaryKind.Add -> left + right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Subtract -> left - right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Multiply -> left * right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Divide -> left / right |> VariableValue.Number
```
oxilumin: Seems I didn't sync the code from my laptop. Will update div by 0 tomorrow.

Test/VimCoreTest/InterpreterTest.cs
```diff
@@ -1317,13 +1317,95 @@ public void RHSCanBeArbitraryExpression()
...
+                AssertValue("x", 6);
+            }
+
+            [WpfFact]
+            public void RHSCanBeBinaryDivideExpressionAndHandleDivByZero()
+            {
+                Create("");
+                _variableMap["x"] = VariableValue.NewNumber(7);
+                ParseAndRun("let x=24/0");
+                AssertValue("x", 7);
```
jaredpar: Should validate here that the error is issued for divide by zero.

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -213,16 +206,39 @@ type ExpressionInterpreter
...
+            else
+                let leftNumber = _getValue.ConvertToNumber leftValue
+                let rightNumber = _getValue.ConvertToNumber rightValue
+                match leftNumber, rightNumber, binaryKind with
+                | Some left, Some right, BinaryKind.Add -> left + right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Subtract -> left - right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Multiply -> left * right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Divide when right = 0 -> divByZero() 
+                | Some left, Some right, BinaryKind.Divide -> left / right |> VariableValue.Number
+                | Some left, Some right, BinaryKind.Modulo when right = 0 -> divByZero()
```
jaredpar: Should use a different error here as it's mod by zero, not divide by zero.

Test/VimCoreTest/InterpreterTest.cs
oxilumin: Done.
I've found that regardless of what we issue in RunBinaryExpression - the error is overwritten later, so I modified TestableStatusUtil a bit to test this case. I'll probably need to replace that by moq later, but that would require updating a lot of tests so I wanted to separate it into a different PR.

jaredpar:
Thanks for doing this!

Issue #1909, opened 2017-01-31
relativenumbers


Ybalrid:
Hi,

I'm quite a fan of the vim option "relativenumber". Somebody implemented it separately in this extentensioon https://marketplace.visualstudio.com/items?itemName=BrianSchmitt.RelativeNumber&showReviewDialog=true to work with VsVim. 

Problem is, while it's working great, I think this is a functionality that VsVim should have, and that it should react like vim does regarding theses configuration variables : 

number
relativenumber

When you set number, you get normal line numbering. If you set only relative number, you get relative line number from the current cursor line (with this line being line 0) 

With both variable set, you get the absolute line number on the curent line, and relative numbers on other lines. 

Just an idea, it would be a neat feature to support. ^^"

ricksladkey:
@brianschmitt What do you say to idea of making https://github.com/brianschmitt/relativenumber part of VsVim? The idea is apparently wildly popular!

Ybalrid:
YES! I always install both plugin together

ricksladkey:
@oxilumin I'm pretty sure no one is working on this and many users are eager for it to be implemented, so please continue the work to integrate the feature into VsVim. I'm glad you are looking into performance because a poor implementation could be very costly and yet only be noticeable on slow machines or large files.

jaredpar:
I'm definitely not working on it right now either so go for it :smile:

oxilumin:
I've finished the code for the feature itself. It seems that it's ~20-30% faster than built-in absolute line numbers margin, so there should be no performance impact.

I wanted to discuss some settings before making a PR:

1) Do we want to be able to use custom colors/font? Right not it's bound to the formatting defined for "line number" classified text (as in default line numbers margin).

2) Do we want to interact with the native line numbers margin (like hiding it when relative numbers are on)? I think we should not because somebody else might depend on/modify status of native margin or it could change in future releases of VS, so for now - we need to once disable it manually in VS settings.
On the other hand - if we replace the native margin - we get by-language customization for free. So whenever the native margin is disabled for some language - so would relative lines margin be.

jnothstine:
I currently use a relative numbers plugin that is separate from VSVIM.  If VSVIM supports it, I will remove the plugin I have.  That plugin supports showing both the native line numbers and the relative line numbers and I really like having both.  Here is an example of what I see.

[cid:image001.png@01D4A2A2.8C9FC6B0]

Thanks for working on this!

Jerry

[...]

PR #2452, opened 2019-01-07
Add support for relative line numbers

oxilumin:
Closes - #1909

Support for "number" & "relativenumber" vim options.
By default native lines margin is always replaced by the new one (an option to always use the native one could be easily added)
if both number & relativenumber are not set - the margin is disabled altogether.
Margins are configurable per view and default to number & norelativenumber, so need to add code to .vsvimrc to make relative numbers permanent.

Overall this implementation seems to be quite fast, after some tuning it's a bit faster than the native margin. To achieve that results of the most expensive part (text rendering) are cached and whenever we need to update the margin - cached numbers are just rearranged.
Also there is a logic to minimize number of times we need to run the code if layout event is triggered by something that doesn't change line numbers - we ignore it.
Text folding should be supported - folded text is ignored for relative line numbers and taken into account for absolute line numbers.

Some parts still might need some work and it could be beneficial to have a shared cache with LFU logic for line numbers over several thousand (if somebody were to open a huge file and scroll through it - the cache would stay until the buffer is closed). But the native margin doesn't seem to suffer because of this so the performance gain here could be not noticeable.


Src/VimCore/VimSettings.fs
```diff
@@ -725,16 +729,25 @@ type internal EditorToSettingSynchronizer
                 GetVimSettingValue = SettingSyncData.GetSettingValueFunc LocalSettingNames.ExpandTabName true
                 IsLocal = true
             })
-           
+
         _settingList.Add(
             {
-                EditorOptionKey = DefaultTextViewHostOptions.LineNumberMarginId.Name
-                GetEditorValue = SettingSyncData.GetBoolValueFunc DefaultTextViewHostOptions.LineNumberMarginId
+                EditorOptionKey = LineNumbersMarginOptions.NumberOptionName
```
jaredpar: Why change this key here? One feature of VsVim is letting users decide if settings from vsvimrc or Visual Studio should win during startup. Using the standard editor key here is necessary for implementing this feature. With a different key they will be seen as different options.

Src/VimWpf/RelativeLineNumbers/ILineFormatTracker.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Windows.Media;
+
+namespace Vim.UI.Wpf.RelativeLineNumbers
+{
+    public interface ILineFormatTracker
+    {
+        Brush Background { get; }
+
+        System.Windows.Media.TextFormatting.TextLine MakeTextLine(int number);
```
jaredpar: Consider just adding the `using` here and unqualifying the name.

Src/VimWpf/RelativeLineNumbers/LineNumberDrawer.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Linq;
+using System.Windows;
+using System.Windows.Controls;
+
+namespace Vim.UI.Wpf.RelativeLineNumbers
+{
+    public class LineNumberDrawer
+    {
+        private readonly Canvas _canvas;
+
```
jaredpar: Nit: unnecessary blank line

Src/VimWpf/RelativeLineNumbers/LineNumberVisual.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Windows;
+using System.Windows.Media;
+using System.Windows.Media.TextFormatting;
+
+namespace Vim.UI.Wpf.RelativeLineNumbers
+{
+    public class LineNumberVisual : UIElement
+    {
+        private readonly System.Windows.Media.TextFormatting.TextLine _textLine;
```
jaredpar: Consider adding a `using` and unqualifying the name here.

Src/VimWpf/RelativeLineNumbers/LineNumberFormatTracker.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        private TextFormattingRunProperties _formatting;
+
+        private TextFormatter _textFormatter;
+
+        private bool _formatChanged;
+
+        private bool _numbers;
+
+        private bool _relativeNumbers;
```
jaredpar: Nit: don't have a huge preference if fields should be separated by blank lines or not (pattern in VsVim is generally no separation). But prefer within a class it's either but not both 😄

Src/VimWpf/RelativeLineNumbers/ILineFormatTracker.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public interface ILineFormatTracker
+    {
+        Brush Background { get; }
+
+        System.Windows.Media.TextFormatting.TextLine MakeTextLine(int number);
+
+        double NumberWidth { get; }
+
+        bool Numbers { get; }
```
jaredpar: Can you add a comment to what this property does? It's not clear at a glance.

Src/VimWpf/RelativeLineNumbers/Line.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Diagnostics;
+
+namespace Vim.UI.Wpf.RelativeLineNumbers
+{
+    [DebuggerDisplay("{Number} {Baseline}")]
+    public sealed class Line
```
jaredpar: Did you consider making this a `struct` to save on allocations? Looking through I didn't see an obvious place where it needed to be  a `class`. Given this is paired up with lines I'm concerned it could lead to excessive allocations in large files.

Src/VimWpf/RelativeLineNumbers/LineNumberFormatTracker.cs
```diff
@@ -0,0 +1,143 @@
...
+            _formattedLineSource = textView.FormattedLineSource;
+
+            _classificationFormatMap.ClassificationFormatMappingChanged += (s, e) => UpdateFormat();
+
+            _textView.Options.OptionChanged += (s, e) => TryUpdateVimOption(e.OptionId);
+
+            UpdateFormat();
+        }
+
+        private void TryUpdateVimOption(string optionId)
```
jaredpar: This type is essentially keeping a copy of the `numbers` and `relativenumbers` settings and having to do a bit of work to keep them in sync here. I think it would be simpler to just use `IVimGlobalSettings` directly here as the implementation. The values there will already be kept in sync via the `EditorToSettingsSynchronizer` type and it has an event that can be listened to for when the values change.

Src/VimWpf/RelativeLineNumbers/LineNumberDrawer.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (!_formatTracker.TryClearReformatRequest())
+            {
+                return;
+            }
+
+            _canvas.Background = _formatTracker.Background;
+            _canvas.Margin = new Thickness(_formatTracker.NumberWidth, 0, 0, 0);
+
+            _store.Clear();
```
jaredpar: Why is the store cleared on every request here? Are the values just not re-usable between requests?

Src/VimWpf/RelativeLineNumbers/LineNumbersCalculator.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            return allLines.Where(x => x.IsFirstTextViewLineForSnapshotLine);
+        }
+
+        private int GetCaretIndex()
+        {
+            var caretLine = _textView.Caret.ContainingTextViewLine;
+
+            var firstVisibleLine =
+                _textView.TextViewLines.First(x => x.IsFirstTextViewLineForSnapshotLine);
```
jaredpar: The `TextViewLines` property can throw if we're in the middle of rendering already. Are we sufficiently guarded from that happening here?

Src/VimWpf/RelativeLineNumbers/LineNumbersCalculator.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (GetVisualLineNumber(caretLine, out int caretVisualLineNumber) &&
+                GetVisualLineNumber(firstVisibleLine, out int referenceVisualLineNumber))
+            {
+                return caretVisualLineNumber - referenceVisualLineNumber;
+            }
+
+            return -1;
+        }
+
+        private bool GetVisualLineNumber(ITextViewLine line, out int visualLineNumber)
```
jaredpar: Consider naming `TryGet...` as this is a try pattern method.

Src/VimWpf/RelativeLineNumbers/RelativeLineNumbersMargin.cs
```diff
@@ -0,0 +1,66 @@
...
+            _linesTracker.LineNumbersChanged += async (x, y) => await RedrawLinesAsync().ConfigureAwait(true);
+            _formatTracker.VimNumbersFormatChanged += async (x, y) => await RedrawLinesAsync().ConfigureAwait(true);
+        }
+
+        private async Task RedrawLinesAsync()
+        {
+            await _refreshLock.ExecuteInLockAsync(RedrawLinesInternalAsync)
+                .ConfigureAwait(true);
+
+            async Task RedrawLinesInternalAsync()
```
jaredpar: Why does this need to be async here? Feel like this is adding a bit of complication + some possible race conditions.

Src/VsVimShared/HostFactory.cs
```diff
@@ -114,6 +118,50 @@ private void BeginSettingSynchronization(IVimBuffer vimBuffer)
...
 
+        private void EnableVimLineNumbersMargin(IEditorOptions textViewOptions)
+        {
+            textViewOptions.SetOptionValue(DefaultTextViewHostOptions.LineNumberMarginId, false);
+            textViewOptions.SetOptionValue(LineNumbersMarginOptions.LineNumbersMarginOptionId, true);
+
+            textViewOptions.OptionChanged += OnLineNumberOptionChangedSync;
+        }
+
+        private void DisableVimLineNumbersMargin(IEditorOptions textViewOptions)
```
jaredpar: This method is never called

Src/VsVimShared/HostFactory.cs
```diff
@@ -114,6 +118,50 @@ private void BeginSettingSynchronization(IVimBuffer vimBuffer)
             _editorToSettingSynchronizer.StartSynchronizing(vimBuffer, settingSyncSource);
         }
 
+        private void EnableVimLineNumbersMargin(IEditorOptions textViewOptions)
+        {
+            textViewOptions.SetOptionValue(DefaultTextViewHostOptions.LineNumberMarginId, false);
+            textViewOptions.SetOptionValue(LineNumbersMarginOptions.LineNumbersMarginOptionId, true);
```
jaredpar: Why are these options being hard coded to `false` here?

Src/VimWpf/RelativeLineNumbers/LineNumberDrawer.cs
oxilumin: Reformat request means format of the line number text changes (font, size, color, background), so we can not reuse cached rendered line numbers anymore. On the other hand we don't want to invalidate them all at once as it could be a long task and potentially could cause some visible performance drop during reformatting on slower PCs.

Alternative would be to set "dirty" flag for each object and invalidate it as we draw it for the first time after the update. It would make the code a bit more complicated but I don't see any performance benefit from it as we save only on one allocation which is negligible compared to generating new bitmap for a number with the new formatting style. If you wish I could do that though.

Src/VimWpf/RelativeLineNumbers/LineNumberFormatTracker.cs
oxilumin: Ok. I agree, every project has their own coding styles, so I'll put this into my .editorconfig for VsVim.
I'm just used from my last job to group fields into logical scopes.

Src/VimWpf/RelativeLineNumbers/ILineFormatTracker.cs
oxilumin: Sure. Actually it should be "Number" and "RelativeNumber" or "VimNumber" and "VimRelativeNumber" as they are properties for Vim options.

Src/VimWpf/RelativeLineNumbers/LineNumbersCalculator.cs
oxilumin: We should be safe here as this is called in the UI thread and LineNumbersTracker encapsulates all the logic for deciding when we should trigger actions on "NumbersChange". It also ensures that it only fires the event if "InLayout" is false for the TextView (as well as "IsClosed").

Src/VimWpf/RelativeLineNumbers/RelativeLineNumbersMargin.cs
oxilumin: I've checked for race conditions when I was writing it and it should be Ok. Initially it all was not async and was running in UI thread, but when I tried to run it in a low resources VM requests were piling up and it added to the overall delay in fast scroll scenarios. With async here we can squash all pending requests into one and make it a bit better on old PCs. It's not game-changing optimization here, but I thought that the logic of this SafeRefreshLock can be used for other operations in future to improve overall performance.

Src/VsVimShared/HostFactory.cs
oxilumin: I wanted to bind it to a new VimApplication setting once we merge settings cache, so users have option to opt out of this relative line numbers margin and fallback to old behavior with native line numbers margin.

Src/VimCore/VimSettings.fs
oxilumin: We can't use default value here as It'd enable the native VS line numbers margin whenever "Number" is set to true. Even if both "Number" and "RelativeNumber" are set to true (the case which is not supported by the native margin). So we need to do something different here.

I gave it a second thought and I think that we could add do 2 mappings here.
If VS wins and native margin is shown - we set "number" (so the user has at least some numbers margin) and let vim decide what to do with relativenumber. If VS wins and native margin is hidden - we set both "number" and "relativenumber" to false. 

If Vim wins - we use existing logic to hide/show margins.

Also we could show native margin whenever number is true and relativenumber is false (instead of using our custom one). This would eliminate the need of "Enable"/"Disable" settings for the new margin and also would let users have the "old" behavior (with native margin) if they don't set relativenumber to true. I'll try to see if there are any problems with this approach and implement it.

Src/VimWpf/RelativeLineNumbers/Line.cs
oxilumin: Done

Src/VimWpf/RelativeLineNumbers/LineNumberFormatTracker.cs
oxilumin: Done. Don't we actually want to use LocalSettings here? So we can have different options for different file extensions/buffers?

Src/VimWpf/RelativeLineNumbers/Util/SafeRefreshLock.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Vim.UI.Wpf.RelativeLineNumbers.Util
+{
+    public class SafeRefreshLock
```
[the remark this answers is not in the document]
oxilumin: Hi,
I did an extra perf analysis and it seems that even without refresh lock we are faster than the native VS margin, I might have a bias against doing any extra work synchronously in the UI thread  that either is not needed anymore or could be done in a background thread, so I tried to decouple this part.
If you don't mind some performance drop for fast scrolling scenarios on slower PC - I'm removing it on these weekends (I believe that would be the solution you lean towards) and we could push a synchronous version. My concern here was that VsVim macro execution and some actions with counts could lead to automatic scroll that is visually not fast enough (but that might be another workitem for optimizing those particular scenarios themselves).
If you prefer to  keep it - I totally agree with you that custom sync primitives have inherent complexity and I had to address it by providing comments for invariants analysis, state diagram and a good description (And thanks for noting it, I'll stick to this practice in future).
So if you decide to keep it - I'll add it into the code.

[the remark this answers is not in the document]
jaredpar: Lets' try the sync version for now. We can go back to the async if customers start reporting issues with it.

ricksladkey:
@oxilumin Are you up to finishing the last few requested changes? This PR seems very close to being merged and is widely anticipated by VsVim users.

Src/VimWpf/RelativeLineNumbers/LineNumbersCalculator.cs
ricksladkey: As it turns out, `ContainingTextViewLine` can throw as well. The new PR has a fix for this.

ricksladkey:
Closing this PR due to inactivity. It was reopened as a new PR #2571.

Issue #2463, opened 2019-02-17
VS 2019 Bugs


jaredpar:
Been working on making it so I can develop the VsVim code base using Visual Studio 2019 Preview 3. Had to insert some work arounds to the code base for bugs I found. This is tracking the behavior and removing the work around once the VS bugs are fixed. 

1. The core undo / redo primitives can only function when the `ITextUndoHistory.Properties` value has a key / value pair of `typeof(ITextView), ITextView`. Lacking that any undo / redo which involves moving the caret will cease to function. In 2017 and before the `ITextView` was put into the property bag via `AfterTextBufferChangeUndoPrimitive.Create`. Staring with 2019 that is no longer done. Yet the implementation of Undo / Redo still depend on it being there. Temporarily forcing it to be there. 
1. MEF composition fails do to following values on `Microsoft.VisualStudio.Language.CodeCleanUp.CodeCleanUpFixerRegistrationService` being `readonly`: `ProfileService` and `mefRegisteredCodeCleanupProviders`. 
1. MEF composition fails do to `Microsoft.VisualStudio.UI.Text.Wpf.FileHealthIndicator.Implementation.FileHealthIndicatorButton` constructor having a `null` value for `style`.
1. `ITextCaret.MoveTo` is not respecting the argument when on EOL in certain cases.

jaredpar:
Update on the status of the issues here:

1. This is "By Design". The editor has this fundamental constraint that the property bag must have the `ITextView` available for undo / redo to work. The behavior change is when they inserted it. The logic is no during a focus change which doesn't reliably occur during unit testing. Editor team confirmed it's correct to just add this if missing when undo / redo occurs. 
2. Logged and fixed
3. Logged and fixed
4. Looks like we've worked around any remaining issues here.

PR #2464, opened 2019-02-17
Fix developing VsVim with VS2019

jaredpar:
This change is primarily to enable developing the VsVim code base using Visual Studio 2019. This is in addition to the 2017 support. Fixing up this brought along a number of other infrastructure changes that make this a bit cleaner. 

Note: this change is all about enabling the development of the code base. This means removing 2015 support is all removing the ability to develop the code base using VS2015 as an editor. The product VsVim is still fully supported on VS2015. 

Summary of the changes here: 
1. Updated the VS2019 references to be the preview 3 binaries.
1. Changed the 2019 specific project to reference the correct set of binaries. Previously it was referencing a mix of 2019 and 2017 binaries. 
1. Add a project and unit tests to ensure each project is referencing the correct version of VS SDK. There were a few cases missed in the initial 2019 support PR. That's understandable because the rules are complex. Asking everyone to remember them all is not a path for success. Instead I did my best to codify them as a test that will self enforce this.
1. Changed the `#if` macros for the development environment to be `VSVIM_DEV`. This will hopefully make it clearer they're about development not the product build. 

This also found a few 2019 bugs that had to be worked around. This work is tracked in #2463


Src/VimEditorHost/Implementation/Misc/BasicExperimentationServiceInternal.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Primitives;
+using Microsoft.VisualStudio.Language.Intellisense.Utilities;
+using System.Threading;
+using Microsoft.VisualStudio.Text.Utilities;
+
+namespace Vim.EditorHost.Implementation.Misc
+{
+    [Export(typeof(IExperimentationServiceInternal))]
+    internal sealed class BasicExperimentationServiceInternal : IExperimentationServiceInternal
```
ricksladkey: It's not clear to me why this is needed. Maybe add a comment to a link. Here is the main thing that pops when googling that interface:

* https://github.com/dotnet/roslyn/issues/27428

jaredpar: I'll update the comment here. Unfortunately it's not going to be much more informative than the data in the issue you linked to. We're implementing it here because the editor expects it to be implemented. There is no real documentation on what it does. It's unfortunate they didn't use an optional import as the implementation isn't really needed here.

Src/VsSpecific/VsSpecific/TestUtil.cs
```diff
@@ -0,0 +1,32 @@
...
+#if DEBUG
+    /// <summary>
+    /// The version specific DLLs will change whether or not the actually depend on editor components as 
+    /// our minimum version changes and features become available in the core code base. This means tests 
+    /// written to verify version consistency can get out sync as references in project files won't
+    /// be written into the compiled assemblies unless the types are used.
+    /// 
+    /// This class is present to make sure we always reference the editor binaries even as our minimum
+    /// supported version changes.
+    /// </summary>
```
ricksladkey: Great explanation!

Test/VsVimTest/ReferenceTest.cs
```diff
@@ -0,0 +1,305 @@
...
+                        return (VsVersion)value;
+                    }
+
+                    throw new InvalidOperationException($"Invalid enum value {value}");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Make sure the correct VS SDK binaries are referenced in the core binaries.
```
ricksladkey: Great that this is automated.

jaredpar: Agree. The rules here are individually easy enough but there are so many to remember and it can be difficult to trace them through the build process. Automated so hopefully I can forget about this going forward 😄

Issue #2465, opened 2019-02-18
Make incremental search asynchronous


jaredpar:
The incremental search feature today is completely synchronous. As the developer types out a search string for every character the search will be initiated, completed and the results updated on the screen before the next key stroke is processed. 

That is potentially wasteful when the developer is typing quickly or typing out a fairly long search phrase. Only after the final character is typed, or when the developer pauses to look at the screen, are the results of the search relevant. Further when the user cancels the search, say with <kbd>Escape</kbd>, there is no actual cancel going on. The search fully completes then the incremental session is cancelled. 

In the 99% case this doesn't really matter. The code paths here are fairly speedy and hence his synchronous behavior typically goes unnoticed. 

Lately though I've been working in some [incredibly large files](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Test/Semantic/Semantics/NullableReferenceTypesTests.cs) and this behavior is **painfully obvious** there. The IDE can go unresponsive for several seconds when searching in such an environment to the point you type a few characters and then slowly watch the IDE catch up with the search. 

To combat this I'm going to change the incremental search to be asynchronous up until the point a developer hits <kbd>Enter</kbd>. That should reduce work here and make working in extremely large files a more enjoyable experience.

PR #2471, opened 2019-02-22
Enable Dispatcher in our tests

jaredpar:
This changes the underpinnings of how our WPF based tests execute. This re-uses more of the core WPF primitives, like `Dispatcher`, instead of the hand constructed ones. This means operations like `async`, `Dispatcher`, etc ... should work much more naturally in the unit tests. Specifically it opens up the door for having `async Task` methods. Those did work today but had a number of issues where they could lead to deadlocks. 

A good chunk of this is just re-using [test primives](https://github.com/dotnet/roslyn/tree/master/src/EditorFeatures/TestUtilities/Threading) from [roslyn](https://github.com/dotnet/roslyn). Our VSIX test infrastructure has evolved together (both repos merging back and forth). This is just another iteration of that.

The need for this came about from working on #2465. The testing of incremental search can now get asynchronous in a few cases and the most natural way to do this is have `async Task` methods instead of `void`. Chasing that down revealed the need for this. I decided to merge this separately from the incremental search work to keep the PRs smaller.


jaredpar:
Looks like we're hitting the `WeakEventTable` shutdown bug. Need to implement a similar fix as here: 

https://github.com/dotnet/roslyn/pull/27843

Src/VimTestUtils/Utilities/StaContext.cs
```diff
@@ -0,0 +1,83 @@
...
+                    // xUnit creates its own synchronization context and wraps any existing context so that messages are
+                    // still pumped as necessary. So we are safe setting it here, where we are not safe setting it in test.
+                    SynchronizationContext.SetSynchronizationContext(syncContext);
+
+                    staThreadStartedEvent.Set();
+
+                    Dispatcher.Run();
+                }));
+
+                StaThread.Name = $"Sta Thread";
```
ricksladkey: This looks like it might be a mistake, but I'm guessing the intent is "flexibility".

Test/VimCoreTest/CommonOperationsIntegrationTest.cs
```diff
@@ -44,14 +44,14 @@ protected ScrollOffsetTest()
 
             private void AssertFirstLine(int lineNumber)
             {
-                TestableSynchronizationContext.RunAll();
+                Dispatcher.DoEvents();
```
ricksladkey: These all had to change, and they might have to change again. Conceptually, we are performing some kind of pumping applicable to the test infrastructure. Exposing a single class method to do that (which would currently call `Dispatcher.DoEvents()` seems like it might prevent the need for any future changes to tests doing this kind of testing.

Src/VimTestUtils/Utilities/StaContext.cs
jaredpar: Mostly the intent here is to give the thread a name that is easy to spot in the Debugger. A way to differentiate between all of the other STA threads in the process.

Test/VimCoreTest/CommonOperationsIntegrationTest.cs
jaredpar: That's a good point. Let me try that.

PR #2472, opened 2019-02-26
Make incremental search asynchronous

### LINKED ISSUE #2465: Make incremental search asynchronous

jaredpar:
This change makes incremental search asynchronous up until the point the developer hits the <kbd>Enter</kbd> key. Key strokes typed up until this point will spawn search requests on the thread pool and update the incremental search state when they complete. In the case another key stroke is typed before the search completes then a new search will be started and the existing one ignored. 

Ideally the existing searches would be cancelled when a new key press occurs. But the editor APIs don't allow for cancellation in their `ITextSearchService` APIs. Hence there is nothing to really do but let the search run its course. That doesn't affect the UI responsiveness though. 

The bulk of the work here is getting the unit tests updated. Many tests just work because they complete a search with <kbd>Enter</kbd> hence don't observe the asynchronous nature here. Many others though verify the state of vim in the middle of a search. Those necessarily have to move to `async` in order to correctly validate the state.

Dogfooding this I see a significant improvement when searching in extremely large files. Before I could watch my key strokes be typed out as the search progressed on. Now I can type and just wait for the las search to take hold. 

closes #2465

### DIFF
--- a/Src/VimCore/IncrementalSearch.fs
+++ b/Src/VimCore/IncrementalSearch.fs
@@ -14,7 +14,6 @@ open Vim.VimCoreExtensions
 /// cancelled.
 type internal IncrementalSearchSession
     (
-        _key: obj,
         _vimBuffer: IVimBuffer,
         _operations: ICommonOperations,
         _searchPath: SearchPath,
@@ -23,7 +22,9 @@ type internal IncrementalSearchSession
 
     let mutable _searchData = SearchData("", _searchPath, _isWrap)
     let mutable _searchResult : SearchResult option = None
+    let mutable _historySession : IHistorySession<ITrackingPoint, SearchResult> option = None
     let mutable _isActive = true
+    let _key = obj()
     let _globalSettings = _vimBuffer.GlobalSettings
     let _searchStart = StandardEvent<SearchDataEventArgs>()
     let _searchEnd = StandardEvent<SearchResultEventArgs>()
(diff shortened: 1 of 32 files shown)


Src/VimCore/IncrementalSearch.fs
```diff
@@ -5,231 +5,344 @@ open Microsoft.VisualStudio.Text.Operations
 open Microsoft.VisualStudio.Text.Editor
 open Microsoft.VisualStudio.Text.Outlining
```
jaredpar: For this file it may be easier to just look at the new code and ignore the diff. Practically none of the original implementation remains.

PR #2473, opened 2019-02-28
Keep selection for all Edit commands

### LINKED ISSUE #2470: map visual studio command ExpandSelectiontoContainingBlock
The following mapping doesn't work.

```
nmap <Leader>vv :vsc Edit.ExpandSelectiontoContainingBlock<CR>
```
Executing command `:vsc Edit.ExpandSelectiontoContainingBlock` in command line doesn't work either.

The VS version is 15.9.7

jaredpar:
This changes the keep selection logic to include all "Edit." commands.

closes #2470

### DIFF
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -1149,11 +1149,10 @@ namespace Vim.VisualStudio
                 return true;
             }
 
-            var comparer = StringComparer.OrdinalIgnoreCase;
-            if (comparer.Equals(command, "Edit.SurroundWith"))
+            if (command.StartsWith("Edit.", StringComparison.OrdinalIgnoreCase))
             {
-                // Need to keep the selection here so the surround with command knows the selection
-                // to surround.
+                // The Edit commands often change the selection as a part of their implementation. Need
+                // to just keep their results to match user expectations.
                 return true;
             }
 


ricksladkey:
I agree that this change will fix the identified issue, but I am concerned that the number of commands being added is quite large and I don't know if any of them will encounter problems as a result of keeping the selection. I know there is some reason we don't keep the selection by default, but I can't think of a specific example where we don't (and shouldn't) so that I could try to gauge whether any of the newly added ones are now going to misbehave.

jaredpar:
@ricksladkey 

> I know there is some reason we don't keep the selection by default, but I can't think of a specific example where we don't (and shouldn't) so that I could try to gauge whether any of the newly added ones are now going to misbehave.

The one that comes to mind for me is Go To Definition. That command will in some C++ circumstances end up selecting the target. Hence a `gd` command ends up moving to visual mode. 

>  I am concerned that the number of commands being added is quite large

That's true. When I looked at my setup I didn't see many and found a few others that should be excluded. Hence I just added them all. But your comment reminds me that the set of commands isn't fixed. I will back it out to just the affected command for now.

ricksladkey:
@jaredpar I took some time to look into the whole "should keep selection after host command" area. It turns out that `<C-]>` or `gd` are not "host commands" in the sense that they are not executed with `:vsc [command]`. The VsVim infrastructure for "go to definition" is unrelated to this logic and clears the selection independently of it.

After thinking about it, I see no reason we should **ever** clear the selection after a host command. The only way a host command can ever be executed is if a users issues `:vsc [command]`, almost certainly through a mapping. If the user desires that the selection be cleared, they could "say so".

My recommendation is to add a "bang interpretation" of `:vsc` where `:vsc! [command]` means to run the command and clear the selection. Then the user has complete control of the selection clearing and we can remove all the logic from VsVim trying to handle a multitude of commands on a case-by-case basis.

Does this sound like a good solution to the "should keep selection after host command" problem? If so, I will prepare a merge request.

jaredpar:
@ricksladkey I really like that take on the problem here. I think it cleans up this area nicely.

PR #2484, opened 2019-04-07
Add simple setting cache layer

jaredpar:
This is an alternative to #2443. The approach is simpler, it's just a simple cache for the limited amount of setting types that we use in VsVim. Mostly did this because want to get a new release out soon and this was the simplest work around that got past the perf issue.


Src/VsVimShared/Implementation/Settings/VimCollectionSettingsStore.cs
```diff
@@ -0,0 +1,131 @@
...
+                catch (Exception e)
+                {
+                    Report("Unable to create the settings collection", e);
+                }
+            }
+        }
+
+        private void Report(string message, Exception e)
+        {
+            message = message + ": " + e.Message;
```
ricksladkey: I think an interpolated string is both more efficient and more readable, so it's a win/win.

jaredpar: Good point. Let me fix that.

jaredpar:
@ricksladkey 

> A question that comes to mind: is it theoretically possible to get a behavioral change in the API by using two copies of Visual Studio simultaneously? 

Potentially. In the case that both were directly writing to the shared file on every setting change then it would be possible to observe the change from one VS instance to the other. Concretely enabling logging in one would in theory automatically enable it in the other in this scenario.

> If so, is it an practical problem too?

Don't believe so. VS already has some un-intuitive behavior when you have multiple copies open and change settings in both of them. The behavior, IIRC, is that the last one to shutdown successfully wins the setting battle. There is no attempt to merge and no warning.

This change though actually makes VsVim a bit more consistent. There are a number of features that depend on being able to get a settings changed event. In the case off two VS instances that event never happens for the other instance of VS. Now though each VS instance is isolated a bit and hence they should always get events when settings observably change.

PR #2491, opened 2019-04-16
Environment variables and string equality in expressions

### LINKED ISSUE #2458: Echoing environment variables
I could swear that in previous versions of VsVim I was able to echo environment variables like so:
:echo $HOME
:echo $VIM

I believe I have the most current release installed (2.6.0.0), at least according to the "Extensions and Updates" window in Visual Studio.

I've also tried this:
:echo %HOME%
:echo %VIM%

Both syntaxes return "Parse error".

Please forgive me if I'm missing something obvious. I did try to peruse Kazark's old implEcho branch, and that just brought up more questions in my mind.

### LINKED ISSUE #2469: Detecting 2017 vs 2019 in vsvimrc
Annoylingly, all my mappings to resharper functionality don't work in 2019 as there is a "Extensions" prefix,  but 2019 and 2017 are loading the same file, can I conditionally distingush which version I'm dealing with / have different files load depending on version?

ricksladkey:
#### Issues

- Closes #2469
- Closes #2458

### DIFF
--- a/Test/VimCoreTest/InterpreterTest.cs
+++ b/Test/VimCoreTest/InterpreterTest.cs
@@ -1322,9 +1322,16 @@ namespace Vim.UnitTest
                 Create("");
                 var variable = "magic";
                 var value = "xyzzy";
-                Environment.SetEnvironmentVariable(variable, value);
-                ParseAndRun($"let x = ${variable}");
-                AssertValue("x", value);
+                try
+                {
+                    Environment.SetEnvironmentVariable(variable, value);
+                    ParseAndRun($"let x = ${variable}");
+                    AssertValue("x", value);
+                }
+                finally
+                {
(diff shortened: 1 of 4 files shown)


Test/VimCoreTest/InterpreterTest.cs
```diff
@@ -1316,6 +1316,17 @@ public void RHSCanBeArbitraryExpression()
                 AssertValue("x", "Hello, world!");
             }
 
+            [WpfFact]
+            public void RHSCanBeEnvironmentVariable()
+            {
+                Create("");
+                var variable = "magic";
+                var value = "xyzzy";
+                Environment.SetEnvironmentVariable(variable, value);
```
jaredpar: Should unset this at the end of the test to make sure another test doesn't accidentally depend on it.

PR #2494, opened 2019-04-20
Fix non-leading tab expansion

### LINKED ISSUE #2493: The 'retab' command mangles non-leading tabs
I came across a code base that had mixed tabs and spaces and I was trying to clean it up. I first tried to use Visual Studio Edit -> Advanced -> Untabify Selected Lines, but this turned out only to untabify leading tabs.

I thought to myself, surely vim has a builtin solution, and I shortly learned about `:retab`. While it does have the advantage of untabifying non-leading tabs, it unfortunately does not expand them correctly.

Here is a simple example. First create a text buffer using the settings `:set noexpandtab tabstop=4`:

```
cat<Tab>dog
```
which will be rendered as:

```
cat dog
```
Now set `:set expandtab` and run `:retab` and the result should not visually change, but the buffer now looks like this:

```
cat    dog
```

In other words, one tab (that started on column 3 and took up one screen column) was expanded to four spaces, as though the tab were aligned to a multiple of four.

ricksladkey:
#### Issues

- Closes #2493


Src/VimCore/CommonOperations.fs
```diff
@@ -247,10 +247,23 @@ type internal CommonOperations
...
+        // lines, but shouldn't insert tabs when filling in non-leading virtual
+        // space because vim never inserts tabs in that situation.
         if x.CaretVirtualPoint.IsInVirtualSpace then
             let blanks: string = 
                 let blanks = StringUtil.RepeatChar x.CaretVirtualColumn.VirtualSpaces ' '
-                x.NormalizeBlanks blanks
+                if x.CaretPoint = x.CaretLine.Start then
+
+                    // The line is completely empty so use tabs if appropriate.
+                    let spacesToColumn = x.GetSpacesToPoint x.CaretPoint
```
jaredpar: Is there any case where columns vs. points would return a different answer here? Given we're in virtual space pretty sure this is no because there can't be any wide code points.

ricksladkey: This computation would work correctly in the presence of wide code points, but given the previous conditions, the value of `spacesToColumn` will always be zero so the calculation is a formality.

PR #2495, opened 2019-04-20
Add VimBuffer to CsharpScriptGlobals

TKumadaki:
The following code does not work as I expected.

**.vsvimrc**
`:autocmd    BufWinEnter    *.*    :csx Test<Enter>`

**Test.csx**
```
using System.Windows;
using Vim;
using Vim.Extensions;

var activeVimBuffer = Vim.ActiveBuffer;
if (activeVimBuffer.IsNone())
{
    MessageBox.Show("Fail!!");
    return;
}
IVimBuffer vimBuffer = activeVimBuffer.Value;
MessageBox.Show("Success!!");
```
`Vim.ActiveBuffer` has the following comments.

**Buffer actively processing input.
This has no relation to the IVimBuffer which has focus.** 

`VimBuffer` is activated when `KeyInputStart` occurs.
It is not yet active when the `BufWinEnter` event occurs.

`VimBuffer that executed csx command` And `ActiveBuffer`, `FocusedBuffer`.
These need to be distinguished.

so I decided to add `VimBuffer` which executed csx command to `CSharpScriptGlobals`.

With this change, `CSharpScriptGlobals.Vim` is no longer needed.
But `Vim` is a root object, so I leave it.
I delete `CSharpScriptGlobals.Vim` if the reviewer wants.


Src/VsSpecific/VsSpecific/CSharpScriptGlobals.cs
```diff
@@ -11,14 +11,16 @@ public class CSharpScriptGlobals
         public LineRangeSpecifier LineRange { get; }
         public bool IsScriptLocal { get; } = false;
         public IVim Vim { get; } = null;
+        public IVimBuffer VimBuffer { get; } = null;
 
-        public CSharpScriptGlobals(CallInfo callInfo, IVim vim)
+        public CSharpScriptGlobals(CallInfo callInfo, IVim vim, IVimBuffer vimBuffer)
```
ricksladkey: Functions that take an `IVimBuffer` argument don't also need an `IVim` argument because every `IVimBuffer` contains a reference to `IVim`.

ricksladkey:
VsVim ensures that all user operations, even loading the vimrc file, occur in the context of some buffer. As a result, I think this change is both necessary and appropriate.

Src/VsSpecific/VsSpecific/CSharpScriptGlobals.cs
TKumadaki: Thank you for the review.
I deleted unnecessary arguments.

jaredpar:
Can you rename the "C# scripting.md" file to be "CSharp scripting.md"? The # in the file name breaks some tooling (unfortunately).

TKumadaki:
I changed the file name.

PR #2505, opened 2019-04-26
Work around peek window command margin input problems

### LINKED ISSUE #2492: Backspace issue in peek definition window
In a peek definition window, backspace does not immediately update the displayed command buffer in normal mode.

Example:

1. Open peek definition window
2. Typo `:wqa`
3. Backspace once or twice
4. The command buffer still displays `:wqa`

The displayed command only changes after focus leaves the peek definition window, or after the entire command is removed.

Using Visual Studio 2015 and the latest VsVim (2.6.0).

ricksladkey:
#### Issues

- Close #2492

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
+++ b/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Windows;
+using System.Windows.Input;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Text;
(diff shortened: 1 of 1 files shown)


Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
```diff
@@ -181,6 +194,55 @@ private bool TryProcessWithBuffer(KeyInput keyInput)
...
+        {
+            var keyMap = new Dictionary<VimKey, Key>
+            {
+                { VimKey.Back, Key.Back },
+                { VimKey.Left, Key.Left },
+                { VimKey.Right, Key.Right },
+                { VimKey.Home, Key.Home },
+                { VimKey.End, Key.End },
+            };
+            foreach (var pair in keyMap)
```
jaredpar: Rather than iterate here why not make the key of the dictionary be `KeyInput`? That would allow for a `TryGetValue` call.

ricksladkey: Yes, that makes sense. Will do.

ricksladkey: Fixed the dictionary key, made the dictionary static and added a couple more keys that were not being handled.

Issue #1793, opened 2016-06-22
typing certain characters in Normal mode causes them to be inserted


jankner:
Some characters get inserted in the document when you type them in Normal mode. These are the characters I've discovered: ¤ ¨ £ € §  ´

jaredpar:
Interesting.  Can you tell me how you type these?  In particular what keyboard you are using?

mwinterb:
Oh boy, a tiny part of my childhood comes in handy! ¤ is typed with Alt+0164 (numpad for all of those, holding down Alt the entire time, then release). Character Map includes the keystrokes for several others, a random sampling hasn't gotten a single non-insert, except for the "letter" characters. I don't know enough about keyboards and input to say if the Alt+ mechanism is reported the same to an application as with actually typing those on a real keyboard, though, so this info may not be 100% helpful.

jankner:
I used a Swedish keyboard to type those characters, but the Alt-number method has the same result. For example ¤ is typed as Shift-4, and § is typed using the key left of the 1 key.

jkadg:
On a german keyboard one can insert € bei typing AltGr-E. VsVim inserts €, but doesn't change to insert mode at all.

jaredpar:
Thanks.  I will try that out and see if I can repro the problem.

ricksladkey:
I was able to reproduce this to enter into the buffer in insert or normal mode with a US keyboard using the Alt+0164 technique described by @mwinterb. I'll take a look into the issue.

PR #2516, opened 2019-05-02
Fix ASCII control character processing on international keyboards

### LINKED ISSUE #2462: inconsistent Ctrl-ú (Ctrl-[) handling in different modes (Czech keyboard)
After I submitted #2461 I found out that `Ctrl-ú` (which is `Ctrl-[` equivalent on Czech keyboard) does not work in the visual selection mode. For example the command `Vjjj<Ctrl-ú>` selects four lines and remains in the visual selection mode (the terminating `<Ctrl-ú>` has no effect).

When I switch to "English US" keyboard, the corresponding `Vjjj<Ctrl-[>` selects four lines and cancels the selection.

What is strange that `<Ctrl-ú>` (when using Czech layout) works ok in the insert mode, i.e. it terminates the insert mode and returns back to the normal mode. So VsVim somehow recognizes this key correctly, but not in visual selection mode.

ricksladkey:
#### Issues

- Closes #2462

#### Discussion

Unfortunately, this had to change the model for how `IMode.CanProcess` works and when it returns true. The following discussion is very arcane, but I'll try to describe it in detail for future reference.

As background, if VsVim declines to handle a keyboard input at all of the various input points, the input falls back to Visual Studio which would then  assume the user was typing and insert the keyboard input into the buffer. There was even an issue (issue #1793) where non-ASCII characters got inserted into the buffer in normal mode instead of beeping.

As a result, many modes in VsVim typically report that they can process many, many more inputs than are actually valid inputs for that mode. This is because they desire to process them as an error rather than let Visual Studio handle the input erroneously.

This whole approach is predicated on the premise that a VsVim mode gets one chance to process they input. But the reality is that a mode may get **multiple** chances to process the raw keystrokes typed by the user. If we decline to handle a key combination during the `KeyDown` event, that key combination **may or may not** have an alternate interpretation that appears later during the `TextInput` event. Frankly, when we are in the `KeyDown` event, we have no idea whether a later `TextInput` event will be generated or what key input it will generate. There are control keys, dead keys, magic unicode keystroke sequences, etc. that are entirely outside of our control.

The ASCII control character `0x27 / ESC` on the Czech keyboard is an example of this kind of key. During the `KeyDown` event, we encounter a composite key that does not map to any ASCII character: `Ctrl+ú`. This is a perfectly valid key input from vim's point of view. And from VsVim's for that matter. But, if we decline to handle the `KeyDown` event, we will later receive a `TextInput` event for an entirely different character, this time an ASCII one: `0x27 / ESC`. But on the US keyboard, the same thing will happen. On `KeyDown`, VsVim will see the `Ctrl+[` composite key input, and if declined, will result in `0x27 / ESC` during the `TextInput` event.

The US keyboard doesn't have any problem here because both of the key inputs are the same from vim's point of view. But on the Czech keyboard, the only way we will ever find out that the keyboard produces ASCII `0x27 / ESC` by pressing the control key down at the same time as the `ú` key, is by declining to process it during the `KeyDown` event.

The best way for VsVim to handle this is by only handling `<C-ú>` as a composite key input when the user has a mapping for that key. If there is no such mapping, and they couldn't be trying to enter text into the buffer, then we can decline to handle the composite key input because **processing it would result in an error**. More specifically, it would beep.

This means that a mode may get several opportunities to see the same keystrokes, each converted into different key inputs. Accepting one too early prevents us from see the later ones. But if we decline to handle the first one, and **there is no second one**, Visual Studio could end up inserting characters into the buffer in a non-input mode.

As a result, we now have a new `CanProcess` model:

- A mode should return `true` for `CanProcess` only if it is either a valid key or we are certain it cannot be interpreted as a different valid key
- In non-insert modes we must handle undesired insertable text **at the very last minute**, i.e. just before we would pass it on to Visual Studio

In the current Visual Studio / VsVim architecture, the point of no return for key processing is during `StandardCommandTarget.Exec`. If we return false, Visual Studio will process the key input itself. if we don't want it to do that, that's where we need to decide if we return true.

### DIFF
--- a/Src/VimCore/FSharpUtil.fs
+++ b/Src/VimCore/FSharpUtil.fs
@@ -559,7 +559,7 @@ module CharUtil =
     let IsWhiteSpace x = System.Char.IsWhiteSpace(x)
     let IsNotWhiteSpace x = not (System.Char.IsWhiteSpace(x))
     let IsControl x = System.Char.IsControl x
-    let IsAscii x = x >= 0 && x <= 127
+    let IsAscii x = x >= '\u0000' && x <= '\u00ff'
 
     /// Is this the Vim definition of a blank character.  That is it a space
     /// or tab
--- a/Src/VimCore/FSharpUtil.fs
+++ b/Src/VimCore/FSharpUtil.fs
@@ -559,6 +559,7 @@ module CharUtil =
     let IsWhiteSpace x = System.Char.IsWhiteSpace(x)
     let IsNotWhiteSpace x = not (System.Char.IsWhiteSpace(x))
     let IsControl x = System.Char.IsControl x
+    let IsAscii x = x >= 0 && x <= 127
 
     /// Is this the Vim definition of a blank character.  That is it a space
     /// or tab
--- a/Src/VimCore/FSharpUtil.fs
+++ b/Src/VimCore/FSharpUtil.fs
@@ -375,7 +375,7 @@ module internal GenericListUtil =
     let OfSeq (col: 'T seq) = System.Collections.Generic.List<'T>(col)
 
 [<RequireQualifiedAccess>]
-type internal CharComparer =
+type CharComparer =
     | Exact
     | IgnoreCase
 
@@ -510,7 +510,7 @@ type internal CharSpan
         let length = endIndex - startIndex
         CharSpan(str, startIndex, length, charComparer)
 
-module internal CharUtil =
+module CharUtil =
 
     let PrintableCategories =
         seq {
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -518,9 +518,7 @@ type internal InsertMode
                     // is never a direct insert. But on some international
                     // keyboards, it might translate to an ASCII control
                     // character. See issue #2462.
-                    match Map.tryFind keyInput _commandMap with
-                    | Some rawInsertCommand -> Some rawInsertCommand
-                    | None -> None
+                    None
 
                 else if keyInput.HasKeyModifiers && not (CharUtil.IsLetterOrDigit c) then
 
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -511,14 +511,26 @@ type internal InsertMode
                     let keyInputSet = KeyInputSet(keyInput)
                     RawInsertCommand.InsertCommand (keyInputSet, command, commandFlags) |> Some
 
-                if keyInput.KeyModifiers <> VimKeyModifiers.None && not (CharUtil.IsLetterOrDigit c) then
-                    // Certain keys such as Delete, Esc, etc ... have the same behavior when invoked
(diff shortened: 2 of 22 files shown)


Src/VimCore/FSharpUtil.fs
```diff
@@ -559,6 +559,7 @@ module internal CharUtil =
     let IsWhiteSpace x = System.Char.IsWhiteSpace(x)
     let IsNotWhiteSpace x = not (System.Char.IsWhiteSpace(x))
     let IsControl x = System.Char.IsControl x
+    let IsAscii x = x >= 0 && x <= 127
```
jaredpar: Believe the type of `x` will be `int` here not `char`.

Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -511,14 +512,28 @@ type internal InsertMode
...
-                    // Certain keys such as Delete, Esc, etc ... have the same behavior when invoked
-                    // with or without any modifiers.  The modifiers must be considered because they
-                    // do participate in key mapping.  Once we get here though we must discard them
+                if keyInput.KeyModifiers = VimKeyModifiers.Control then
+
+                    // A key with only the control modifier that isn't mapped
+                    // is never a direct insert. But on some international
+                    // keyboards, it might translate to an ASCII control
+                    // character. See issue #2462.
+                    match Map.tryFind keyInput _commandMap with
```
jaredpar: When will this `match` ever succeed? It looks like a duplicate of the first `match` call in this method.

Src/VimCore/Modes_Normal_NormalMode.fs
```diff
@@ -359,17 +356,8 @@ type internal NormalMode
...
     member x.CanProcess (keyInput: KeyInput) =
-        if _runner.IsWaitingForMoreInput then 
-            true
-        elif _runner.DoesCommandStartWith keyInput then
-            true
-        elif Option.isSome keyInput.RawChar && VimKeyModifiers.None = keyInput.KeyModifiers then
-
-            // We can process any printable character (think international input)
-            // or any character which is part of the standard Vim input set.
-            CharUtil.IsPrintable keyInput.Char || Set.contains keyInput.Char _coreCharSet
```
jaredpar: This old code always smelled a bit to me. I'm glad you've been able to so thoroughly identify the problem here and get a better solution.

jaredpar:
Minor feedback but overall this change looks.

Thanks for the detailed PR message here. Keyboard in put is central to VsVim and it's important to understand the why of our logic here. This particular bug has really been a puzzle to me and I'm so glad you got to the bottom of it.

Src/VimCore/FSharpUtil.fs
ricksladkey: Ha, ha, as much as I love type inference, it does have pitfalls for the unwary!

Src/VimCore/Modes_Insert_InsertMode.fs
ricksladkey: Ah, right. I cribbed that code as a simplification of the code directly after this section, but that other code is actually checking a different key which isn't redundant.

PR #2518, opened 2019-05-04
Convert some tracing to debug build only

### LINKED ISSUE #2517: Reduce the volume of output in the trace window
Some of the output in the trace window is only suitable for a developer and not really useful for general consumption. For example, there is a lot of output associated with the glyph margin, but any problem with the glyph margin is usually possible to describe with a repro, which the a develop and exercise in their own debug build.

It would be good if the output could be pared down to content more focused on helping users in the field diagnose problems with VsVim.

ricksladkey:
#### Issues

- Closes #2517

### DIFF
--- a/Src/VimApp/MainWindow.xaml.cs
+++ b/Src/VimApp/MainWindow.xaml.cs
@@ -127,7 +127,7 @@ namespace VimApp
             InitializeComponent();
 
 #if DEBUG
-            VimTrace.TraceSwitch.Level = TraceLevel.Info;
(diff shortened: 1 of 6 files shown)


jaredpar:
Think this catches me up on code reviews. Sorry I've been a bit slow this week. I'm mostly heads down preparing for my //BUILD talk. Once that's done Tuesday I should be getting back to my usual self.

ricksladkey:
@jaredpar I actually targeted these issues for 2.8.0 with the intent of not creating any urgency to look at these PR's until after the 2.7.0 release! But I am glad that you got a chance to a look at them before.

PR #2519, opened 2019-05-04
Support opening web links with 'gx' and Ctrl+Click

### LINKED ISSUE #2515: Ctrl+Click does not follow link in spite of tooltip
With the improved fidelity of the mouse support, the native Visual Studio `Ctrl+Click` to follow links no longer works. From one point of view, that is expected because gvim does not support that method to follow links. But from another point of view, Visual Studio is displaying a tooltip that says it will work and one could argue that VsVim should support that in the Visual Studio ecosystem.

In any case, VsVim does not yet support gvim's native method of following links, i.e. `gx`. So either of the following or both would be an improvement:

- Implement `gx` to follow links
- Detect links when going to a definition and follow the link instead

ricksladkey:
#### Issues

- Close #2515

### DIFF
--- a/Src/VimCore/CommonOperations.fs
+++ b/Src/VimCore/CommonOperations.fs
@@ -1304,6 +1304,17 @@ type internal CommonOperations
                 let remainder = text.Substring(index)
                 gapText + x.NormalizeBlanks remainder 0
 
+    /// Open link under caret
+    member x.OpenLinkUnderCaret () =
+        match x.WordUnderCursorOrEmpty with
+        | "" ->
+            Result.Failed(Resources.Common_GotoDefNoWordUnderCursor) 
+        | link ->
+            if _vimHost.OpenLink link then
+                Result.Succeeded
+            else
+                Result.Failed(Resources.Common_GotoDefFailed link)
+
     member x.ScrollLines dir count =
         for i = 1 to count do
             match dir with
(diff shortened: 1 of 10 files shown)


Src/VimCore/CommonOperations.fs
```diff
@@ -1123,16 +1123,18 @@ type internal CommonOperations
...
-        else
-            match _wordUtil.GetFullWordSpan WordKind.BigWord _textView.Caret.Position.BufferPosition with
-            | Some(span) -> 
-                let msg = Resources.Common_GotoDefFailed (span.GetText())
-                Result.Failed(msg)
-            | None ->  Result.Failed(Resources.Common_GotoDefNoWordUnderCursor) 
+        match x.WordUnderCursorOrEmpty with
+        | "" ->
+            Result.Failed(Resources.Common_GotoDefNoWordUnderCursor) 
+        | word when word.StartsWith("http") ->
```
jaredpar: This will be a case sensitive comparison.

ricksladkey: Who are these people who use upper case for URLs! :smile:

ricksladkey: I also noticed that will accidentally attempt to open a link when a big-word delimited identifier starts with, say, `Http`, which is actually pretty likely in a lot of networking code. I changed it to case-insensitive prefix of `http:` or `https:`.

PR #2524, opened 2019-05-05
Record all status messages in the trace log

### LINKED ISSUE #2523: Status messages do not appear in the trace log
In some cases, status messages occur in quick succession and so the fact that earlier messages occurred is not immediately apparent.

In general, it would be helpful if status messages appeared in the trace log and in particular, it would be very helpful when status messages occur back-to-back.

ricksladkey:
#### Issues

- Closes #2523

### DIFF
--- a/Src/VimCore/StatusUtil.fs
+++ b/Src/VimCore/StatusUtil.fs
@@ -17,7 +17,7 @@ type internal StatusUtil() =
         and set value = _vimBuffer <- value
 
     member x.DoWithBuffer func (msg: string) = 
-        VimTrace.TraceInfo("Status Start\n{0}", msg)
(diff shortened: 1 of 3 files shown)


Src/VimCore/StatusUtil.fs
```diff
@@ -16,16 +16,21 @@ type internal StatusUtil() =
         with get () = _vimBuffer
         and set value = _vimBuffer <- value
 
-    member x.DoWithBuffer func = 
+    member x.DoWithBuffer func (msg: string) = 
+        VimTrace.TraceInfo("Status Start\n{0}", msg)
```
jaredpar: Should use `Environment.NewLine` or windows style new lines here.

ricksladkey: Right! Fixed.

PR #2525, opened 2019-05-06
Host command selection management

### LINKED ISSUE #2470: map visual studio command ExpandSelectiontoContainingBlock

### LINKED ISSUE #2535: In C++, CTRL-] opens the Find Symbol Results panel instead of going to the definition directly.
When I use CTRL-] in C++, visual studio 2017 opens a Find Symbols Results window instead of directly jumping to the definition. However if I use visual studio's F12 (Edit.GoToDefinition), it is able to jump directly to the definition.

Is there a workout to have CTRL-] have the some behavior as F12?

ricksladkey:
#### Issues

- Closes #2470
- Closes #2535

##### Selection Management

This is an alternative approach to PR 2473 that addresses the issue in #2470, one that is designed to prevent the need for VsVim to maintain any record of which Visual Studio commands do or do not need the selection cleared. Instead, it adds a special "has bang" version of `:vsc! [command]` which means: execute the command and then clear the selection. Any `:vsc [command]` without the bang does not modify the selection.

Users who have any existing mappings for keys where the old selection clearing logic was chosen and where that behavior was desirable will need to add bang to their mappings. I don't know of any such commands offhand where this would be the case, but I imagine there might be some.

In any case, the user now has full control of the selection clearing logic in those few situations where it was necessary. Because vim does not support the `:vsc` command, how VsVim behaves when you use it cannot be incompatible with vim. And it is still true, in general, that, like vim, VsVim usually clears the selection after any colon-command, just not for `vsc` commands.

##### Go To Definition / Go To Declaration

This change also includes a revised approach to the long-standing problem of the `Edit.GoToDefinition` command when used with the C++ language service. As a reluctant but frequent C++ programmer, I am very excited about this change as it makes `<C-]>` work **exactly** like `F12` and `gd` work **exactly** like `Edit.GoToDeclaration`.

The solution for C++ is to use a different Visual Studio API for command execution so that we can post the command asynchronously instead of executing it directly. I don't know why this works but it does.

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/VisualStudioCommandDispatcher.cs
+++ b/Src/VsVimShared/Implementation/Misc/VisualStudioCommandDispatcher.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.ComponentModel.Composition;
+using EnvDTE;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace Vim.VisualStudio.Implementation.Misc
+{
+    [Export(typeof(ICommandDispatcher))]
+    internal class VisualStudioCommandDispatcher : ICommandDispatcher
+    {
+        private readonly _DTE _dte;
+        private readonly IVsUIShell _uiShell;
+
+        [ImportingConstructor]
+        internal VisualStudioCommandDispatcher(SVsServiceProvider serviceProvider)
+        {
+            _dte = serviceProvider.GetService<SDTE, _DTE>();
+            _uiShell = serviceProvider.GetService<SVsUIShell, IVsUIShell>();
+        }
+
+        public bool ExecuteCommand(ITextView textView, string command, string args, bool postCommand)
+        {
+            // Many Visual Studio commands expect the focus to be in the editor
+            // when  running.  Switch focus there if an appropriate ITextView
+            // is available.
+            if (textView is IWpfTextView wpfTextView)
+            {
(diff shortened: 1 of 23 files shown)


Src/VsVimShared/Implementation/Misc/VisualStudioCommandDispatcher.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            // Some commands like 'Edit.GoToDefinition' only work like they do
+            // when they are bound a key in Visual Studio like 'F12' when they
+            // are posted instead of executed synchronously. See issue #2535.
+            if (postCommand)
+            {
+                var dteCommand = _dte.Commands.Item(command, 0);
+                var guid = new Guid(dteCommand.Guid);
+                return _uiShell.PostExecCommand(ref guid, (uint)dteCommand.ID, 0, args) == VSConstants.S_OK;
```
jaredpar: Curious: how did you discover this approach? This is the first I've seen of it.

ricksladkey: When researching the background of this bug, I ran into this StackOverflow question and answer:

- [How do I use DTE.ExecuteCommand(“Edit.GoToDefinition”) in a VS2010 C++ macro?
](https://stackoverflow.com/questions/20168945/how-do-i-use-dte-executecommandedit-gotodefinition-in-a-vs2010-c-macro)

PR #2531, opened 2019-05-08
Support inserting literal characters

### LINKED ISSUE #2020: Weird behavior after clicking into an escape sequence
Visual Studio displays non-printable characters in source code using the according terminal escape sequence. A string containing the `SOH` character (`'\1'`) is thus displayed like this:

```
public static class Program
{
    public static void Main()
    {
        var foobar = "foo^Abar";
    }
}
```

The escape sequence (`^A`) is highlighted accordingly. Moving over the escape sequence using `hjkl` in VsVim works great, however if the user clicks into the escape sequence using the mouse, the text cursor will be moved into the sequence and stay in insert mode. All keys pressed are ignored by VsVim and instead inserted between the carret and the according escape character. Trying to move out of such a sequence will thus result in the string looking like this: `"foo^jjjjAbar"`.

### LINKED ISSUE #2285: Insert mode unicode input
Need to implement <kbd>Ctrl-V,u</kbd> and <kbd>Ctrl-V,U</kbd> for insert mode. Found this to be invaluable when i'm tracking down surrogate pair isssues.

### LINKED ISSUE #2467: Neither Ctrl-V not Ctrl-Q quote Ctrl characters or the escape key
Hi!
Maybe I just haven't found this in the documentation yet, but I find myself unable to devine macros with the Esscape key or any other not-standard keys in them.
On Unix, Ctrl-V used to be the "insert the next character literally", like Escape or Ctrl-M. On windoes gvim. it's either Ctrl-V (vi mapping) or Ctrl-Q (windows mapping).

So, how do I put an Escape-Character or Return or a control character in a macro (i.e in the colon mode) using vsvim?
like for prepending a line with a text and moving to the next line
:map v Icooltext<Escape?>j
or, for, say, splitting the records of a .csv
: %s/,/<Return?>/g

ricksladkey:
#### Issues

- Closes #2020
- Closes #2285
- Closes #2467

### DIFF
--- a/Src/VimCore/Modes_Insert_InsertMode.fs
+++ b/Src/VimCore/Modes_Insert_InsertMode.fs
@@ -1088,9 +1088,13 @@ type internal InsertMode
     /// Process the character of a literal insertion
     member x.ProcessLiteral (keyInput: KeyInput) = 
         _sessionData <- { _sessionData with ActiveEditItem = ActiveEditItem.None }
-        keyInput.Char
-        |> string
-        |> x.InsertText
+        let insertCommand =
+            keyInput.Char
+            |> string
+            |> InsertCommand.InsertLiteral
+        let keyInputSet = KeyInputSet keyInput
+        let commandFlags = CommandFlags.Repeatable ||| CommandFlags.InsertEdit
+        x.RunInsertCommand insertCommand keyInputSet commandFlags
 
     // Insert the raw characters associated with a key input set
     member x.InsertText (text: string): ProcessResult =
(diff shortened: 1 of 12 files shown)


Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -329,6 +332,7 @@ type internal InsertMode
                 ("<C-g>", RawInsertCommand.CustomCommand this.ProcessUndoStart)
                 ("<C-^>", RawInsertCommand.CustomCommand this.ProcessToggleLanguage)
                 ("<C-k>", RawInsertCommand.CustomCommand this.ProcessDigraphStart)
+                ("<C-q>", RawInsertCommand.CustomCommand this.ProcessLiteralStart)
```
jaredpar: Did not realize this was an alternate command for `control-v`.

ricksladkey: Given that 99% of the users who type `<C-v>` in insert mode in Visual Studio, will continue to expect it to work as paste from clipboard, I chose not to also assign that key. This is the old "do we default to Windows-friendly shortcuts on Windows or not." We could have a separate discussion about it. In the meantime, anyone who wants the pure vim `<C-v>` experience can map it to `<C-q>` in insert mode.

PR #2540, opened 2019-05-12
Change block insert to repeat the insert instead of inserting a string

### LINKED ISSUE #2420: Inserting tabs in block selection still does not work properly
If I have the following code:
```
   123456789
1: aVariableCalledA = b +
2:     c +
3:     d;
```

If I select text the line 2 and 3 in column 5 using a visual block selection,
press ``I`` to insert a tabulator, the result is:
```
   12345678901234
1: aVariableCalledA = b +
2:         c +
3:             d;
```
Expected was:
```
   1234567890
1: aVariableCalledA = b +
2:         c +
3:         d;
```
Our VS tab settings are:
- tab size: 8
- indent size: 4
- insert spaces: true

For the release 2.6.0.0 issues have been closed adressing this issue, so I'm surprised to see this still going wrong.

ricksladkey:
#### Issues

- Closes #2420

### DIFF
--- a/Src/VimCore/InsertUtil.fs
+++ b/Src/VimCore/InsertUtil.fs
@@ -183,22 +183,22 @@ type internal InsertUtil
             TextViewUtil.MoveCaretToPosition _textView position
         | None -> textEdit.Cancel()
 
-    /// Apply the given TextChange to the specified BlockSpan 
-    member x.ApplyBlockInsert (text: string) atEndOfLine startLineNumber spaces height = 
+    /// Apply the given TextChange to the specified BlockSpan
+    member x.ApplyBlockInsert (insertCommand: InsertCommand) atEndOfLine startLineNumber spaces height =
 
-        // Don't edit past the end of the ITextBuffer 
-        let height = 
+        // Don't edit past the end of the ITextBuffer.
+        let height =
             let maxHeight = x.CurrentSnapshot.LineCount - startLineNumber
             min maxHeight height
 
-        // It is possible that a repeat of a block edit will begin part of the way through
-        // a wide character (think a 4 space tab).  If any edits have this behavior the first
-        // pass will break them up into the appropriate number of spaces
-        let fixOverlapEdits () = 
+        // It is possible that a repeat of a block edit will begin part of the
+        // way through a wide character (think a 4 space tab).  If any edits
+        // have this behavior the first pass will break them up into the
+        // appropriate number of spaces
+        let fixOverlapEdits () =
             use textEdit = _textBuffer.CreateEdit()
-
             let currentSnapshot = x.CurrentSnapshot
-            for i = 0 to (height - 1) do 
+            for i = 0 to height - 1 do
 
                 let lineNumber = startLineNumber + i
                 let line = SnapshotUtil.GetLine currentSnapshot lineNumber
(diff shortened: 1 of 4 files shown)


Src/VimCore/InsertUtil.fs
```diff
@@ -208,70 +208,46 @@ type internal InsertUtil
...
-                    let text =
-                        if _localSettings.ExpandTab then
-                            StringUtil.ExpandTabsForColumn text column.ColumnNumber tabStop
-                        else
-                            text
-
-                    // Unfortunately, we may have to delete some of the preceding
-                    // spaces if they are no longer needed to maintain the
-                    // column position.
-                    if text.StartsWith("\t") then
```
jaredpar: Did you verify that we have block insert tests that cover this scenario?

Src/VimCore/InsertUtil.fs
```diff
@@ -208,70 +208,46 @@ type internal InsertUtil
...
 
-            if textEdit.HasEffectiveChanges then 
+            if textEdit.HasEffectiveChanges then
                 textEdit.Apply() |> ignore
             else
                 textEdit.Cancel()
 
-        // Actually apply the edit to the lines in the block selection
-        let doApply () = 
-            use textEdit = _textBuffer.CreateEdit()
```
jaredpar: This line gave me a bit of pause. I'm trying to remember why I implemented this as a direct `ITextEdit` operation vs. a series of `ITextBuffer` edit operations. 

The only thing I can think of off the top of my head is perf / simplicity. The nice part of `ITextEdit` is that we don't have to worry about other extensions doing edit in between every single line change. That probably isn't a big deal though here. Also doing it as a single edit makes perf "better" but probably not discernably so.

Mostly commenting here so others are aware of this in case it comse up again. I can't see anything wrong with this but it tickles my brain a bit.

Src/VimCore/InsertUtil.fs
ricksladkey: I did not check, but the code itself is just trying to mimic what would automatically occur if a tab insert command were run. In other words, the old code was inserting a literal string, but realized that if they user had actually typed tab as part of an insert, it would (as vim does) "fix up" any no longer necessary previous spaces.

This is another benefit of repeating the insert instead of inserting strings: this behavior occurs naturally as part of the insertion command instead of being hacked into block insertion with special (and duplicated) handling.

I just confirmed that it does indeed work correctly. I will check whether there is coverage for this scenario in any of the existing tests.

Src/VimCore/InsertUtil.fs
ricksladkey: At the moment, there is no way to request that insertion command execution occur in the context of a text edit, and as a result, there is no easy way to batch insertion commands other than to make sure that they are inside of an enclosing undo transaction. But, they already are.

In this particular case, the repeat loop for each  subsequent line in the block insert correctly uses no state information from the previous snapshot and so benefit of performing all the edits against a single snapshot is unnecessary.

It could be argued that a single text edit is more efficient, but given that insert commands currently operate on the latest snapshot, it would require a major reorganization to try to achieve a very modest performance gain in a fairly uncommon insertion use case.

The most important thing is that undo undoes the whole block insertion, and it does because of the enclosing undo transaction.

Src/VimCore/InsertUtil.fs
ricksladkey: OK, I checked and there is a test:

```csharp
                /// <summary>
                /// A block inserted tab with 'noexpandtab' should eat preceding spaces if possible
                /// </summary>
                [WpfFact]
                public void UnalignedNoExpandTabs()
```

PR #2545, opened 2019-05-14
Include line numbers in parser errors

### LINKED ISSUE #1153: Show line numbers in parse errors
Is there a way to see the output of the parse error?

Unfortunately I have a rather large vimrc that's shared across gvim, vsvim and MacVim. Since my local *rc files source the main vimrc (held in my dropbox) I make changes at home (for MacVim) and when I return to work (VsVim) I'm getting "parse error". The problem is I have no idea what's causing the parse error. I can't remember what all I changed (after I fix it this time I'm initializing a git repo in this directory). 

On several occasions I've resorted to commenting lines until I found it.

ricksladkey:
#### Issues

- Closes #1153

### DIFF
--- a/Test/VimWpfTest/CommandLineEditIntegrationTest.cs
+++ b/Test/VimWpfTest/CommandLineEditIntegrationTest.cs
@@ -47,7 +47,8 @@ namespace Vim.UI.Wpf.UnitTest
                 VimEditorHost.EditorFormatMapService.GetEditorFormatMap(_vimBuffer.TextView),
                 VimEditorHost.ClassificationFormatMapService.GetClassificationFormatMap(_vimBuffer.TextView),
                 CommonOperationsFactory.GetCommonOperations(_vimBuffer.VimBufferData),
-                _clipboardDevice);
+                _clipboardDevice,
+                false);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Test/VimWpfTest/CommandLineEditIntegrationTest.cs
```diff
@@ -47,7 +47,8 @@ protected virtual void Create(params string[] lines)
                 VimEditorHost.EditorFormatMapService.GetEditorFormatMap(_vimBuffer.TextView),
                 VimEditorHost.ClassificationFormatMapService.GetClassificationFormatMap(_vimBuffer.TextView),
                 CommonOperationsFactory.GetCommonOperations(_vimBuffer.VimBufferData),
-                _clipboardDevice);
+                _clipboardDevice,
+                false);
```
jaredpar: Consider passing the constant by name here so the intent is clear: `isFirstCommandMargin: false`

ricksladkey: Good idea.

ricksladkey: Done.

PR #2547, opened 2019-05-16
Ensure new window view properties are met

### LINKED ISSUE #2546: Loading a file doesn't scroll the new window correctly unless 'scrolloff' is set
If `:set scrolloff=0` is in effect (the default), when navigating to an unloaded mark, or going to a tag (e.g using `:vimhelp`), the caret is properly positioned in the new window, but the window does not scroll to the ensure that the caret is visible.

ricksladkey:
#### Issues

- Closes #2546

#### Discussion

The goal of this change is to allow VsVim to faithfully manage window properties like "visible" and "scroll offset" according to the Vim model, even though the window is newly created. To accomplish this we need to schedule an action to run at a future time. If we perform the action too early, or if we give up, then we lose the opportunity to maintain control over the view into the buffer.

Here is the new flow of control when loading a file into a new window where we want to set the line and column position:

- We enter the new common operation `LoadFileIntoNewWindow`.
- Initially the window doesn't exist
- We call into `LoadFileIntoNewWindow` on the vim host
- That method hands the task off to Visual Studio
- Still synchronously, we can gain access to the text view corresponding to that window
- We set the caret position in the new window
- The window is still being constructed and so we cannot scroll the window yet
- We return the new partially initialized text view back to common operations
- We use VsVim infrastructure to translate that text view into an `IVimBufferData`
- We use new infrastructure to to acquire the common operations for that `IVimBufferData`
- We call `EnsureAtCaret` on the new common operations
- We are now again in common operations, but on the correct text view
- When we go to ensure the view properties, we notice that the text view is not ready
- We schedule a synchronous version of `EnsureAtPoint` to run when the text view **is** ready
- Finally, `EnsureAtPoint` runs as early as possible, but only when the the text view is ready

So you can see that this requires a level of cooperation **between** common operations instances that we previously have not needed. It also requires a way to tell if the text view is ready which is not something we can do properly from `VimCore` because we need WPF and `VimCore` intentionally does not have access to WPF. What this means is that new infrastructure was needed in common operations and in `IVimHost` in order to support this kind of capability.

I hope this explains why so many changes were necessary for such a seemingly innocent task!


Src/VimCore/CommonOperations.fs
```diff
@@ -149,6 +150,45 @@ type internal CommonOperations
...
+            doAction()
+        else
+
+            // Otherwise, keep checking on each layout change whether the text
+            // view is ready for the action yet.
+            let bag = DisposableBag()
+            _textView.LayoutChanged
+            |> Observable.subscribe (fun _ -> tryDoAction bag)
+            |> bag.Add
+
```
ricksladkey: It occurred to me that we could convert `Vim.SetupInitialMode` to use this mechanism but I had a brain freeze trying to decide where it should go so that `Vim`, `CommonOperations`, and `SelectionChangeTracker` could all access it.

jaredpar: It does seem like we should share the code between the this and `SetupInitialMode`; the coding pattern here is so similar that it seems to beg for a shared helper. Particularly because `SetupInitialMode` and this code has a different concept of "ready": one lets the host determine and the other checks for the presence of `ITextViewLines`. 

The code in `SelectionChangeTracker` is a bit different. It's not concerned with the "ready" state of the `ITextBuffer` nor does it use `LayoutChanged`. It only cares about the closed state. The pattern around `SynchronizationContext` is there though in all three places. 

After thinking about it a bit I believe there are two helpers we should add to `ICommonOperations`:

- `Post`:  this abstracts away the `SynchronizationContext.Post` pattern. The contract being the provided delegate will only run if the `ITextView` is still open after the post operation completes.
- `DoWhenReady`: the method you have today but delegating to `Post` for the `SynchronizationContext` bit. 







---
In reply to: [284557364](https://github.com/VsVim/VsVim/pull/2547#discussion_r284557364) [](ancestors = 284557364)

Src/VimCore/CommonOperations.fs
```diff
@@ -2152,6 +2209,14 @@ type internal CommonOperations
             x.AdjustTextViewForScrollOffsetAtPoint point
         if Util.IsFlagSet viewFlags ViewFlags.VirtualEdit && point.Position = x.CaretPoint.Position then
             x.AdjustCaretForVirtualEdit()
+    
+    /// Ensure the view properties are met at the caret
+    member x.EnsureAtCaret viewFlags = 
+        x.DoActionWhenReady (fun () -> x.EnsureAtPointSync x.CaretPoint viewFlags)
```
jaredpar: During the initial load is it possible for this method to get invoked multiple times hence resulting in it executing multiple times once the underlying `ITextView` is ready? These functions are idempotent in isolation hence so I don't think this is a problem even if it does happen. But wanted to make sure you agree on this.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4664,6 +4664,9 @@ type IVimHost =
     /// Is the ITextBuffer in a dirty state?
     abstract IsDirty: textBuffer: ITextBuffer -> bool
 
+    /// Whether the ITextView is loaded
```
jaredpar: Given the use case here think we should add something to the comment that indicates the view is ready for operations to be performed on it.

Src/VimWpf/VimHost.cs
```diff
@@ -219,6 +219,15 @@ public virtual bool IsDirty(ITextBuffer textBuffer)
             return false;
         }
 
+        public virtual bool IsLoaded(ITextView textView)
+        {
+            if (textView is IWpfTextView wpfTextView)
+            {
+                return wpfTextView.VisualElement.IsLoaded;
```
jaredpar: Curious: why does the trick in `SetupInitialMode`, where we just check to see if we can get the lines, not work for this change?

jaredpar:
Done with review pass (iteration 8).

Src/VimWpf/VimHost.cs
ricksladkey: In my debugging, the `GetTextViewLines` became available, but it did not reflect the final state of the text view. For example, our viewport is (say) fifty lines high, but the text view lines collection only had length two. The method `SetupInitialMode` only cares that Visual Studio has properly initialized the view settings (like tabstop and shiftwidth); it doesn't care about the actual view itself. By contrast, when we are ensuring view properties like "is visible" or "scroll offset", we really need an accurate picture of text view in relation to the viewport.

Prior to WPF transitioning the element to `IsLoaded <- true`, the VisualElement is not even part of the visual tree yet. It has no parent element, blah, blah, blah, and so it couldn't possibly be sized correctly or answer our critical viewport related questions or operations properly yet.

Src/VimCore/CommonOperations.fs
ricksladkey: All three cases want the code to run:

1. In the future but as soon as possible, and
2. Between when the text view becomes ready and when it is closed

We can and do receive an external caret position changed event before the window is ready for us to apply our view properties. It is a historical artifact that this code is even part of the selection change tracker as it has nothing directly to do with selection change tracking.

In the old code, the caret position changed event was only being watched in order to enforce scroll offset and so it had no effect when scroll offset was zero. It turns out that we need to enforce all our view properties (not just scroll offset) and so that event runs more often.

The problem with ICommonOperations is that IVim doesn't have direct access to ICommonOperations, maybe intentionally or maybe because of mutual dependency. If IVim even had the ICommonOperations factory, it could be done, but anyway that decision was what stalled my first attempt at unifying them.

Src/VimCore/CommonOperations.fs
ricksladkey: Yes, we already had the issue of it possibly running multiple times, but those multiple times would be synchronous, not posted. The `EnsureAtCaret` is indeed idempotent as long as the buffer is unmodified and the caret isn't moved. I believe that posted actions are stable in that if action A is posted before action B, A will be scheduled before B.

However, since we don't **always** post, it is possible for us to post an `EnsureAtCaret` and a synchronous call to `EnsureAtCaret` to run first. However, in that case I believe that is just the order we would want them to run in. In either case, it is not non-deterministic.

Src/VimCore/CoreInterfaces.fs
ricksladkey: I'll do that. Now that I'm thinking about it, I wonder whether it should be named literally. The entire reason we need host support is that we need information that only the IWpfTextView can give us and we don't have access to WPF from Vim.Core by design. Hmm, I think the host should provide the DoWhenReady functionality and then we no longer need IsLoaded.

Src/VimCore/CommonOperations.fs
ricksladkey: Changes based on your comments:

- I implemented a fully WPF-aware version of `DoActionWhenTextViewReady` in `IVimHost`
- I forwarded `DoActionWhenReady` in `ICommonOperations` to `DoActionWhenTextViewReady` in `IVimHost`
- I added `DoActionWhenReady` to the `ICommonOperations` interface
- I did the work necessary so that `IVim` can use `ICommonOperations` for a specific `IVimBuffer`
- I refactored `SetUpInitialMode` to use the `DoActionWhenReady`
- I moved `OnCaretPositionChanged` out of `SelectionChangeTracker` and into `Vim`

ricksladkey:
Ready for re-review.

Src/VimWpf/VimHost.cs
```diff
@@ -224,6 +224,35 @@ public virtual bool IsDirty(ITextBuffer textBuffer)
...
+                    action.Invoke(null);
+                }
+            }
+            if (textView is IWpfTextView wpfTextView && !wpfTextView.VisualElement.IsLoaded)
+            {
+                // FrameworkElement.Loaded Event:
+                //
+                // Occurs when a FrameworkElement has been constructed and
+                // added to the object tree, and is ready for interaction.
+                wpfTextView.VisualElement.Loaded += (sender, e) => doAction();
```
jaredpar: Think we need to unsubscribe to this event after it fires. Otherwise the delegate, which is likely to have initialization data only, will be held in memory for the lifetime of the `IWpfTextView`

Src/VimCore/SelectionChangeTracker.fs
```diff
@@ -98,43 +94,6 @@ type internal SelectionChangeTracker
         else
             x.SetModeForSelection()
 
-    /// If the caret changes position and it wasn't initiated by VsVim then we should be 
-    /// adjusting the screen to account for 'scrolloff'
```
jaredpar: How do we account for non-VsVim caret movements now?

jaredpar: NM, found that this function moved.

---
In reply to: [287606028](https://github.com/VsVim/VsVim/pull/2547#discussion_r287606028) [](ancestors = 287606028)

Src/VimWpf/VimHost.cs
ricksladkey: I was thinking that the vim buffer and the text view have the same lifetimes so there would not be a issue with leaking, but if we unsubscribe there is definitely no risk, so that is a better solution. I've made that change.

Src/VimCore/SelectionChangeTracker.fs
ricksladkey: Right, if the selection change tracker were called ExternalEventHandler then it would have made sense to keep the handling of external selection events and external caret events together. As it is, the selection change tracker now only tracks selection changes.

PR #2552, opened 2019-05-18
Support vim modelines

### LINKED ISSUE #886: Support vim modelines
It would be great if VsVim handled vim modelines.  I often use modelines to set up per-file tab/indent preferences. For example:

```
# vim:ts=4:sw=4:et
```

Thanks for this excellent extension!

### LINKED ISSUE #2548: Vim help files use the user's tab settings rather than those specified in the files
Inline vim help are classified as Visual Studio text files and currently display using the user's tab settings, which are often some other value than eight. The help files use tabs extensively and expect the setting `:set tabstop=8` to be in effect. This is accomplished via the modeline mechanism, such as this line near the bottom of help.txt:

```
 vim:tw=78:isk=!-~,^*,^\|,^\":ts=8:noet:ft=help:norl:
```

Although full modeline support is desirable, what is mostly needed for this issue is automatically setting `ts=8` in vim help files.

ricksladkey:
#### Issues

- Closes #886
- Closes #2548

### DIFF
--- a/Src/VimCore/ModeLineInterpreter.fs
+++ b/Src/VimCore/ModeLineInterpreter.fs
@@ -15,8 +15,8 @@ type internal ModeLineInterpreter
 
     /// Regular expressions to parse the modeline
     static let _escapedModeLine = @"(([^:\\]|\\:?)*)";
-    static let _firstPattern = @"[ \t]vim:[ \t]*set[ \t]+" + _escapedModeLine + ":"
-    static let _secondPattern = @"[ \t]vim:(.*)$"
+    static let _firstPattern = @"[ \t]vim:(.*)$"
+    static let _secondPattern = @"[ \t]vim:[ \t]*set[ \t]+" + _escapedModeLine + ":"
     static let _nextGroup = _escapedModeLine + @"(:|$)"
     static let _settingPattern = @"([\w[\w\d_]*)";
     static let _assignment = @"^" + _settingPattern + @"=(.*)$"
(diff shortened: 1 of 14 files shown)


Src/VimCore/ModeLineInterpreter.fs
```diff
@@ -0,0 +1,186 @@
...
+               result
+
+        // Try to process the first few and last few lines as modelines.
+        let tryProcessModeLines modeLines =
+            let lineCount = _textBuffer.CurrentSnapshot.LineCount
+            let snapshot = _textBuffer.CurrentSnapshot
+            seq {
+                yield seq { 0 .. min (modeLines - 1) (lineCount - 1) }
+                yield seq { max modeLines (lineCount - modeLines) .. lineCount - 1 }
+            }
```
jaredpar: Isn't this seq just equivalent to all the line numbers in the snapshot?

Src/VimCore/ModeLineInterpreter.fs
```diff
@@ -0,0 +1,186 @@
...
+            else
+                false
+
+        // Split the options string into fields.
+        let splitFields (options: string) =
+            options.Replace(@"\:", ":").Split(' ', '\t')
+
+        // Process the "first" format of modeline, i.e. "vim: ...".
+        let processFirstForm modeLine =
+            let m = Regex.Match(modeLine, _firstPattern)
```
jaredpar: This means that for every line we're constructing the `Regex` object for `_firstPattern` and then executing it fresh against the current line. Worried that could be expensive. Did you consider making the fields be of type `Regex` and using the compiled option to ensure speedy execution here?

Src/VimCore/ModeLineInterpreter.fs
```diff
@@ -0,0 +1,186 @@
...
+                    |> Seq.map (fun m -> splitFields m.Groups.[1].Value)
+                    |> Seq.concat
+                    |> Seq.tryFind (fun option -> not (processOption option))
+                Some modeLine, firstBadOption
+            else
+                None, None
+
+        // Process the "second" format of modeline, i.e. "vim: set ... :".
+        let processSecondForm modeLine =
+            let m = Regex.Match(modeLine, _secondPattern)
```
jaredpar: Same feedback about constructing `Regex` objects.

Src/VimCore/ModeLineInterpreter.fs
ricksladkey: The default for `modelines` is `5`, so typically we search the first five and the last five lines. The min / max calculations here handle files that are shorter than 10 lines long.  Let's say there are 100 lines. Then the expression is like:

```
seq {
    yield seq { 0 .. 4 }
    yield seq { 95 .. 99 }
}
```

Src/VimCore/ModeLineInterpreter.fs
ricksladkey: That's a good point; I'll make that change.

Src/VimCore/ModeLineInterpreter.fs
ricksladkey: Added a test to prove that a modeline in the middle of a large file is not processed.

Src/VimCore/ModeLineInterpreter.fs
ricksladkey: Changed these all to static compiled regular expression objects.

PR #2556, opened 2019-05-21
Only commands typed interactively enter the command history

### LINKED ISSUE #2318: Commands issued as part of a mapping should not be recorded in the command history
In vim, commands issued as part of a mapping are not recorded in the command history. VsVim does not distinguish between commands typed by the user and commands issued as part of a mapping.

Repro:

```
:nmap zz :echo "foo"<CR>
```

Type `zz` in normal mode and `foo` appears in the status bar.

Type `:<C-p>` and the command is populated with `:echo "foo"`

ricksladkey:
#### Issues

- Closes #2318

#### Discussion

I tried so many times to fix this bug without making a complete mess of things and I think I've finally come up with a fairly clean and minimally invasive solution. The issue is conceptually a simple one:

- Suppress the normal mechanism of adding a command to the command history when that command was executed as the result of a keyboard mapping.

The problem is that whether a given key input is mapped and the point at which we execute a command are removed from each other by many processing layers. To solve the problem, we have to carry metadata about the key input along with the key input as it weaves its way through the many layers. As a result, the time-honored `IMode.Process(KeyInput)` method has now become `IMode.Process(KeyInputData)` and the key mapping infrastructure now has to report whether a given key input was the result of a mapping as it is doing its work.

The hardest part about this was that the `BindData<'T>` family of types that pervades the key processing of VsVim is "hard-coded" to operate on the `KeyInput` type. I tried four different approaches to extend these types to be able to also operate on a `KeyInputData`.

1) I tried simply changing `BindData<'T>` so that it is "hard-coded" to operate on the `KeyInputData` type. This led to sweeping changes of code that hasn't changed for years and is utterly uninterested in the benefit that `KeyInputData` provides.

2) I tried making `BindData<'T>` "more" generic and thus, becomes `BindData<'S, 'T>` where `'S` is usually `KeyInput` but can be `KeyInputData`. This also leads to sweeping changes because much of the unaffected code needs to supply or facilitate the resolution of an addition type. Furthermore, `BindData` which is already obtuse, becomes even more unwieldy to understand and maintain.

3) I tried make approach 2 but named it `GenericBindData<'S, 'T>` and then created specialized subtypes `BindData<'T>` as an alias for `GenericBindData<KeyInput, 'T>` and `MappedBindData<'T>` as an alias for `GenericBindData<KeyInputData, 'T>`. This **almost** worked, because F# supports generic type aliases, but unfortunately, C# does not. That meant new sweeping changes to the test code which is written almost entirely in C#.

4) Finally, I created a parallel type called `MappedBindData<'T>` that is just like `BindData<'T>` except that it operates on `KeyInputData` instead of `KeyInput`. This approach led to far less sweeping changes, but has two disadvantages. First, it leads to some code repetition that seems like it "ought" to be factored out as commonality. Second, we need a way to convert from one kind o bind data types to the other kind so that the old code and the new code can interoperate at the interfaces.

I ended up going with the fourth approach and accepted its trade-offs as a reasonable compromise to keeping the changes from spiraling out of control.

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -3647,8 +3647,8 @@ type BindDataStorage<'T> =
         | Simple bindData -> Simple (bindData.Convert mapFunc)
         | Complex func -> Complex (fun () -> func().Convert mapFunc)
 
-/// This is the result of attemping to bind a series of KeyInput values into a Motion
-/// Command, etc ... 
+/// This is the result of attemping to bind a series of KeyInputData values
+/// into a Motion Command, etc ...
 [<RequireQualifiedAccess>]
 type MappedBindResult<'T> = 
 
@@ -3666,8 +3666,8 @@ type MappedBindResult<'T> =
 
     with
 
-    /// Used to compose to MappedBindResult<'T> functions together by forwarding from
-    /// one to the other once the value is completed
+    /// Used to compose to MappedBindResult<'T> functions together by
+    /// forwarding from one to the other once the value is completed
     member x.Map (mapFunc: 'T -> MappedBindResult<'U>): MappedBindResult<'U> =
         match x with
         | Complete value -> mapFunc value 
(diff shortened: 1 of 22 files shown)


Src/VimCore/CoreInterfaces.fs
```diff
@@ -3644,6 +3647,119 @@ type BindDataStorage<'T> =
         | Simple bindData -> Simple (bindData.Convert mapFunc)
         | Complex func -> Complex (fun () -> func().Convert mapFunc)
 
+/// This is the result of attemping to bind a series of KeyInputData values
+/// into a Motion Command, etc ...
```
ricksladkey: See the discussion section of this PR for an explanation of the mapped bind family of types.

Src/VimCore/KeyInput.fs
```diff
@@ -81,6 +81,17 @@ type KeyInput
...
+[<Sealed>]
+type KeyInputData
+    (
+        _keyInput: KeyInput,
+        _wasMapped: bool
+    ) =
+
+    member x.KeyInput = _keyInput
+    member x.WasMapped = _wasMapped
+    static member Create keyInput wasMapped = KeyInputData(keyInput, wasMapped)
```
jaredpar: Consider adding a `ToString` override to facilitate debugging

Src/VimCore/CoreInterfaces.fs
jaredpar: I agree with the sentiment of your PR description here: this feels like it should be factorable. But I can't think of a clean way to do this. It's a small amount of repetition though so I don't feel like it's too bad. 


---
In reply to: [285844744](https://github.com/VsVim/VsVim/pull/2556#discussion_r285844744) [](ancestors = 285844744)

jaredpar:
Thanks for the detailed write up here. I agree that 4 seems like the best approach here. The code changes are much less invasive than I expected them to be and the approach just makes sense as I read through the changes.

Src/VimCore/KeyInput.fs
ricksladkey: Great idea. Done.

image

PR #2568, opened 2019-05-25
More Azure DevOps experimentation

jaredpar:
This change is focusing on getting both the test configurations running in a single job.


jaredpar:
Next step is getting publishing to the Open VSIX Gallery working again.

PR #2571, opened 2019-05-27
Add support for relative line numbers (take two)

### LINKED ISSUE #1909: relativenumbers

ricksladkey:
#### Issues

- Closes #1909

#### Discussion

This is a revived and updated PR based on @oxilumin's PR #2452. I haven't made any changes other than merging with master, resolving conflicts, and fixing some build warnings related to the Visual Studio Threading assembly references. I have tested it against Visual Studio 2019 and it works as advertised.

I think Alex is not responding to inquiries and he is perhaps focused on other projects. But I think the old PR was almost merged and, it does work, and so many people have asked for this feature, that I am resubmitting his work as a new PR for review.


ricksladkey:
Screenshot of relative line numbers in action:

image

ricksladkey:
Builds locally with VS2019 but one test fails on appveyor:

```
    Vim.VisualStudio.UnitTest.ReferenceTest.EnsureCorrectVisualStudioVersion [FAIL]
      System.Exception : Unrecognized reference Microsoft.VisualStudio.Threading
      Stack Trace:
        Test\VsVimTest\ReferenceTest.cs(235,0): at Vim.VisualStudio.UnitTest.ReferenceTest.AssemblyData.GetReferenceData(AssemblyName assemblyName)
        Test\VsVimTest\ReferenceTest.cs(134,0): at Vim.VisualStudio.UnitTest.ReferenceTest.AssemblyData.GetTransitiveReferenceData(Assembly assembly)
        Test\VsVimTest\ReferenceTest.cs(258,0): at Vim.VisualStudio.UnitTest.ReferenceTest.<>c.<EnsureCorrectVisualStudioVersion>b__4_0(Assembly a)
           at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
        Test\VsVimTest\ReferenceTest.cs(258,0): at Vim.VisualStudio.UnitTest.ReferenceTest.EnsureCorrectVisualStudioVersion()
  Finished:    VsVim.UnitTest
=== TEST EXECUTION SUMMARY ===
```

I'll see if I can figure out what is going on.

ricksladkey:
Couple of problems:

1. I cannot get `:set linenumber` to take effect as a local setting for all buffers. Still looking into this.
2. VimApp crashes trying to use `ThreadHelper.JoinableTaskFactory` which doesn't work outside Visual Studio:

Here is a workaround for the second problem:

```c#
            // See:
            // https://github.com/microsoft/vs-threading/blob/master/doc/testing_vs.md
            //_joinableTaskFactory = ThreadHelper.JoinableTaskFactory;

            var jointableTaskContext = new JoinableTaskContext();
            _joinableTaskFactory = jointableTaskContext.Factory;
```

But it's not clear what is the best way to implement this so that both VijmApp and VsVim choose the right factory.

jaredpar:
> But it's not clear what is the best way to implement this so that both VijmApp and VsVim choose the right factory.

Can we just use the concrete types in `Microsoft.VisualStudio.Threading` directly? Despite its name the DLL doesn't actually depend on Visual Studio. Roslyn uses this DLL in it's testing and they don't depend on VS being on the machine for build + test.

ricksladkey:
@jaredpar I came up with a solution using MEF like:

```c#
    [Export(typeof(IJoinableTaskFactoryProvider))]
    internal class VsVimJoinableTaskFactoryProvider : IJoinableTaskFactoryProvider
```

Following approach 3 in these docs:

- https://github.com/microsoft/vs-threading/blob/master/doc/library_with_jtf.md

I think this is "good enough" to re-review now. I am going to go over your old review comments on take one next.

ricksladkey:
Regarding issue 1. (`:set relativenumber` not taking effect):

This was an artifact of the trying to find a way to have one Visual Studio text view option get synchronized to two different vim settings. For now I came up with a simpler solution:

- The `:set number` setting is still synchronized to the Visual Studio native line number margin
- The `:set relativenumber` setting is now synchronized to the enabling the relative line number margin

With no further changes, this make both line number margins present if `:set number relativenumber` are both in effect. So I took an idea from https://github.com/brianschmitt/relativenumber (thanks @brianschmitt) to **hide** the native line number margin instead of disabling it.

New settings policy:

- `set nonumber norelativenumber` - no margins, no change from VsVim 2.6.0
- `set number norelativenumber` - native margin only, no change from VsVim 2.6.0
- `set nonumber relativenumber` - native margin hidden, new margin visible
- `set number relativenumber` - native margin hidden, new margin visible

All four look and behave exactly like gvim with the same settings. The only quirk is that `number` is synchronized with Visual Studio and `relativenumber` is not, so if you turn on `relativenumber` in `.vsvimrc` then it will affect all windows, even windows the editor doesn't think should show line numbers.

jaredpar:
@ricksladkey 

> I think this is "good enough" to re-review now. I am going to go over your old review comments on "take one" next.

I'm definitely fine with the approach here. I was more speculating if that would help with the VsVim vs. VimApp problems that you were seeing.

Src/VimApp/VimAppJoinableTaskFactoryProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    [Export(typeof(IJoinableTaskFactoryProvider))]
+    internal class VimAppJoinableTaskFactoryProvider : IJoinableTaskFactoryProvider
+    {
+        private static readonly object s_lock = new object();
+        private static JoinableTaskContext s_joinableTaskContext;
+
+        private readonly JoinableTaskFactory _joinableTaskFactory;
+
+        [ImportingConstructor]
+        VimAppJoinableTaskFactoryProvider()
```
jaredpar: `internal`

Src/VimApp/VimAppJoinableTaskFactoryProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            // See:
+            // https://github.com/microsoft/vs-threading/blob/master/doc/library_with_jtf.md
+            // https://github.com/microsoft/vs-threading/blob/master/doc/testing_vs.md
+            lock (s_lock)
```
jaredpar: Don't think the lock should be necessary here. This MEF should be a singleton so it should only be constructed once.

jaredpar: Looked through the docs you linked to and didn't see a `lock` there. Is there a scenario I'm not thinking of? 

---
In reply to: [287919513](https://github.com/VsVim/VsVim/pull/2571#discussion_r287919513) [](ancestors = 287919513)

Src/VimWpf/VimWpf.csproj
```diff
@@ -10,6 +10,8 @@
     <Reference Include="Microsoft.VisualStudio.CoreUtility, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Language.Intellisense, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Language.StandardClassification, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.Shell.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
```
jaredpar: The Shell reference shouldn't be necessary here.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Vim.UI.Wpf.Implementation.RelativeLineNumbers.Util
+{
+    public class SafeRefreshLock
```
jaredpar: Know this code is from the original PR but I can't really discern the intent of this type. It seems over complicated. Seems like we could just use a simple `SemaphoreSlim` here or similar.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (!TryObtainRedrawLock())
+            {
+                return;
+            }
+
+            await operation().ConfigureAwait(false);
+
+            if (!TryReleaseRedrawLock())
+            {
+                await ExecuteInLockAsync(operation).ConfigureAwait(false);
```
jaredpar: >ExecuteInLockAsync [](start = 22, length = 18)

This is an example: so in this case `operation` is executed maybe twice?

Src/VsVimShared/VsVimJoinableTaskFactoryProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Vim.VisualStudio
+{
+    [Export(typeof(IJoinableTaskFactoryProvider))]
+    internal class VsVimJoinableTaskFactoryProvider : IJoinableTaskFactoryProvider
+    {
+        private readonly JoinableTaskFactory _joinableTaskFactory;
+
+        [ImportingConstructor]
+        VsVimJoinableTaskFactoryProvider()
```
jaredpar: `internal`

Src/VimApp/VimAppJoinableTaskFactoryProvider.cs
ricksladkey: The docs were very insistent that if two or more `JoinableTaskFactory` objects were in use that there would be negative consequences. Although the MEF scenario is predictable, unit tests can construct the provider directly, and if tests were run in parallel we want to be certain that only one gets created.

Admittedly we don't do any of that now, but the lock seemed like an insurance policy that makes it clear that there is one and only one object created per app domain. A similar effect could be achieved by just making it static and then we would rely on static constructor locking. But that scenario creates the object at an unpredictable early time, and given the ominous warnings, this approach ensures it isn't created unless we intend to use it.

TL;DR I was being paranoid.

Src/VimWpf/VimWpf.csproj
ricksladkey: Right you are! The addition of the `VsVimJoinableTaskFactoryProvider` to the VimWpf project (I put it there by analogy with `VsVimHost` caused some reference issues, starting with compilation errors related to `ThreadHelper` which doesn't have a `JoinableTaskFactory` property in older versions of the Shell assembly. I also ran into one slightly odd error about the options dialog having to derive from `Dialog`, which it clearly did derive from.

I've removed it now.

Src/VimApp/VimAppJoinableTaskFactoryProvider.cs
ricksladkey: Right. Done.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
ricksladkey: Well, having fixed a few minor issues, I know know how to write the pure synchronous version and so we could make the async version still supported but not enabled.  I did not drill into the logic of the "safe lock" but I agree that I would rather risk a small performance hit that lingering fears that a bug lurks in there somewhere.

That being said, the code that runs in this lock is side-effect free and operates on immutable snapshots, and so I think that the worst that could happen is that it calculates two units of work in parallel. To put it another way, the lock is for preventing wasted CPU cycles, not for ensuring correctness.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
ricksladkey: Yes, that appears to be the intent that while we performed the operation (the side-effect free operation), another request was queued and that means, e.g., the text buffer we were using has a new snapshot or the view changed.

It does seem the async approach could be a useful strategy for all kinds of things, like the glyph margin or taggers (or even searching) that do **some** preparatory work on immutable data before interacting directly with the view on the UI thread. So I'm leaning toward keeping the async version because it only applies to a setting we didn't used to support, and will be used by a limited number of users, and so the consequences should be limited.

Src/VsVimShared/VsVimJoinableTaskFactoryProvider.cs
ricksladkey: Fixed.

ricksladkey:
OK, I pushed some small changes:

- Made it so that the new margin updates correctly when you set it "on the fly" in the window
- Made it so that you can swap back and forth between native and the new margin "on the fly"
- Addressed review feedback items

ricksladkey:
I've turned on relative line numbers in my daily use editor and I encountered a crash right where @jaredpar predicted there might be a crash: in `ContainingTextViewLine`. Even worse, this crash brought down all of Visual Studio!

```
   at Microsoft.VisualStudio.Text.MultiSelection.Implementation.SelectionUIProperties.get_ContainingTextViewLine()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
   at Microsoft.VisualStudio.Text.Editor.Implementation.CaretElement.get_ContainingTextViewLine()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.LineNumbersCalculator.TryGetCaretIndex(Int32& caretIndex)
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.LineNumbersCalculator.CalculateLineNumbers()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.GetNewLineNumbers()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.<<RedrawLinesAsync>g__RedrawLinesInternalAsync|19_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.Util.SafeRefreshLock.<ExecuteInLockAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.<RedrawLinesAsync>d__19.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.<OnEditorOptionsChanged>d__16.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.<<-ctor>b__14_2>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_0(Object state)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)
   at MS.Win32.HwndSubclass.SubclassWndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam)
```

ricksladkey:
OK, I fixed that particular crash because we don't need to use a text view line to get access to the point at the start of the text buffer line containing the caret. In fact, if the line is wrapped and the caret is in the wrapped part of the line, the start of that "line" is actually somewhere in the middle of the line.

In any case, I made the async code bullet-proof against exceptions that could cause a hard crash using try / catch / report with protected operations.

I will continue to test with relative line numbers in my daily use editor to make sure that it is stable.

ricksladkey:
Bug: the relative line number margin fails to update if you delete the entire contents of the buffer, e.g. `1GdG`.

ricksladkey:
Ran into another exception:

It looks like the async approach needs way more bulletproofing if we are going to use it. My guess is that although a layout was not in progress when a refresh was requested, a layout began **during** the refresh and disposed the previous text view lines.

```
  <entry>
    <record>970</record>
    <time>2019/05/29 20:20:05.775</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>
System.Exception: Unable to get new line numbers: Cannot access a disposed object.&#x000D;&#x000A;Object name: &apos;FormattedLine&apos;. ---&gt; System.ObjectDisposedException: Cannot access a disposed object.&#x000D;&#x000A;
Object name: &apos;FormattedLine&apos;.&#x000D;&#x000A;
   at Microsoft.VisualStudio.Text.Formatting.Implementation.FormattedLine.ThrowIfInvalid()&#x000D;&#x000A;
   at Microsoft.VisualStudio.Text.Formatting.Implementation.FormattedLine.get_Start()&#x000D;&#x000A;   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.LineNumbersCalculator.TryGetCaretIndex(Int32&amp; caretIndex)&#x000D;&#x000A;
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.LineNumbersCalculator.CalculateLineNumbers()&#x000D;&#x000A;
   at Vim.UI.Wpf.Implementation.RelativeLineNumbers.RelativeLineNumbersMargin.GetNewLineNumbers()&#x000D;&#x000A;
--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;
   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)&#x000D;&#x000A;
   --- End of inner exception stack trace ---
</description>
  </entry>
```

ricksladkey:
OK, more changes:

1. I made it synchronous (but kept the infrastructure for async)
2. I fixed the refresh bug that showed incorrect line numbers when deleting the whole buffer

I will continue to use this for a few days and keep an eye out for any problems.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
ricksladkey: So daily use testing showed that the whole async strategy, if effective, needs to be done more carefully. I converted the margin to be entirely synchronous.

ricksladkey: The code necessary to make it async is small, so I deleted it. It'll be in the git history if we ever need to revive it.

Src/VimWpf/Implementation/RelativeLineNumbers/Util/SafeRefreshLock.cs
ricksladkey: Deleted.

ricksladkey:
Last batch of changes just removed the conditional async support (pure sync now) and bullet-proofed all accesses to text view lines.

ricksladkey:
Thanks to @oxilumin for contributing this major and oft-requested feature!

brianschmitt:
Great job @ricksladkey & @oxilumin

Hugo-Matias:
Is this available on VS2022? I can´t seem to find the option to enable relative line numbers on the extension settings.

lfcyja:
@Hugo-Matias Works for me, just create the `~/.vimrc` file and put the desired settings in it. In my case, this was `:set number relativenumber` and lead to the following in VS 2022:
image

Hugo-Matias:
@lfcyja Oh I see! I was looking at Visual Studio's extension options...

Thanks a lot for the tip, it is working here as well!

PR #2575, opened 2019-05-27
Transition CI to Azure DevOps

### LINKED ISSUE #2557: Experimenting with Azure DevOps
Part of completing #2549 meant that there is now an observable difference between our behaviors between VS2019 and VS2017 (and earlier). This means in order to full validate our build we essentially need to run tests twice. 

This makes sense conceptually. VsVim needs to run on different platforms with different behaviors. It makes sense to validate the implementation on each of those platforms. Up until now the behavior differences have been so small that there really wasn't a big risk here. But now with async completion, and the MEF composition fun, the gap feels very real. 

I'd been thinking of moving to Azure DevOps anyways. This type of dual testing is fairly easy to setup there. Wanted to do some experiments with this in order to see how it feels for VsVim.

While I'm doing this you may see some emails, or apps, showing up in the org related to this change. There will be a few PRs as well but they will be in draft mode and / or to branches under `/dev/jaredpar`. Those are meant to be experiments on my part to see how it all feels.

If it all feels good I'll send a PR to master representing the transition *proposal* that we can use as a decision point.

jaredpar:
## Functional Change
This PR is meant to demonstrate what using Azure DevOps on VsVim will look like and what capabilities that would give us. For example this PR expands the set of testing we perform on VsVim:

1. Unit Tests run against the VS2017 and VS2019 editors
1. Validates all of our deployment criteria: version numbers match, VSIX contents correct
1. Uploads VSIX to the [Open VSIX Gallery](http://vsixgallery.com/)

Only 1a and 3 were done in AppVeyor. Even with the extra testing the validation still completes in the same amount of time: ~5 minutes. 

## Impact of Merging
Should we decide to go forward with this then it will require some fix up on PRs opened before the merge. They won't be able to run on Azure DevOps because they don't have this content in them. Hence I'll leave AppVeyor alive for those PRs until they all get merged. New PRs will just pick up the Azure DevOps change naturally. 

## Visual Tour of Changes
The PR view will now display four different jobs links when a PR is submitted

ci1

The VsVim-CI one is just the outer job. It will pass only if all of the other ones pass. The remaining three represent the checks that are running:

- Running unit tests on VS2017 editor
- Running unit tests on VS2019 editor
- Producing a Release VSIX + upload to the gallery

Clicking on the job link will actually move you to the "Checks" tab of the PR. This is meant to be a summary view of the Azure DevOps runs. To get to the actual Azure DevOps run you click on the "View more details on Azure Pipelines" link. 

ci2

Personally I often find the Azure display more useful than checks. It has the raw output which is often more useful. For example when xunit crashes the checks display will simply say "Tests exited with code 1" while Azure will display the full stack trace of the exception (which is the info you actually need). 

One useful feature of the checks page though is the ability to re-run a job. Consider for example if a flaky test failed and you just need to re-run that leg. For that case just hit the "Re-Run" link for the job in the checks page. 

ci4

From the Azure Pipelines page you can click through the jobs, see the raw output from every leg and download published artifacts. To see what tests are run, failing, etc ... click on the Tests link and it will give you the data in a nice display. 

ci3

closes #2557

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -1,0 +1,2 @@
+@echo off
+powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0Scripts\Build.ps1""" -build %*"
--- a/Documentation/Developing.md
+++ b/Documentation/Developing.md
@@ -1,0 +1,57 @@
+# Developing VsVim
+
+When developing please follow the
+[coding guidelines](https://github.com/VsVim/VsVim/blob/master/Documentation/CodingGuidelines.md)
+
+## Requirements
+VsVim can be developed using Visual Studio 2017 or 2019. The required workloads
+are:
+
+- .NET Desktop Development
+- F# Language
+- Visual Studio Extension Development
+
+## Multi Editor Support
+VsVim supports multiple versions of Visual Studio: 2015 through 2019. While the
+underlying editor components are remarkably compatible between the versions,
+there are subtle behaviors differences that do show up. Further there are 
+features, like async completion, which appear only in later versions of 
+Visual Studio. 
+
+The the unit tests are designed to test against multiple versions of Visual
+Studio. This is done at build time with a series of `#if` directives to 
+configure the unit tests to load a specific version of the VS editor
+components.
+
+By default VsVim will run unit tests against the version of Visual Studio that
+is being used to edit the source code. This can be configured though by doing
+the following:
+
+- Setting `%VsVimTargetVersion%` to 15.0 or 16.0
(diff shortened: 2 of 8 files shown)


jaredpar:
@ricksladkey yes. I actually decided to leave the appveyor.yml file checked in until all open PRs were converted over. Once they're merged or brought up to date with master then we can delete appveyor.yml. That way it would minimize disruption.

jaredpar:
@ricksladkey 

> FYI, I think all my outstanding PRs are converted over.

Great! Once I get the PR to enable VS2015 testing merged I'll move on to removing the AppVeyor support.

PR #2582, opened 2019-05-30
Fix semantics of inverted collections in vim regular expressions

### LINKED ISSUE #1471: hlsearch highlighting doesn't match search matches because of universe negation
Reproduce case:
1. Issue: `:set hlsearch`
2. Create a file as follows: Hello[Enter]World[Enter]

> Hello
> World
1. Hit escape to go into normal mode
2. Execute the following search: `/o[^o]\+W`
3. Note that VsVim reports "Pattern not found"
4. Note that `hlsearch` caused the o in hello and the W in World to be highlighted:

image

And here's a similar case:
1. With the search completed, delete all the file content
2. Re-add the content: Hello[Enter]World[Enter]
3. Note that the same text as previously shown is highlighted

In case it matters I have R# 8.2 installed. I'm using the latest VsVim at the time of reporting the issue and am running VS 2013 Premium.

ricksladkey:
#### Issues

- Closes #1471

#### Discussion

Here's a summary of how Vim collection semantics work:

| Vim Regex | BCL Regex | Vim Help Topic |
| - | - | - |
| `[a]` | `[a]` |  `/[` |
| `[^a]` | `[^\r\na]` |  `/[\n]` |
| `\_[a]` | `[\r\na]` | `/[\n]` |
| `\_[^a]` | `[^a]` | `/[\n]` |

### DIFF
--- a/Src/VimCore/VimRegex.fs
+++ b/Src/VimCore/VimRegex.fs
@@ -580,19 +580,28 @@ module VimRegexFactory =
         else
             false
 
+    /// Clear state related to previous parsing
+    let clearState (data: VimRegexBuilder) =
+        let isStartOfPattern = data.IsStartOfPattern
+        let isStartOfCollection = data.IsStartOfCollection
+        let isAlternate = data.IsAlternate
+        data.IsStartOfPattern <- false
+        data.IsStartOfCollection <- false
+        data.IsAlternate <- false
+        isStartOfPattern, isStartOfCollection, isAlternate
+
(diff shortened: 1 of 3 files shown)


Src/VimCore/VimRegex.fs
```diff
@@ -572,15 +580,29 @@ module VimRegexFactory =
         else
             false
 
+    /// Clear state related to previous parsing
+    let clearState (data: VimRegexBuilder) =
```
jaredpar: `ClearState`

PR #2584, opened 2019-05-31
Move to AsyncPackage

### LINKED ISSUE #2578: Need to use AsyncPackage
**Describe the bug**
Today VsVim uses a standard Visual Studio `Package` type for the shell integration. This needs to transition to using `AsyncPackage` per the new Visual Studio guidelines. 

https://aka.ms/asyncpackage

Presently this is a soft requirement. The tooling is just giving warnings when uploading a new VSIX. Eventually this will turn to a much harder requirement.

jaredpar:
Extensions need to move to using `AsyncPackage` over `Package` in more recent Visual Studio versions. 

closes #2578


Src/VsVim/VsVim.csproj
```diff
@@ -111,7 +113,7 @@
   <PropertyGroup>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="15.5.72" />
+    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="15.9.3039" />
```
jaredpar: The 15.5.72 version of `CreatePkgDef` had a bug when the package reference `Microsoft.VisualStudio.Threading` at version 14.0. There was an assembly binding redirect in the app.config file that forced it to 15.3. This failed as the version in the output directory of the VSIX would always be 14.0. The newer version removed this binding redirect and hence can work with 14.0 properly.

Src/VsVim/VsVimPackage.cs
```diff
@@ -38,11 +41,21 @@ public VsVimPackage()
...
+            // The cast to IComponentModel MUST occur after the call to SwitchToMainThreadAsync. The 
+            // call to GetServiceAsync can occur on any thread but the cast to the actual service 
+            // implementation must occur on the UI thread (for STA services at least). The implementation
+            // can go through QueryService if the object is a CCW and that must be done on the appropriate
+            // thread.
+            //
+            // For the services we use here they are likely to be managed objects and hence not really 
+            // applicable to this rule. But this is the best practice and should be followed in either
+            // case.
+            _componentModel = (IComponentModel)componentModel;
```
ricksladkey: It seems incredible that it is "this easy"!

jaredpar: The hard part was figuring out this was all the code I actually needed here. 😦

ricksladkey: That's what I call brain surgery!

PR #2597, opened 2019-06-04
Enable VS2015 testing

jaredpar:
This expands the set of VS versions that we can run VsVim tests against to include VS2015. Combined with our existing infrastructure for VS2017 and VS2019 all supported versions are now capable of being tested from the command line or IDE. Further this adds the VS2015 validation to our regular PR loop.


jaredpar:
Pushed to the wrong branch (jarepdar instead of jaredpar). Created new PR to replace this one.

PR #2599, opened 2019-06-05
Add GUI option to disable / enable VsVim

### LINKED ISSUE #1545: Add a UI option to disable VsVim
It is so rare that I need to disable VsVim I have to search online each time to remember Ctrl+Shift+F12 combination.  It would be convenient it there was a button to do so on the Options menu.  It would also help when pairing and so that the other person who's not a vim user can quickly switch without knowing the shortcut.

ricksladkey:
#### Issues

- Closes #1545


ricksladkey:
Here is what the new option looks like:

image

PR #2600, opened 2019-06-05
Support setting environment variables using let

### LINKED ISSUE #1355: Chaining _vsvimrc to _vimrc to *real* _vimrc
I'm running the latest (1.6.0.2). Note, my "real" _vimrc lives in ~/dotfiles/_vimrc which is a git repo. Here's the setup that I initially tried (doesn't work with vsvim):
### ~/_vsvimrc:

```
set vsvim_useeditordefaults
so ~\_vimrc
```
### ~/_vimrc:

```
"For windows computers... fix problems caused by using mklink
"and editing/saving $MYVIMRC by creating a dummy vimrc that points
"to the real one
let $MYVIMRC='~/dotfiles/_vimrc'
so $MYVIMRC
```

As a workaround, I changed _vsvimrc to skip using ~/_vimrc. Then everything works the way it should...:
### ~/_vsvimrc:

```
set vsvim_useeditordefaults
so ~\dotfiles\_vimrc
```

ricksladkey:
#### Issues

- Closes #1355

### DIFF
--- a/Test/VimCoreTest/InterpreterTest.cs
+++ b/Test/VimCoreTest/InterpreterTest.cs
@@ -1333,9 +1333,12 @@ namespace Vim.UnitTest
             [WpfFact]
             public void RHSCanBeEnvironmentVariable()
             {
+                // Note this test cannot be run in parallel due to
+                // the manipulation of global state.
                 Create("");
                 var variable = "magic";
                 var value = "xyzzy";
+                var oldValue = Environment.GetEnvironmentVariable(variable);
                 try
                 {
                     Environment.SetEnvironmentVariable(variable, value);
@@ -1344,7 +1347,7 @@ namespace Vim.UnitTest
                 }
                 finally
                 {
-                    Environment.SetEnvironmentVariable(variable, null);
+                    Environment.SetEnvironmentVariable(variable, oldValue);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


Test/VimCoreTest/InterpreterTest.cs
```diff
@@ -1470,6 +1470,23 @@ public void RHSCanBeStringNotEqualExpression()
...
+                var variable = "magic";
+                var value = "xyzzy";
+                try
+                {
+                    ParseAndRun($"let ${variable} = '{value}'");
+                    Assert.Equal(value, System.Environment.GetEnvironmentVariable(variable));
+                }
+                finally
+                {
+                    Environment.SetEnvironmentVariable(variable, null);
```
jaredpar: Really minor nit: As unlikely as it is that someone would define `%magic%` i'm sure there is someone who does it. Consider saving and restoring the original value here vs. nulling it out.

ricksladkey: Yes, I was thinking about this. Not corrupting the process environment is probably good etiquette, if not strictly necessary.

The only possible negative consequence to manipulating the environment is interfering with another test, and the only way to prevent that is to ensure that tests that manipulate the environment like this are run sequentially rather than in parallel. I believe **all** our tests run sequentially now so even restoring it is unnecessary. It would be better if we could annotate the test as non-parallelizable, but I don't know if that formality is necessary at the moment.

I'll restore the former value and make a comment about running tests in parallel.

PR #2602, opened 2019-06-05
Enable VS2015 Testing

jaredpar:
This expands the set of VS versions that we can run VsVim tests against to include VS2015. Combined with our existing infrastructure for VS2017 and VS2019 all supported versions are now capable of being tested from the command line or IDE. Further this adds the VS2015 validation to our regular PR loop.


jaredpar:
Part of this change includes a new function, `CreateTextViewDisplay`, which creates a visible WPF editor instance during unit testing. 

While debugging into the VS2015 test failures I discovered the VS Integration tests weren't actually testing the editor properly for completion scenarios. There are a number of editor services which change their behavior when `IsVisible == false` for the editor window. In some cases it will actually dismiss completion entirely. 

Generally I try and avoid visible UI strongly in unit tests as it can be a distraction when developers are running tests locally. In this case though I think it's unavoidable. I asked the editor team how they approach the situation and having a short lived `Window` which hosts the WPF editor is the approach that they take.

jaredpar:
@ricksladkey 

>  I tried using the new Build.ps1, but got script failures, which I will look into when I get a chance.

Can you paste the errors? Probably can reason out pretty quick what I did wrong.

> One thing you could fix is that we need to add 14.0 as a test target in the usage message of the build script.

Indeed. Let me fix that.

ricksladkey:
> Can you paste the errors? Probably can reason out pretty quick what I did wrong.

Not sure you did anything wrong; I might be running it wrong!

```
C:\Build\VsVim>powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -command "Set-PSDebug -Trace 2; Scripts\Build.ps1 -build"
DEBUG:    1+ Set-PSDebug -Trace 2;  >>>> Scripts\Build.ps1 -build
DEBUG:   17+  >>>> Set-StrictMode -version 2.0
DEBUG:     ! CALL function '<ScriptBlock>'  (defined in file 'C:\Build\VsVim\Scripts\Build.ps1')
DEBUG:   18+  >>>> $ErrorActionPreference="Stop"
DEBUG:     ! SET $ErrorActionPreference = 'Stop'.
DEBUG:   20+  >>>> [string]$rootDir = Split-Path -parent $MyInvocation.MyCommand.Definition
Error: Invalid array passed in, ']' expected. (1): [
at Exec-CommandCore, C:\Build\VsVim\Scripts\Common-Utils.ps1: line 53
at Exec-Command, C:\Build\VsVim\Scripts\Common-Utils.ps1: line 85
at Get-MSBuildPath, C:\Build\VsVim\Scripts\Build.ps1: line 111
at Build-Solution, C:\Build\VsVim\Scripts\Build.ps1: line 253
at <ScriptBlock>, C:\Build\VsVim\Scripts\Build.ps1: line 306
at <ScriptBlock>, <No file>: line 1

C:\Build\VsVim>
```

jaredpar:
@ricksladkey ah think I know what is going on there. Will have a PR out in just a sec.

jaredpar:
#2631 should fix it.

Guessing that you have multiple instances of VS installed on your machine. My laptop, where I do VsVim work, typically only has one installed. That probably causes the output to be past the standard buffer length and hence broken up across a couple of `Write-Ouptput` calls in `Exec-CommandCore`

ricksladkey:
> Guessing that you have multiple instances of VS installed on your machine.

Ha, ha, yes, so many versions of Visual Studio! Thanks.

PR #2603, opened 2019-06-06
Use <C-c> to cancel an external edit instead of <Esc>

### LINKED ISSUE #2587: <ESC> in insert mode in inline rename is exiting inline rename
**Describe the bug**
In the latest internal previews of 16.2 (Int Preview 28929.309), hitting escape while in inline rename to exit insert mode also exits the inline rename session.

**To Reproduce**

1. Hit inline rename on a method.
2. VsVim enters insert mode, and the inline rename dialog is brought up.
3. Hit `esc`.
4. Inline rename is cancelled, and normal mode is entered.

**Expected behavior**
Insert mode is exited, and normal mode is entered (as occurred previously). Inline rename is _not_ cancelled.

- Visual Studio version: 16.2 Int Preview 28929.309
- VsVim version: 2.7.0.0
- Programming Language: C#
    - [ ] ReSharper
    - [ ] Visual Assist

ricksladkey:
#### Issues

- Closes #2587


Src/VsVimShared/Implementation/Roslyn/RoslynListenerFactory.cs
```diff
@@ -78,23 +84,30 @@ private void OnIsRenameActiveChanged(object sender, EventArgs e)
...
+        private void OnModeChange(object sender, SwitchModeEventArgs args)
         {
-            if (_inRename && _roslynRenameUtil.IsRenameActive)
+            if (_inRename && _roslynRenameUtil.IsRenameActive && args.ModeArgument.IsCancelOperation)
             {
                 _roslynRenameUtil.Cancel();
             }
         }
 
         internal static bool IsRoslynContentType(IContentType contentType)
```
jaredpar: Roslyn is generally C# / VB ony. Now that we are tracking F# as well should probably use a diff name. Suggest: `IsTargetContentType`

Src/VimCore/CoreInterfaces.fs
```diff
@@ -2612,6 +2612,9 @@ type ModeArgument =
     /// Enter command mode with a partially entered command and then return to normal mode
     | PartialCommand of Command: string
 
+    /// Cancel any operation that is in-progress, such as refactoring
+    | CancelOperation
```
jaredpar: Generally `ModeArgument` is an option to the new mode in a switch where as `CancelOperation` is somewhat an argument to the previous mode. Think we should change the name / comment to make this clear. Maybe `CancelPreviousModeOperations`.

Src/VimCore/ExternalEdit.fs
```diff
@@ -4,15 +4,32 @@ namespace Vim
...
         member x.CommandNames = Seq.empty
-        member x.CanProcess keyInput = keyInput = KeyInputUtil.EscapeKey
+        member x.CanProcess keyInput =
+            keyInput = KeyInputUtil.EscapeKey 
+            || keyInput = KeyInputUtil.CharWithControlToKeyInput 'c'
         member x.Process keyInputData = 
-            if keyInputData.KeyInput = KeyInputUtil.EscapeKey then
-                ProcessResult.OfModeKind ModeKind.Normal
-            else
+            match keyInputData.KeyInput with
```
jaredpar: Rather than hooking mode switch did you consider just putting a more direct API on `ExternalEditMode`? The issue with hooking mode switch is that it's more general purpose and cancelling existing operations is generally implicitly true. For instance you cancel word completion when we transition insert to normal mode. Refactoring is the odd one here in that it prefers not to cancel on change and has to be told to. I've been trying to think of other features which would want to do this and can't think of any. So this seems pretty specific to external edit and wondered if you'd thought about putting the API there.

jaredpar:
Done with review pass (iteration 8)

Src/VsVimShared/Implementation/Roslyn/RoslynListenerFactory.cs
ricksladkey: But the problem then is much bigger: everything with RoslynRename in its name is named incorrectly. There are about half a dozen files and classes that will need to be renamed. Should I fix those as well? Example: RoslynRenameUtil.cs => InlineRenameUtil.cs.

Src/VimCore/ExternalEdit.fs
ricksladkey: The current separation of concerns prevents any direct communication between VimCore and ExternalEditMode. As a result, VimCore must somehow signal indirectly that a general "cancel operation" action occurred which the inline rename can then notice. That event must originate at the VimBuffer level or higher because we can press `<C-c>` in any mode. But the encapsulation is so severe that modes don't even have access to any VimBuffer APIs.

Internal actions like word completion have the luxury participating directly in the logic of the mode doing the cancelling. But external operations like inline rename can't hook `<C-c>` through any of the normal mechanisms. If a "cancel operation" event is insufficient, we would need something like an `IExternalOperation.Cancel` API.

Src/VimCore/CoreInterfaces.fs
ricksladkey: I don't think so. It is wrong to say that the operation being cancelled belongs to or is any way related to the previous mode. The previous mode is just some random mode that happened to be the current mode when the user pressed `<C-c>`. Conceptually the "cancel operation" mode argument is informing the VimBuffer of the cancellation, not the old mode or the new mode.

Src/VimCore/ExternalEdit.fs
ricksladkey: Here are some discussion comments about the new architecture that occurred in the issue this PR is intended to close:

- https://github.com/VsVim/VsVim/issues/2587#issuecomment-499340066
- https://github.com/VsVim/VsVim/issues/2587#issuecomment-500098665

The important point is that `ExternalEditMode` can be completely bypassed, but we **still** need a way to cancel refactoring without resorting to the mouse.

Src/VsVimShared/Implementation/Roslyn/RoslynListenerFactory.cs
ricksladkey: I renamed all of "RoslynRename" to "InlineRename" as that is what the class is called within Visual Studio.

Src/VimCore/ExternalEdit.fs
ricksladkey: Upon reflection, I think a "cancel operation" as a mode argument is reasonable. It means "upon entering the new mode, any in-progress host operation that wishes to participate in the "cancel operation" mechanism should now cancel that operation.

PR #2614, opened 2019-06-08
Support window settings in modeline

### LINKED ISSUE #2613: Window settings specified in the modeline are not applied
If a vim modeline specifies a window setting, such as `:set nowrap`, that setting is not applied when loading the text buffer into a new window.

ricksladkey:
#### Issues

- Closes #2613

### DIFF
--- a/Src/VimCore/ModeLineInterpreter.fs
+++ b/Src/VimCore/ModeLineInterpreter.fs
@@ -94,6 +94,7 @@ type internal ModeLineInterpreter
                 _insecureWindowSettingNames
                 |> Seq.contains windowSetting.Value.Name
                 |> not
+
             else
                 false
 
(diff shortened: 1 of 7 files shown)


Src/VimCore/ModeLineInterpreter.fs
```diff
@@ -62,19 +72,45 @@ type internal ModeLineInterpreter
...
+
+            else
+                false
+
+        // Set the setting if applicable
+        let setSetting settingName settingValue =
+            if _localSettings.GetSetting settingName |> Option.isSome then
+
+                // Only set local settings the first time the buffer is
+                // checked.
```
jaredpar: What is the expected behavior when multiple `ITextView` are opened on a single `ITextBuffer`. Presently there is a 1:1 relationship between `ModeLineInterpretter` and `ITextView` meaning if we open multiple views we will interpret the mode line mulitple times. My reading of the comment indicates the opposite is meant to be true.

ricksladkey: No, there is a one-to-one relationship between `ModeLineInterpreter` and `IVimTextBuffer` which has a one-to-one relationship with `ITextBuffer`. This is what the member function accepts window settings as an argument. The window settings apply to the text view, the local settings apply to the text buffer.

ricksladkey: But your point is well taken that it would be possible to use the other approach. Conceptually a text buffer can exist without a text view but a text view cannot exist without a text buffer so I chose to load loading settings from the buffer.

ricksladkey: In any case, there is a detailed unit test that verifies the correct behavior.

PR #2616, opened 2019-06-09
Fix text alignment on block caret overlay

### LINKED ISSUE #2615: Block caret overlay misaligned with underlying text
The block caret actually draws the character underneath the caret onto an opaque text adornment and so any misalignment between the text rendering of the text view and the text rendering of the block caret is manifested as a slight jiggle in the character's position as the caret rolls over it or as the caret blinks.

Zooming the document window to 400% makes this easier to see:

image

Previously this alignment was nearly perfect but as of release 16.1.2 or so of Visual Studio 2019 (and a similar recent release of Visual Studio 2017), the jiggle is back.

While we are on the subject of the block caret, the block caret rendering does not work correctly when the caret is positioned over a character that is rendered as vim-style control character, e.g. `^C` for ASCII 3. In that case, the caret appears mostly solid depending on the text colors chosen for the caret and for control characters.

image

ricksladkey:
#### Issues

- Closes #2615

### DIFF
--- a/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
+++ b/Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
@@ -387,7 +387,8 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
                 var line = TextViewLineContainingCaret;
                 height = line.Height;
                 var point = _textView.Caret.Position.BufferPosition;
-                if (point.Position < _textView.TextSnapshot.Length)
+                var codePointInfo = new SnapshotCodePoint(point).CodePointInfo;
+                if (point.Position < point.Snapshot.Length)
                 {
                     var pointCharacter = point.GetChar();
                     if (_controlCharUtil.TryGetDisplayText(pointCharacter, out string caretString))
@@ -396,9 +397,7 @@ namespace Vim.UI.Wpf.Implementation.BlockCaret
                         caretCharacter = caretString;
                         width = line.GetCharacterBounds(point).Width;
                     }
-                    else if (Char.IsHighSurrogate(pointCharacter) &&
-                        point.Position < _textView.TextSnapshot.Length - 1 &&
-                        Char.IsLowSurrogate(point.Add(1).GetChar()))
+                    else if (codePointInfo == CodePointInfo.SurrogatePairHighCharacter)
                     {
                         // Handle surrogate pairs.
                         caretCharacter = new SnapshotSpan(point, 2).GetText();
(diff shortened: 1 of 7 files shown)


ricksladkey:
Here is what the new caret alignment looks like:

image

and here is what the new control character display looks like:

image

ricksladkey:
I don't edit files with tabs in them very often, but when I do, I wish the block caret floated over the last space of the tab like it does in vim. So I've rolled that feature into this PR:

image

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -338,36 +376,45 @@ private FormattedText CreateFormattedText()
...
-                        && !_controlCharUtil.IsDisplayControlChar(pointCharacter))
+                    if (_controlCharUtil.TryGetDisplayText(pointCharacter, out string caretString))
+                    {
+                        // Handle control character notation.
+                        caretCharacter = caretString;
+                        width = line.GetCharacterBounds(point).Width;
+                    }
+                    else if (Char.IsHighSurrogate(pointCharacter) &&
+                        point.Position < _textView.TextSnapshot.Length - 1 &&
+                        Char.IsLowSurrogate(point.Add(1).GetChar()))
```
jaredpar: Consider using `point.Snapshot` vs `_textVIew.Snapshot`. Makes the code easier to read (don't have to verify they're the same snapshot.

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -338,36 +376,45 @@ private FormattedText CreateFormattedText()
...
                     var pointCharacter = point.GetChar();
-                    if (pointCharacter != '\t'
-                        && !_controlCharUtil.IsDisplayControlChar(pointCharacter))
+                    if (_controlCharUtil.TryGetDisplayText(pointCharacter, out string caretString))
+                    {
+                        // Handle control character notation.
+                        caretCharacter = caretString;
+                        width = line.GetCharacterBounds(point).Width;
+                    }
+                    else if (Char.IsHighSurrogate(pointCharacter) &&
```
jaredpar: Think we should add a helper to `SnapshotPoinUtil` for this. Pretty sure we've written this code in a few places.

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: Excellent idea.

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: Converted it to use `SnapshotCodePoint.CodePointInfo`.

PR #2620, opened 2019-06-10
Fix Problem with Completion (C-n) in VS2017  (#2619)

### LINKED ISSUE #2619: Problem with Completion (C-n) in VS2017: Picking completion with space
**Describe the bug**
After starting a vim completion (C-n or C-p), choosing a completion with space results in an Exception and completion is not chosen. (VS 2017)

**To Reproduce**

1.  open a file (existing with some text)
2.  enter Insert mode, and type beginning of some word
3.  start completion with C-n
4.  Press space to choose completion

- Visual Studio version: 2017
- VsVim version: 2.7
    - [ ] ReSharper
    - [ ] Visual Assist

**Additional context**
From logs:
System.NotImplementedException: The method or operation is not implemented.&#x000D;&#x000A;   at Vim.VisualStudio.Specific.Implementation.WordCompletion.Legacy.WordLegacyCompletionSession.Vim.IWordCompletionSession.Commit()&#x000D;&#x000A;   at Vim.Modes.Insert.InsertMode.ProcessWithWordCompletionSession(IWordCompletionSession wordCompletionSession, KeyInput keyInput)&#x000D;&#x000A;   at

I have diagnosed the problem (it is in the LegacyCompletion for 2015/2017). Will try to submit a PR.

sqlkoala:
- Closes #2619

Only small changes. Will provide some details as comments/discussion.


ricksladkey:
> I wanted to see if I can add a test, but appVeyor skipped it:

Merge VsVim's master into sqlkoala:fix-2619 and then push your branch. That way you will pick up the necessary commits from master than enable the new CI testing. Compare the "Show all checks" from  another PR to see the difference and what checks will run.

jaredpar:
> Merge VsVim's master into sqlkoala:fix-2619 and then push your branch. That way you will pick up the necessary commits from master than enable the new CI testing.

Agree. That branch looks like it's 150+ commits behind VsVim/master which is why it doesn't have the new Azure Pipelines YML file.

sqlkoala:
I think I got it. Here is a build that failed in the expected way (new test):
https://ci.appveyor.com/project/jaredpar/vsvim/builds/25200307
`
    Vim.UnitTest.InsertModeIntegrationTest+MiscTest.WordCompletion_Legacy_Commit_Space [FAIL]
      Assert.Equal() Failure
                     (pos 4)
      Expected: cat dog
      Actual:   cat  dog
                     (pos 4)
`
With the latest commit (testing for "cat  dog", two spaces), the test passes.

jaredpar:
Not sure why the Azure tests haven't triggered. Going to close / re-open the PR to see if it will trigger properly.

jaredpar:
Finally figured out why the Azure Pipelines weren't running for this pull request. I had forgotten to enable "building from forks" for the pipeline. It was working fine for myself and @ricksladkey because the branch being merged was already in the VsVim repository. This pull request was coming from a fork hence it didn't trigger. 

Should all be working now. Sorry for the noise on the pull request to debug this through.

sqlkoala:
image

I will retaliate with lots of single line commits ;-)

sqlkoala:
@jaredpar Could you help me out with the async test? My copy+paste+adjust test does not work the way I expected. The completion is not executed, so at the end it compares "cat   dog" to "c dog".

Src/VimCore/Modes_Insert_InsertMode.fs
```diff
@@ -17,7 +17,10 @@ type WordCompletionUtil
     ) =
 
     let _globalSettings = _vim.GlobalSettings
-    static let _commitKeyInput = [ KeyInputUtil.EnterKey; KeyInputUtil.TabKey; KeyInputUtil.CharToKeyInput(' ') ]
+    /// here we list keys that should commit a completionSession and be consumed
+    /// (= not beeing entered in the file)
+    /// therefore Enter and Tab, but not Space
+    static let _commitKeyInput = [ KeyInputUtil.EnterKey; KeyInputUtil.TabKey ]
```
jaredpar: The crux of the problem here is that legac and async completion are different on subtle aspects like this: whether space does or does not commit the session. For async it does not and hence we need to take special steps to make that happen but for legacy it does. 

There are approaches we can take here. Probably the best is to provide an `IAsynCompletionCommitManager` implementation that lists space as a commit character. That is non-trivial though because I believe then we also have to implement `TryCommit` which is a chunk of work. 

Maybe a simpler approach is let the `IWordCompletionSession` list the set of keys for which `Commit` should be called. Essentially move this up to the legac / async layer. The legacy layer could then omit space and let it just go through normal processing while async could list it.

sqlkoala: Thanks @jaredpar for the analysis and suggestions. So the problem is deeper then I saw it. First approach is out of my reach. I can try the second.
(I might open another PR to run a few test builds there, please ignore it)

sqlkoala: Ok, now I understand the changing behavior from legacy to async. With the old completion (active), you could type a "," " " oder "."  to pick the current entry and insert the char after it. (this is from a user perspective, not the code path).

Now, with async, you have to commit/pick an entry with one of the three keys (enter, tab, space) and this key is consumed. Now the "tradeoff" with space in this list is this:
VS 2019 users can pick an entry with space, while VS 2017 users loose the previous behavior of pressing space only once to pick an entry and insert a space (x). Pfff.
I would want to submit this PR with just the one line bugfix and tests. We can see if (x) is worth more effort. Thanks for your help and explanations.

jaredpar:
Thanks for contributing this!

PR #2628, opened 2019-06-12
Address focus problems

### LINKED ISSUE #2617: Incremental search as a motion from visual mode misbehaves
If you have a text buffer that contains:

```
cat
dog
```

and then start visual mode and search for `doABCEF`, upon typing the `A` the command margin is cleared and upon typing further characters, the search string starts to get mangled with insertions occurring at the beginning separated by slashes:

```
/F/E/D/C/B
```

### LINKED ISSUE #2625: Command margin focus problems
Switching from a window in command mode to another document window (e.g. with <kbd>Ctrl</kbd>+<kbd>Tab</kbd>) and then switching back does not restore the focus to the command margin. Similarly, when switching from Visual Studio to another application and then back to Visual Studio, if the active document window is in command mode the command margin is not focused.

### LINKED ISSUE #2627: Strange intermittent hybrid focus state
When:

- Opening a solution with saved open documents
- Opening a new document window from solution explorer
- Going to a definition in a file that hasn't been opened yet
- Stepping with the debugger

occasionally the document window appears to have the focus, but is in a strange state:

- The block caret is not visible
- The document window's tab title is not highighted
- Certain normal mode keystrokes appear to "work", such as arrow keys and the enter key
- Most other keystrokes appear to be ignored

This problem can be manually fixed (worked around) by:

- Clicking into the window with the mouse
- Switching to another document window and switching back
- Switching to another application and switching back
- Opening a dialog such as the options dialog and then closing it
- Opening a tool window such as the output window and then hiding it

At which point the block caret appears and things work normally.

Although this problem occurs intermittently and is seemingly minor, it is infuriating to type keystrokes only to realize they are not taking effect.

I instrumented the focus and with some additional tracing and reproduced the problem in the VsVim output window:

```
VsVim Debug VsCommandTarget::QueryStatus UserInput Key.Up: 1496a755-94de-11d0-8c3f-00c04fc2aae2 11
VsVim Debug VsCommandTarget::QueryStatus Enable
VsVim VsCommandTarget::Exec UserInput Key.Up: 1496a755-94de-11d0-8c3f-00c04fc2aae2 11
VsVim VimBuffer.Process: none:Up:None
VsVim Error hasAggregateFocus = False, isFocused = False, isKeyboardFocused = False, isKeyboardFocusWithin = False
VsVim Error appLogicalFocus = <none>
VsVim Error textViewLogicalFocus = <none>
VsVim Error keyboardFocus = GenericPaneContentPresenter
VsVim Debug VsCommandTarget::QueryStatus UserInput Key.Up: 1496a755-94de-11d0-8c3f-00c04fc2aae2 11
VsVim Debug VsCommandTarget::QueryStatus Enable
VsVim VsCommandTarget::Exec UserInput Key.Up: 1496a755-94de-11d0-8c3f-00c04fc2aae2 11
VsVim VimBuffer.Process: none:Up:None
VsVim Error hasAggregateFocus = False, isFocused = False, isKeyboardFocused = False, isKeyboardFocusWithin = False
VsVim Error appLogicalFocus = <none>
```

So you can see that Visual Studio is at least routing the arrow keys to the window in spite of the fact that the text view does not have aggregate focus.

After investigating this, I really doubt that this is a VsVim bug, but since VsVim is affected by it, it would be nice if there were a less clumsy workaround.

ricksladkey:
#### Issues

- Closes #2625
- Closes #2617
- Closes #2627

### DIFF
--- a/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
+++ b/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
@@ -130,6 +130,18 @@ namespace Vim.VisualStudio.Implementation.Misc
         /// </summary>
         private bool TryProcessWithBuffer(KeyInput keyInput)
         {
+            // If, for some reason, keyboard input is being routed to the text
+            // view but it isn't focused, focus it here.
+            void CheckFocus()
+            {
+                if (_vimBuffer.TextView is IWpfTextView wpfTextView &&
+                    !wpfTextView.VisualElement.IsFocused)
+                {
+                    VimTrace.TraceError("forcing focus back to text view");
+                    wpfTextView.VisualElement.Focus();
+                }
+            }
+
             // If the IVimBuffer can't process it then it doesn't matter
             if (!_vimBuffer.CanProcess(keyInput))
             {
@@ -213,17 +225,6 @@ namespace Vim.VisualStudio.Implementation.Misc
             }
 
             return handled;
-        }
-
-        private void CheckFocus()
-        {
-            // If, for some reason, keyboard input is being routed to the
-            // text view but it isn't focused, focus it here.
-            if (_vimBuffer.TextView is IWpfTextView wpfTextView && !wpfTextView.VisualElement.IsFocused)
-            {
-                VimTrace.TraceError("forcing focus back to text view");
-                wpfTextView.VisualElement.Focus();
-            }
         }
 
         /// <summary>
--- a/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
+++ b/Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
@@ -174,6 +174,7 @@ namespace Vim.VisualStudio.Implementation.Misc
             // experience for those features
             if (!_broker.IsAnyDisplayActive())
             {
+                CheckFocus();
                 return _vimBuffer.Process(keyInput).IsAnyHandled;
             }
 
@@ -196,6 +197,7 @@ namespace Vim.VisualStudio.Implementation.Misc
             }
             else
             {
+                CheckFocus();
                 // Intentionally using keyInput here and not mapped.  Process will do mapping on the
                 // provided input hence we should be using the original keyInput here not mapped
                 handled = _vimBuffer.Process(keyInput).IsAnyHandled;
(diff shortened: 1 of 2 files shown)


Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
```diff
@@ -213,6 +215,17 @@ private bool TryProcessWithBuffer(KeyInput keyInput)
             return handled;
         }
 
+        private void CheckFocus()
+        {
+            // If, for some reason, keyboard input is being routed to the
+            // text view but it isn't focused, focus it here.
+            if (_vimBuffer.TextView is IWpfTextView wpfTextView && !wpfTextView.VisualElement.IsFocused)
+            {
+                VimTrace.TraceError("forcing focus back to text view");
```
jaredpar: Why would this be an error? Is there something we are doing wrong which is changing the focus?

Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
```diff
@@ -213,6 +215,17 @@ private bool TryProcessWithBuffer(KeyInput keyInput)
             return handled;
         }
 
+        private void CheckFocus()
```
jaredpar: Could be a local function

Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
ricksladkey: It's not our error, but it is an error: the text view should not be receiving keystrokes unless it has the keyboard focus. But from a practical point of view, errors always get logged to the output window, so here `TraceError` means "trace important".

Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
ricksladkey: So it could. I'l make that change.

Src/VsVimShared/Implementation/Misc/StandardCommandTarget.cs
ricksladkey: Ha, ha, I don't know if nobody else notices these things besides me, they just ignore them, or if the problem is exclusive to my particular machine. I've noticed it for a long time, and it still occurs in both VS2017 and VS2019. I was suspicious that VsVim was the culprit, but after spending some time looking into it, I'm pretty sure it is a Visual Studio problem.

More background here: https://github.com/VsVim/VsVim/issues/2627#issue-454987003

PR #2631, opened 2019-06-12
Force full string before parsing JSON

jaredpar:
The return of `Exec-Command` can be a collection of `string` entries.
Need to use `Out-String` to force it into a single `string` before
attempting to parse out the JSON.


ricksladkey:
Merging this so I can try it out on my machine.

PR #2633, opened 2019-06-12
Sqlkoala test ignore

sqlkoala:


ricksladkey:
FYI, I recently learned of the "draft PR" feature from @jaredpar, which is a good mechanism to use in this situation:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/

jaredpar:
Yeah "Draft PRs" are great for this type of experimentation.  A Draft PR will go through all of the same checks as a normal PR but it doesn't send notifications to anyone who is watching the repository. It's also a positive indicator that "this PR isn't ready yet" to anyone who happens to stumble onto it.

PR #2638, opened 2019-06-13
Implement ':crewind' line command

### LINKED ISSUE #2637: Implement :cr[ewind]
**Feature category**
- Check(Type 'x') any that Applicable item:
    - [x] Vim
    - [ ] Vim plugin

If the function you request is realized by Vim or Vim plugin,
When you link the URL that is related,
We will make it easier to understand its function.

**Is your feature request related to a problem? Please describe.**
When navigating the list of errors with `:cn` and `:cp`, I want to be able to jump to the first error with the keyboard. In Vim, this would be done with the `:cr[ewind]` command.

**Describe the solution you'd like**
Implement `:cr[ewind]`

**Describe alternatives you've considered**
Mapping my own ex command, but that requires upper case which is annoying

ricksladkey:
#### Issues

- Closes #2637

### DIFF
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -683,19 +683,60 @@ namespace Vim.VisualStudio
 
         public override bool GoToQuickFix(QuickFix quickFix, int count, bool hasBang)
         {
-            // This implementation could be much more riguorous but for next a simple navigation
-            // of the next and previous error will suffice
-            var command = quickFix.IsNext
-                ? "View.NextError"
-                : "View.PreviousError";
-            for (var i = 0; i < count; i++)
+            var result = true;
+
+            switch (quickFix)
             {
-                SafeExecuteCommand(null, command);
+                case var value when value == QuickFix.Next:
+                    for (var i = 0; result && i < count; i++)
+                    {
+                        result = SafeExecuteCommand(null, "View.NextError");
+                    }
+                    break;
+                case var value when value == QuickFix.Previous:
+                    for (var i = 0; result && i < count; i++)
+                    {
+                        result = SafeExecuteCommand(null, "View.PreviousError");
+                    }
+                    break;
+                case var value when value == QuickFix.Number:
+                    result = GoToQuickFix(count - 1);
+                    break;
+                default:
+                    Contract.Assert(false);
+                    result = false;
+                    break;
             }
 
-            return true;
+            return result;
         }
 
+        private bool GoToQuickFix(int index)
+        {
+            try
+            {
+                if (_dte is DTE2 dte2)
+                {
+                    if (dte2.ToolWindows.ErrorList is IErrorList errorList)
+                    {
+                        var tableControl = errorList.TableControl;
+                        var entries = tableControl.Entries.ToArray();
+                        if (index < entries.Length)
+                        {
+                            var desiredEntry = entries[index];
+                            tableControl.SelectedEntries = new[] { desiredEntry };
+                            return desiredEntry.NavigateTo(false);
+                        }
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _protectedOperations.Report(ex);
+            }
+            return false;
+
+        }
         public override void Make(bool jumpToFirstError, string arguments)
         {
             SafeExecuteCommand(null, "Build.BuildSolution");
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -4677,6 +4677,7 @@ type IVimData =
 type QuickFix =
     | Next
     | Previous
+    | Number
 
 type TextViewChangedEventArgs
     (
--- a/Src/VimCore/Interpreter_Expression.fs
+++ b/Src/VimCore/Interpreter_Expression.fs
@@ -671,6 +671,9 @@ and [<RequireQualifiedAccess>] LineCommand =
     /// Previous error in the quick fix list.  int is for count and bool is for the bang option
     | QuickFixPrevious of Count: int option * HasBang: bool
 
+    /// Go to error in the quick fix list.  int is for number and bool is for the bang option
+    | QuickFixRewind of Number: int option * HasBang: bool
+
     /// Quit the curren window without writing it's content.  If the boolean option
     /// is present (for !) then don't warn about a dirty window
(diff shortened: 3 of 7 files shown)


ricksladkey:
Wow, that escalated quickly.

Src/VimCore/CommonOperations.fs
```diff
@@ -160,6 +160,18 @@ type internal CommonOperations
         with get() = _maintainCaretColumn
         and set value = _maintainCaretColumn <- value
 
+    /// Get the common operations for the specified text view
+    member x.GetCommonOperationsForTextView textView =
+        _vim.GetOrCreateVimBuffer textView
```
jaredpar: Calling `GetOrCreateVimBuffer` directly can cause `IVimBuffer` to be created for `ITextView` that the host doesn't want them created for. Consider the powershell or immediate window as an example. The host has to make a decision as to whether or not the creation should take place. Should use `TryGetOrCreateVimBufferForHost` here instead. That will complicate this API a bit but think it's necessary.

Src/VimCore/CoreTypes.fs
```diff
@@ -52,6 +52,27 @@ type JoinKind =
     | RemoveEmptySpaces
     | KeepEmptySpaces
 
+type NavigationKind =
+    | First = 1
```
jaredpar: Numbering should start at 0 here to follow the .NET Design Guidelines [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum)

Src/VimCore/CoreTypes.fs
```diff
@@ -52,6 +52,27 @@ type JoinKind =
...
     | KeepEmptySpaces
 
+type NavigationKind =
+    | First = 1
+    | Last = 2
+    | Next = 3
+    | Previous = 4
+
+type ListKind =
+    | Error = 1
```
jaredpar: Numbering should start at 0 here to follow the .NET Design Guidelines [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum)

Src/VimCore/FSharpUtil.fs
```diff
@@ -580,6 +580,7 @@ module CharUtil =
     let IsLower x = System.Char.IsLower(x)
     let IsLowerLetter x = IsLower x && IsLetter x
     let IsLetterOrDigit x = System.Char.IsLetterOrDigit(x)
+    let IsIdent x = IsLetterOrDigit x || x = '_'
```
jaredpar: This should actually be the content of `isident` setting vs. a hard coded list. Can you file a bug to track implementing this and moving this function over? It should actually be pretty easy once I complete the `iskeyword` feature. The `isindent` option is simple to build on top of that support.

Src/VsVimShared/VsVimHost.cs
```diff
@@ -456,6 +456,58 @@ public override void EndBulkOperation()
...
+                    find.Target = vsFindTarget.vsFindTargetSolution;
+                    find.MatchCase = matchCase;
+                    find.MatchWholeWord = false;
+                    find.PatternSyntax = vsFindPatternSyntax.vsFindPatternSyntaxRegExpr;
+                    find.FilesOfType = filesOfType;
+                    find.ResultsLocation = vsFindResultsLocation.vsFindResults1;
+                    find.WaitForFindToComplete = false;
+                    if (jumpToFirst)
+                    {
+                        _dte.Events.FindEvents.FindDone += onFindDone;
```
jaredpar: In case you haven't run into this problem before, the various event types in DTE can have non-intuitive garbage collection behavior. Because they're created on access, and not stored as a reference in the containing type, they can get garbage collected before the events are raised. Example is `SolutionEvents`

[https://docs.microsoft.com/en-us/dotnet/api/envdte.solutionevents?view=visualstudiosdk-2017#remarks](https://docs.microsoft.com/en-us/dotnet/api/envdte.solutionevents?view=visualstudiosdk-2017#remarks)

As far as I can tell `FindEvents` doesn't suffer from this problem. At the same time this could just be an omission from the documentation. If this is working for you though then likely the documentation is correct here.

jaredpar:
Done with review pass (iteration 18)

Src/VsVimShared/VsVimHost.cs
ricksladkey: Ha, ha, I **did** run into that problem: the 'on find done' callback only worked about 50% of the time. I looked into it without solving it and eventually gave up. I attributed the problem to the fact that `FindDone` was marked "Microsoft Internal Use Only". For that matter, so are all the `SolutionEvents`. I would have eventually figured it out if the `FindDone` documentation contained a similar remarks section.

In any case, I am very ambivalent about the "jump to first" feature. I am not sure I will ever want to use it, and yet it is the default. Let me get it working reliably and I will test it out. Thanks for solving the mystery because it was bugging me.

Src/VimCore/CommonOperations.fs
ricksladkey: Ah, that makes sense.

Src/VimCore/CoreTypes.fs
ricksladkey: I initially did not have the numeric assignments and, as a result, the type was a discriminated union instead of an enum. That interoperates poorly with C# and switch, where I ended up with code like this:

```c#
    switch (navigationKind)
    {
        case value when value == NavigationKind.First:
            ...
            break;
    }
```

I eventually figured out than an enum would work just as well in F# and work better in C# (i.e. we can use normal case labels).

Looking at the existing types, there are a couple of F# enums (`Direction`, `WindowKind` and `ModeKind`) that also need to start at zero. I'll fix them at the same time.

Src/VimCore/FSharpUtil.fs
ricksladkey: Great idea to file as issue so I don't forget.

Src/VimCore/CoreTypes.fs
ricksladkey: Fixed.

Src/VimCore/FSharpUtil.fs
ricksladkey: Opened #2645.

PR #2651, opened 2019-06-19
Support multiple carets

### LINKED ISSUE #2362: Interact with multi-caret selections

### LINKED ISSUE #2656: Integrated multiple caret support
The popularity of multiple cursors appears to have started with (or perhaps gained traction in) the Sublime editor:

- https://www.sublimetext.com/docs/3/multiple_selection_with_the_keyboard.html

Visual Studio Code also provides multiple cursors very much in the style of Sublime:

- https://code.visualstudio.com/docs/editor/codebasics#_multiple-selections-multicursor

Later versions of Visual Studio 2017 and all versions of Visual Studio 2019 support multiple carets and multiple disjoint selections. Visual Studio itself has some limited support for using multiple carets to edit files, specifically the <kbd>Shift</kbd>+<kbd>Alt</kbd>+<kbd>.</kbd> command which adds a new selection at the next occurrence of the current word. In native selection mode, this allows you to type over a new name for multiple separate identifiers at the same time.

Although Vim 8.0 does not have built-in support for multiple cursors, there are many plugins that provide varying versions of this capability, most notably:

- https://github.com/terryma/vim-multiple-cursors

The interesting thing about multiple cursors is that, except for the process of creating them, using them is very intuitive. Any casual Vim user could guess that if you can see multiple cursors and type in a normal mode command, the same command should apply at each of the cursors. Likewise, in insert mode, if you have multiple cursors, the user intuitively expects what is typed to be inserted at each of the cursors.

Some people argue that multiple cursors are unnecessary in Vim because Vim provides a variety of powerful ways to rapidly act on isolated sections of text, most notably but not limited to the repeat command. However, upon reflection, there is one powerful argument for multiple cursors in Vim:

- Any time that the best way to accomplish a text transformation in several places would be to record and use a temporary macro, that same task could be probably be performed more easily and intuitively with multiple cursors

As any one who has ever used Vim macros can tell you, it can be hard to execute the key sequence perfectly during recording, often resulting in several failed attempts before the perfect version that is repeatable so that it achieves the desired effect. And the longer the macro, the more work that is involved and the more likely there is to be a mistake. But, if you could **see** the "macro" playing out in real-time in the other places that you intend to apply the macro, you could bypass this two step process. This rationale does not apply to general-purpose non-temporary macros. For those cases, a macro is the preferred and tried-and-true method.

Whether you think Vim should or should not support multiple cursors, it must be conceded that multiple cursors are wildly popular and quickly becoming a "de rigueur" feature for a modern editor. Even if your editor supports them, you don't have to use them if you don't want to. However, if you do want to use them, your editor has to support them.

N.B.: Vim uses the term "cursor" for the insertion point in the editor, whereas Windows traditionally reserves the word "cursor" for the mouse and uses "caret" for the insertion point. In the VsVim code base the word "caret" is used systematically because all the editor APIs use that terminology. For most purposes, "cursor" and "caret" can be used interchangeably without confusion.

ricksladkey:
#### Issues

- Closes #2362
- Closes #2656

#### Related Pull Requests

- PR #2411

#### Keyboard Bindings

- `<C-A-LeftMouse>` - add or remove caret (normal, visual, select, insert)
- `<C-A-2-LeftMouse>` - add a new selected word or token (normal, visual, select, insert)
- `<C-A-Up>` and `<C-A-Down>` - add a new caret on an adjecent line (normal, insert)
- `<C-A-Up>` and `<C-A-Down>` - add a new selection on an adjecent line (visual, select)
- `<C-A-I>` - split selection into carets (visual, select)
- `<C-A-N>` - select word or token at caret (normal)
- `<C-A-N>` - add next occurrence of primary selection (visual, select)
- `<C-A-P>` - restore previous multi-carets or multi-selections (normal)
- `<Esc>` - clear secondary carets (normal)
- `<C-c>` - clear secondary carets or selections (normal, visual, insert)

#### Discussion

What I've got working so far:

- Adding a new caret with `<C-A-LeftMouse>` (ctrl-alt-click)
- Displaying multiple simultaneous block or selection carets properly
- Using different caret colors for the primary and secondary carets
- Moving the relative positions of the carets in normal mode as you move around
- Using insert mode at all carets simultaneously
- Performing operator commands for all carets, e.g. `cw` and `dw`
- Managing undo transactions so that entering insert mode after a change works
- Adding carets on adjacent lines using `<C-A-Up>` and `<C-A-Down>` (control-alt-arrow)
- Reverting to a single caret using `<C-c>` (control c)
- Using `<C-A-2-LeftMouse>` (ctrl-alt-double-click) to add another word to the selection
- Typing in select mode deletes all selections and enters insert mode with multiple carets
- Each caret gets its own unnamed register
- Entering and using visual character and select character modes with multiple carets


ricksladkey:
Here are some (tasteful) multiple block carets in action:

image

after change word (`cw`):

image

after typing `assem`:

image

after typing `<C-u>` and then `u`:

image

ricksladkey:
I tried to support VS2017, but the reference assemblies are too old and in the course of trying to upgrade the assemblies I ended up in a twisty maze of passages all alike.

```
+		[0]	{"Could not load type 'Microsoft.VisualStudio.Text.UI.Commanding.ICommandHandlerService' from assembly 'Microsoft.VisualStudio.Text.UI, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.":"Microsoft.VisualStudio.Text.UI.Commanding.ICommandHandlerService"}	System.Exception {System.TypeLoadException}
+		[1]	{"Could not load type 'Microsoft.VisualStudio.Text.UI.Commanding.ICommandHandlerServiceFactory' from assembly 'Microsoft.VisualStudio.Text.UI, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.":"Microsoft.VisualStudio.Text.UI.Commanding.ICommandHandlerServiceFactory"}	System.Exception {System.TypeLoadException}
+		[2]	{"Method 'CallExtensionPoint' in type 'Microsoft.VisualStudio.Text.Utilities.GuardedOperations' from assembly 'Microsoft.VisualStudio.Platform.VSEditor, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' does not have an implementation.":"Microsoft.VisualStudio.Text.Utilities.GuardedOperations"}	System.Exception {System.TypeLoadException}
+		[3]	{"Method 'DuplicateSelection' in type 'Microsoft.VisualStudio.Text.Operations.Implementation.EditorOperations' from assembly 'Microsoft.VisualStudio.Platform.VSEditor, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' does not have an implementation.":"Microsoft.VisualStudio.Text.Operations.Implementation.EditorOperations"}	System.Exception {System.TypeLoadException}
+		[4]	{"Could not load type 'Microsoft.VisualStudio.Text.UI.Commanding.Commands.CommandArgs' from assembly 'Microsoft.VisualStudio.Text.UI, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.":"Microsoft.VisualStudio.Text.UI.Commanding.Commands.CommandArgs"}	System.Exception {System.TypeLoadException}
```

For the time being this feature will be VS2019 only even though some versions of VS2017 support multiple carets.

ricksladkey:
I finished my list of basic functionality. Surely there are bugs and limitations but the big picture is that this should be roughly feature parity with vim-multiple-cursors and VSCodeVim multiple cursor support. All the new command bindings use "ctrl+alt" which doesn't (and is unlikely to) conflict with any standard Vim binding.

VsVim's key mapping support should allow anyone to map these awkward keys into more comfortable locations. As an example, you can get close to the finger feel of vim-multiple-cursors with something like:

```
nnoremap <C-n> <C-A-n>
vnoremap <C-n> <C-A-n>
inoremap <C-n> <C-A-n>
```

ricksladkey:
The last thing on my list that is not finished in VS2017 support, which requires updating most of the VS2017 reference assemblies. I tried this once and failed miserably. I may be convinced to try again, or I might need some help from @jaredpar!

jaredpar:
> The last thing on my list that is not finished is VS2017 support, which requires updating most of the VS2017 reference assemblies

Is this because the multiple cursor support only exists in the later versions of VS2017?

ricksladkey:
> > The last thing on my list that is not finished is VS2017 support, which requires updating most of the VS2017 reference assemblies
> 
> Is this because the multiple cursor support only exists in the later versions of VS2017?

Yes, according to @EnCey, this feature was added in Visual Studio version 15.8.

jaredpar:
The way I update reference assemblies is to use `Scripts\Copy-References.ps1`. You'll need to invoke it with something similar to the following:

> .\Scripts\Copy-References.ps1 -vsVersion 2017 -vsDir "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional"

That script though sometimes requires some tweaking to work for the current machine.

One thing I dislike about our current References setup is there isn't really any tracking on which version of VS2017 the reference assemblies represent. That is I can't just look at the directory in source and know what VS version they represent.  Before 2017 this was a mute point as APIs didn't really change past the initially release. Now though VS adds APIs in minor updates and VsVim is increasingly leveraging the new APIs. Up for ideas on how to make that better. Maybe just manual entries via a `README.md` in the directory are a way to go.

ricksladkey:
> One thing I dislike about our current References setup is there isn't really any tracking on which version of VS2017 the reference assemblies represent.

Nowadays, it might be possible to convert to NuGet packages, although the idea of trying this myself in the VsVim solution mortifies me.

I think it might be best if we hold off on VS2017 support and include it in a separate PR. This PR is already oversized and the next PR would be much easier to review when separated out.

ricksladkey:
I think VsVim_Build_Test Vs2015 should be a required check now, right?

jaredpar:
> I think VsVim_Build_Test Vs2015 should be a required check now, right?

Yes it should. Will fix that. 

Going to try and start working my way through reviewing this change today.

ricksladkey:
> Going to try and start working my way through reviewing this change today.

I apologize for the size of this change! However, a way to approach reviewing it might be to break it into two parts. During the first part, you should be able to convince yourself that VsVim behaves almost exactly like it used to behave when there is only one caret or selection. In other words, in spite of the number of changes, nothing really major has changed architecturally and most of the changes you see are to support flexibility or consistency rather than doing things differently.

For example, all callers who manually set or cleared the selection directly using the `ITextSelection` API now route those actions through `TextViewUtil`. This makes it much easier to conditionally change or hook those actions, but conceptually the calling code is doing the exact same thing: settting or clearing the selection. And there are other benefits to consistently using a utility functions rather than directly manipulating the API, included but not limited to currying. This kind of change accounts for a lot of the scattered changes throughout the code base. So yes, there are a lot of scattered changes, but they serve a purpose and are low-risk to the single caret / selection scenario. My desired outcome of the review is primarily to ensure that these changes are "safe" to merge without jeopardizing the non-multiple-cursor base functionality.

The second part of the review can then focus on the new functionality and architecture. For that part of the review you can start with the supplied multiple-selection documentation as a guide to how it works.

Documentation/Multiple Selections.md
```diff
@@ -0,0 +1,96 @@
...
+the primary selection as the real and only selection, while maintaining the
+secondary selections behind the scenes.
+
+#### The Multi-Selection API
+
+Unfortunately, not all versions of the editor API support multiple selections.
+In order to handle this, there are two API entry points in the Vim host:
+
+- `IVimHost.GetSelectedSpans` - member
+- `IVimHost.SetSelectedSpans` - member
```
jaredpar: Once we move past VS2015 support will this API dissapear? At that point we should have direct access to the multi-selection APIs in VimCore hence this shouldn't be necessary anymore (to my understanding).

Documentation/Multiple Selections.md
```diff
@@ -0,0 +1,96 @@
...
+primary selection and any subsequent selected spans are secondary selections,
+sorted in ascending order by position within the text buffer.
+
+In order to insulate the code base from the host, all consumers who use the
+multi-selection API access the selections using related APIs in the common
+operations module:
+
+- `ICommonOperations.SelectedSpans` - property
+- `ICommonOperations.SetSelectedSpans` - member
+- `ICommonOperations.SelectedSpansSet` - event
```
jaredpar: Consider `SelectedSpansChanged`. The `Changed` suffix seems to be more common in the editor APIs. Looking at `MultiSelectionBroker` specifically the event is `MultiSelectionSessionChanged`

Documentation/Multiple Selections.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### VsVim Key Bindings
+
+- `<C-A-LeftMouse>` - add or remove caret (normal, visual, select, insert)
```
jaredpar: Are the modes in the parens listing where the key bindings are valid?

Documentation/Multiple Selections.md
```diff
@@ -0,0 +1,96 @@
...
+- `<C-c>` - clear secondary carets or selections (normal, visual, insert)
+
+#### The Unnamaed Register
+
+Each caret gets their own unnamed register. All secondary caret's unnamed
+registers are copied from the "real" unnamed register whenever a caret is
+added or removed. The result is that if you put from the unnamed register at
+all carets immediately after creating them, all carets will put the same
+value. But if you yank text and the put it while there are multiple carets,
+each caret will put what that caret yanked.
```
jaredpar: Why was this necessary? Reading through the doc it seems that every operation is executed in sequence. Why not just keep the existing unnamed register and then just update it's value accordingly in between the selection operations?

Src/VimCore/CaretRegisterMap.fs
```diff
@@ -0,0 +1,123 @@
+﻿namespace Vim
+
+open System.Collections.Generic
+
+/// Mutable caret index object
+type CaretIndex() =
```
jaredpar: Shouldn't this be `type internal` as it's not exposed in the fsi file?

Src/VimCore/CaretRegisterMap.fs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Caret-aware IRegisterValueBacking implementation for the unnamed register.
+/// If there are multiple carets, each caret gets its own distinct value
+type CaretUnnamedRegisterValueBacking
+    (
+        _caretIndex: CaretIndex,
+        _unnamedRegister: Register
+    ) =
+
+    let valueMap = Dictionary<int, RegisterValue>()
```
jaredpar: `_valueMap`

jaredpar: Won't the caret index values be contiguous? If so then why a map instead of a list?

Src/VimCore/ExternalEdit.fs
```diff
@@ -7,7 +7,7 @@ type internal ExternalEditMode(_vimBufferData: IVimBufferData) =
     member x.ClearSelection () =
         let textView = _vimBufferData.TextView
         let start = textView.Selection.Start
-        textView.Selection.Clear()
```
jaredpar: Random musing: This is where Roslyn analyzers would be nice for F#. Effectively we need to ban `ITextSelection.Clear` in VsVim now. Can't think of a good way to do this in F# except for diligent code review.

Documentation/Multiple Selections.md
ricksladkey: Frankly, I didn't fully understand how Vim.Specific assembly worked and the one tool in my toolbox for escaping from editor-land to WPF-land was `IVimHost` so that's what I initially used. I gradually realized how it was working and that I could eliminate duplicate code by using Vim.Specific. At that point, I was just one step away from ripping out the `IVimHost` API and accessing the `ISelectionUtilService` (which is already in F#) directly from common operations.

I will take a crack at that change later tonight.

Documentation/Multiple Selections.md
ricksladkey: Ah, this is intentional: it doesn't mean when the selected spans change but only when they are set through this API. In other words, the event doesn't fire when external events change the selection. In fact the regular `Selection.SelectionChanged` event still does that. This is our one way of knowing that not only did it change, but it was changed intentionally by VimCore.

So a more accurate name would be something crazy like `SelectedSpansSetUsingTheCommonOperationsApi`. I will clarify this nuance in the documentation comments.

Documentation/Multiple Selections.md
ricksladkey: We could implement the operations correctly as you describe, but then we would need a nearly identical data structure to record what was yanked at each caret.  For example, using this approach avoids a lot of wear and tear on the registers when doing simple things like pure motions. But both approaches could work.

Although I didn't copy their design (I wasn't even aware of their implementation), in hindsight the semantics of how the registers behave is exactly how VSCodeVim behaves.

Src/VimCore/CaretRegisterMap.fs
ricksladkey: This might possible if the caret register map was aware of how many carets there were. Let me think about this.

Src/VimCore/ExternalEdit.fs
ricksladkey: We have a similar situation with several other APIs, notably `ITextView.TextViewLines`, etc. The situation is not as bad you might think because 90% of the time, some existing code is used as model for how to do implement some new code and then the same API gets used because the old code followed the model so the new code also follows the model.

My policy is that code review enforces consistency but the tests enforce correctness. So if someone doesn't use the supplied API but there is no bug as a consequence, it's just inconsistent, not wrong!

Src/VimSpecific/Implementation/Selection/SingleSelectionUtil.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using Vim.VisualStudio.Specific;
+
+namespace Vim.Specific.Implementation.Selection
+{
+    [Export(typeof(ISelectionUtil))]
+    [Export(typeof(IVimSpecificService))]
+    class SingleSelectionUtil : VimSpecificService, ISelectionUtil
```
jaredpar: `internal sealed`

Src/VimSpecific/Implementation/Selection/SingleSelectionUtil.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var caretPoint = textView.Caret.Position.VirtualBufferPosition;
+            var anchorPoint = textView.Selection.AnchorPoint;
+            var activePoint = textView.Selection.ActivePoint;
+            return new[] { new SelectedSpan(caretPoint, anchorPoint, activePoint) };
+        }
+
+        void ISelectionUtil.SetSelectedSpans(ITextView textView, IEnumerable<SelectedSpan> selectedSpans)
+        {
+            var selectedSpan = selectedSpans.First();
```
jaredpar: This will throw if an empty collection is passed for `selectedSpans`

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
```diff
@@ -0,0 +1,83 @@
...
+            var broker = textView.GetMultiSelectionBroker();
+            var primarySelection = broker.PrimarySelection;
+            if (textView.Selection.Mode != TextSelectionMode.Stream)
+            {
+                return new[] { GetSelectedSpan(primarySelection) };
+            }
+            var secondarySelections = broker.AllSelections
+                .Where(span => span != primarySelection)
+                .Select(selection => GetSelectedSpan(selection));
+            return new[] { GetSelectedSpan(primarySelection) }.Concat(secondarySelections);
```
jaredpar: Wow that is unintuitive for `AllSelections` to not include the `PrimarySelection`. Generally I read "all" to mean ... all. Seems like that API would be much better named to be `AllSecondarySelections`. 

This code is definitely correct, but that API surprising.

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
```diff
@@ -0,0 +1,83 @@
...
+                return;
+            }
+
+            var selections = new Microsoft.VisualStudio.Text.Selection[selectedSpans.Length];
+            for (var caretIndex = 0; caretIndex < selectedSpans.Length; caretIndex++)
+            {
+                selections[caretIndex] = GetSelection(selectedSpans[caretIndex]);
+            }
+            var broker = textView.GetMultiSelectionBroker();
+            broker.SetSelectionRange(selections, selections[0]);
```
jaredpar: This will throw if `selectedSpans` is empty

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.ComponentModel.Composition;
+using System.Linq;
+using Vim;
+using Vim.VisualStudio.Specific;
+
+namespace Vim.Specific.Implementation.Selection
+{
+    [Export(typeof(ISelectionUtil))]
+    [Export(typeof(IVimSpecificService))]
+    class MultiSelectionUtil : VimSpecificService, ISelectionUtil
```
jaredpar: `internal sealed`

Src/VsInterfaces/ISharedService.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using Microsoft.FSharp.Core;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
+using System.Collections.Generic;
```
jaredpar: These changes seem unnecessary here.

Src/VsSpecific/VsSpecific/SharedService.cs
```diff
@@ -13,6 +13,7 @@
 using System.ComponentModel.Composition.Primitives;
 using System;
 using Vim.Extensions;
+using Microsoft.VisualStudio.Text;
```
jaredpar: These changes seem unnecessary

Src/VimCore/CoreInterfaces.fs
```diff
@@ -2209,6 +2216,12 @@ type VisualSelection =
 
     with
 
+    member x.IsForward =
+        match x with
+        | Character (_, path) -> path.IsSearchPathForward
+        | Line (_, path, _) -> path.IsSearchPathForward
+        | Block _ -> true
```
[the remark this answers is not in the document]
ricksladkey: Yes, you are right. I'll fix this.

This member is used to determine the primary selected span only when there is one selected span because visual block mode is intentionally not used for multiple carets. Conceptually visual block mode is very similar to multiple carets (in particular block insert and block append) and meta multiple cursors is conceivable but not likely to be useful.

For an editing workflow, one would typically create a normal single-caret block selection, split it into carets, which puts you in normal mode with a caret on each line and then use either:

- normal mode commands
- visual character mode at each caret

the latter of which is very much like "multi-cursor visual block mode".

Documentation/Multiple Selections.md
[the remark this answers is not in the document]
ricksladkey: This is a good question! The reason any registers are worth handling is because we want to support running a macro "in parallel" vs. the normal way "in series". In other words if you could record a macro that doesn't refer to any explicit registers like "move by word, delete word, move by word, put deleted word (e.g. `wdwwP`) that "macro" would no longer work in parallel.

At the moment, the documented behavior is a reasonable compromise that says "basic yank and put" will work for multiple carets but explicit registers and using marks will do what you might expect them to do but that probably won't be useful.

For example, if you place a mark with `ma`, the mark will currently point to the last caret.

Src/VimSpecific/Implementation/Selection/SingleSelectionUtil.cs
ricksladkey: It will throw, but you cannot set no selected spans because there is always at least one caret. So it should check and throw an invalid argument exception. That ambiguity could be avoided if we change from `IEnumerable<SelectedSpan>` to `NonEmptyCollection<SelectedSpan>` but I was trying to keep the API as light as possible because performance is critical if you want to use, say, 100 carets, the memory pressure is fairly severe.

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
ricksladkey: The new multi-selection broker is basically undocumented, at least as far as nuance is concerned. Without documentation, it is impossible to know if we can reliably count on the first selection of all selections to be the primary selection. For example, they could could be auto-sorted by buffer position. If that is the case, then the primary selection might be the third selection.

The contract of the `SelectionUtil` API that we expose to VsVim ensures that the the first selected span is always the primary selection. With better documentation we might be able to make this method more efficient and use less memory, but until then, I chose the safest route.

Src/VsInterfaces/ISharedService.cs
ricksladkey: Sorry, this is left over from when I had the whole thing working through a circuitous path:

- VimCore
- VimHost
- SharedService
- MultiSelectionBroker

That path got much simplified, and I deleted the code, but the usings remained.

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
ricksladkey: Right. I will validate further up that `selections.Length > 0` or use `NonEmptyCollection`.

Src/VimCore/CoreTypes.fs
```diff
@@ -52,6 +52,81 @@ type JoinKind =
     | RemoveEmptySpaces
     | KeepEmptySpaces
 
+[<StructuralEquality>]
+[<NoComparison>]
+[<Struct>]
+[<DebuggerDisplay("{ToString(),nq}")>]
+type SelectedSpan =
```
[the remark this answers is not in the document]
ricksladkey: The Visual Studio type is indeed just called `Microsoft.VisualStudio.Text.Selection` but reusing that name would lead to at least three types with the same name.

I chose "selected span" to mean "a span when it is selected" rather than "a span that is currently selected".

The reason I like `SelectedSpan` is so that the two APIs `GetSelectedSpans` and `SetSelectedSpans` include the type of the expected arguments. But I could be persuaded to rename it `SelectionSpan` like this:

```
    /// Get all the selected spans for the specified text view
    abstract GetSelectedSpans: textView: ITextView -> SelectionSpan seq

    /// Set all the selected spans for the specified text view
    abstract SetSelectedSpans: textView: ITextView -> SelectionSpan seq -> unit
```

Also, I see now that I need to add documentation comments to the entire type and its members.  Oops!

Documentation/Multiple Selections.md
ricksladkey: Yes, I should clarify that.

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
jaredpar: My original comment was wrong: `AllSelections` does include `PrimarySelection`. 

I agree the documentation is lacking here. Eventually I just pulled up the type under ILSpy to figure out how it works. 

> Without documentation, it is impossible to know if we can reliably count on the first selection of all selections to be the primary selection. 

The code seems to allow for the exact opposite: the primary selection being the non-first one. This is based on my reading of `SetSelectionRange`. That takes an `IEnumerable<Selection>` to select and `Selection` which is to be made the primary. The resulting order of `AllSelections` is essentially the `IEnumerable<Selection>` sorted by the `InsertionPoint` value. The `PrimarySelection` value though is just whichever value is reference equal to the `primary` parameter.

Src/VimCore/CommonOperations.fs
```diff
@@ -2387,19 +2427,385 @@ type internal CommonOperations
...
+        then
+            seq {
+                yield! remainingSpans
+                if not isContainedByExistingSpan then
+                    yield SelectedSpan(point)
+            }
+            |> x.SetSelectedSpans
+
+    /// Add the specified selected span
+    member x.AddSelectedSpan selectedSpan =
```
jaredpar: Why does `AddSelectedSpan` not check for an existing span that overlaps the new one but `AddCaretPoint` does check for overlap? Would naiively expect the overlap behavior to be consistent between the two functions.

Src/VimCore/CommonOperations.fs
```diff
@@ -2387,19 +2427,385 @@ type internal CommonOperations
...
+                    Some modeArgument
+                | _ ->
+                    None
+            | _ ->
+                None
+
+        // Get any linked transaction from the specified command result.
+        let getLinkedTransaction result =
+            match getModeArgument result with
+            | Some (ModeArgument.InsertWithTransaction linkedTransaction) ->
```
jaredpar: `InsertWithCountAndNewLine` also has a linked transaction here. Consider just adding a member `LinkedUndoTransaction` to `ModeArgument`

jaredpar: `InsertWithCountAndNewLine` and `InsertBlock` also has a linked transaction here. Consider just adding a member `LinkedUndoTransaction` to `ModeArgument`

Src/VimCore/CommonOperations.fs
```diff
@@ -2387,19 +2427,385 @@ type internal CommonOperations
...
+                    | ModeKind.VisualBlock -> Some VisualKind.Block
+                    | _ -> None
+                | _ ->
+                    None
+            | _ ->
+                None
+
+        // Create a linked undo transaction.
+        let createTransaction () =
+            let name = "MultiSelection"
```
jaredpar: Low pri: Consider including the count of selections in the transaction name. The name appears in our log output, may help to know how many selections there were if we get user reports of bugs.

Src/VimCore/CommonOperations.fs
```diff
@@ -2387,19 +2427,385 @@ type internal CommonOperations
...
+
+                    yield result, newSelectedSpan, newAnchorPoint
+            }
+            |> Seq.toList
+
+        // Do the action for all selections
+        let doActions () =
+
+            // Create a linked transaction for the overall operation.
+            let transaction = createTransaction()
```
jaredpar: This should be a `use` instead of `let` to ensure the transaction is closed out in exceptional circumstances.

Src/VimCore/CommandUtil.fs
```diff
@@ -1603,7 +1609,8 @@ type internal CommandUtil
             CommandResult.Completed ModeSwitch.SwitchPreviousMode
 
     /// Invert the current selection
-    member x.InvertSelection (visualSpan: VisualSpan) (streamSelectionSpan: VirtualSnapshotSpan) columnOnlyInBlock =
+    member x.InvertSelection (visualSpan: VisualSpan) columnOnlyInBlock =
+        let streamSelectionSpan = _textView.Selection.StreamSelectionSpan
```
jaredpar: This doesn't appear to be used.

Src/VimCore/CommandUtil.fs
```diff
@@ -2939,11 +2948,68 @@ type internal CommandUtil
     member x.RunInsertCommand command =
         _insertUtil.RunInsertCommand command
 
+    /// Whether we should run the specified command for all carets
+    member x.ShouldRunNormalCommandForEachCaret command =
```
jaredpar: Think this could be a local function within `RunNormalCommand`

Src/VimCore/CommandUtil.fs
```diff
@@ -3055,21 +3123,77 @@ type internal CommandUtil
         | NormalCommand.Yank motion -> x.RunWithMotion motion (x.YankMotion registerName)
         | NormalCommand.YankLines -> x.YankLines count registerName
 
+    /// Whether we should run the specified visual command for each selection
+    member x.ShouldRunVisualCommandForEachSelection command visualKind =
```
jaredpar: Think this can be a local function inside `RunVisualCommand`

Src/VimCore/CommandUtil.fs
```diff
@@ -3055,21 +3123,77 @@ type internal CommandUtil
...
+            | VisualCommand.ShiftLinesRight -> true
+            | VisualCommand.SubtractFromSelection _ -> true
+            | VisualCommand.CutSelection -> true
+            | VisualCommand.CutSelectionAndPaste -> true
+            | _ -> false
+        | VisualKind.Block ->
+            false
+
+    /// Whether we should clear the selection before running the command
+    member x.ShouldClearSelection command =
```
jaredpar: Think this can be a local function within `RunVisualCommandCore`

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -157,20 +172,22 @@ private bool IsRealCaretVisible
             _textView.LayoutChanged += OnCaretEvent;
             _textView.GotAggregateFocus += OnCaretEvent;
             _textView.LostAggregateFocus += OnCaretEvent;
-            _textView.Caret.PositionChanged += OnCaretPositionChanged;
+            _textView.Selection.SelectionChanged += OnCaretPositionChanged;
```
jaredpar: Don't think the method name is correct anymore given the event it's subscribing to.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4995,6 +5058,15 @@ type IVimHost =
...
     abstract TryCustomProcess: textView: ITextView -> command: InsertCommand -> bool
 
+    /// Whether multi-selection is supported
+    abstract IsMultiSelectionSupported: bool
+
+    /// Get all the selected spans for the specified text view
+    abstract GetSelectedSpans: textView: ITextView -> SelectedSpan seq
+
+    /// Set all the selected spans for the specified text view
+    abstract SetSelectedSpans: textView: ITextView -> SelectedSpan seq -> unit
```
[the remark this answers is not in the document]
ricksladkey: Agreed. I was holding off on that change to avoid pushing commits while you were still reviewing. Originally I was under the impression that this had to be a shared service, and that you had to access a shared service through the vim host. Vim shared changed all that in a good way.

Src/VimCore/CommonOperations.fs
```diff
@@ -2256,7 +2295,8 @@ type internal CommonOperations
     
     /// Ensure the view properties are met at the caret
     member x.EnsureAtCaret viewFlags = 
-        x.DoActionWhenReady (fun () -> x.EnsureAtPointSync x.CaretPoint viewFlags)
+        if _vimBufferData.CaretIndex = 0 then
```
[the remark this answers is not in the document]
ricksladkey: We do not need to consider things like z commands, because those are not commands that `ShouldRunForEachCaret` returns true for. If we say `z.` we mean "center the screen on the primary caret and ignore all secondary carets."

Consider using a caret on every line of a long file, say, a thousand lines long. It is impossible to keep all the carets on the screen at the same time so we only want to ensure view properties:

1. For the primary caret
2. After all actions have been performed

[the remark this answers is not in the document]
ricksladkey: Maybe a better way to describe it is that many Vim commands don't make sense to run for each caret or selection. For example, writing the buffer with `:w<CR>`. If we have five carets we don't want to write the buffer five times. Likewise, centering the screen on the primary caret with `z.` doesn't have a natural multiple-caret interpretation.

Src/VimCore/CommonOperations.fs
ricksladkey: Sorry, `AddCaretAtPoint` should be called `AddOrRemoveCaretAtPoint`. If you add a caret and then click it again, you remove that caret. By contrast, `AddSelectedSpan` is a utility function whose purpose is to always add a new selection." In fact, it is only used once, so it could be a local function.

Src/VimCore/CommonOperations.fs
ricksladkey: No, this transaction can be left open, say, if we are switching to insert mode. But it is true we should complete the transaction if there were an exception.

Src/VimCore/CommandUtil.fs
ricksladkey: Not sure which identifier you mean. If you mean `x.InvertSelection`, it is definitely used by `VisualCommand.InvertSelection`. If you mean `streamSelectionSpan`, it is definitely used about five lines later in the local function `changeSelection`. In general you can tell if something is unused with the F# language service because it will always be "grayed out", i.e. shown in a fainter tag color.

Src/VimCore/CommandUtil.fs
ricksladkey: Ideally we would like to "ask the command" whether it wants to run for each caret, rather than "telling the command" we are going to run it for each caret. in other words, it should be conceptually like flag associated with the abstract normal command. I think this should therefore be moved up into `NormalCommand`.

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: Although it seems odd, this naming does make a certain kind of sense. The selection changed event fires when **any** caret position changes, not just the primary caret. It might be clearer if it was called `OnAnyCaretPositionChanged` or `OnCaretPositionEventOccurred` or even `OnCaretBlinkTimerNeedsUpdating`.

ricksladkey: The need for this change is so that when you add a new caret, the blink timer is reset and the new caret is visible immediately vs. partway through a blink cycle.

Src/VimCore/CommandUtil.fs
jaredpar: I was referring to `streamSelectionSpan` and I just missed that it was used later. The F# language service does make this obvious for general editting but it doesn't work during code reviews (that I know of).

Src/VimCore/CommandUtil.fs
jaredpar: That makes sense

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -94,59 +104,64 @@ public double CaretOpacity
...
                 }
-                return false;
+                textViewLine = null;
             }
+            return false;
+        }
+
+        private bool IsRealCaretVisible(VirtualSnapshotPoint caretPoint)
+        {
+            return IsRealCaretVisible(caretPoint, out var textViewLine);
```
jaredpar: Nit: given that `textViewLine` is unused here you could use a discard to make it explicity:

```
return IsRealCaretVisible(caretPoint, out _);
```

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -424,55 +480,54 @@ private Size CalculateCaretSize(out string caretCharacter)
...
 
-        private Tuple<Rect, double, string> CalculateCaretRectAndDisplayOffset()
+        private Tuple<Rect, double, string> CalculateCaretRectAndDisplayOffset(VirtualSnapshotPoint caretPoint)
         {
-            var size = CalculateCaretSize(out string caretCharacter);
-            var caretPoint = GetRealCaretVisualPoint();
-            var blockPoint = caretPoint;
+            var size = CalculateCaretSize(caretPoint, out string caretCharacter);
+            var point = GetRealCaretVisualPoint(caretPoint);
+            var blockPoint = point;
```
jaredpar: Is it okay for this logic to run when `point` has the vale `s_invalidPoint`?

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
```diff
@@ -47,18 +52,23 @@ internal sealed class BlockCaret : IBlockCaret
...
         private readonly IAdornmentLayer _adornmentLayer;
-        private readonly object _tag = new object();
+        private readonly List<object> _tags = new List<object>();
+        private readonly HashSet<object> _adornmentsPresent = new HashSet<object>();
         private readonly DispatcherTimer _blinkTimer;
         private readonly IControlCharUtil _controlCharUtil;
-        private CaretData? _caretData;
+
+        private List<VirtualSnapshotPoint> _caretPoints = new List<VirtualSnapshotPoint>();
+        private Dictionary<int, CaretData> _caretDataMap = new Dictionary<int, CaretData>();
```
jaredpar: Rather than have four different collections that need to be kept in sync for each caret I think it would be easier to have a single collection and push all the data into a new type (or possibly augment `CaretData`) to have the new type. Consider if we had this type:

```
internal sealed class CaretViewData {
  internal CaretData { get; set; }
  internal bool IsAdornmentPresent { get; set; }
  internal object Tag { get; } = new object();
  internal VirtualSnapshotPoint Point { get; set; }
}
```

Seems like that could allow us to move to a single collection to manager here and would make some of the code below a bit simpler.

ricksladkey: I like this idea. By way of explanation for doing the way that I did, because the block caret is so central to Vim, I tried to make the logic exactly parallel the old block caret, especially for the primary caret but also for all carets. I was pleasantly surprised that the multi-caret support "just worked" and I didn't have any trouble with it. I'll take a look at streamlining it.

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: I will check this.

ricksladkey:
OK, I've completed the major reorganization of bypassing the vim host for access to the selection API. That was a bigger change than I anticipated, but I've got it all working now. Next I will try to address the other individual review feedback items and then the PR will be ready for the next review pass.

jaredpar:
> Next I will try to address the other individual review feedback items and then the PR will be ready for the next review pass.

Did you want me to review the state now or wait for rest of the change?

ricksladkey:
> Did you want me to review the state now or wait for rest of the change?

I think you can wait for the rest, thanks.

Src/VimCore/CaretRegisterMap.fs
ricksladkey: I think it's better to think of this as an associative array where items come into existence when you refer to them.

Documentation/Multiple Selections.md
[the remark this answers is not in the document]
ricksladkey: When the dust settles we can consider what the best strategy for registers other than the unnamed register is.

Src/VimSpecific/Implementation/Selection/MultiSelectionUtil.cs
ricksladkey: Thanks for sussing that out.

Src/VimCore/CommonOperations.fs
ricksladkey: Without catching and re-raising any exception, I had to use a mutable flag to properly close the transaction. Not sure if there is a more elegant way of accomplishing the desired effect.

- No exceptions, command switches to insert -> use and leave open
- No exceptions, command does not switch to insert -> close
- Exception -> Don't interfere but close the transaction

Src/VimWpf/Implementation/BlockCaret/BlockCaret.cs
ricksladkey: I've looked long and hard at this idea, and in my opinion, the data structures exactly match the expected behavior and performance characteristics of their respective operations. For example, the tag array does and should have a different cardinality than caret data map. The adornments present map cannot be converted to an array field without maintaining a separate "double index" to allow efficient lookup. The caret points themselves are simply a cached snapshot of an external interface and doesn't imply that **any** carets are actually visible. Finally, the caret data map leverages the immutability of caret data in its values.

ricksladkey:
@jarepad, OK, this change is now ready for the next review pass. Because of the size of the review, some small items might have slipped through the cracks, but I hope I hit them all and specifically the big ones.

Src/VimApp/VimAppHost.cs
```diff
@@ -14,11 +14,15 @@
 using Vim.Extensions;
 using Vim.Interpreter;
 using Vim.VisualStudio.Specific;
+using System.Collections.Generic;
 
 namespace VimApp
 {
     [Export(typeof(IVimHost))]
+    [Export(typeof(IWpfTextViewCreationListener))]
```
jaredpar: Wow that was an oversight. Not sure I want to figure out how long this has been broken.

Src/VimCore/CaretRegisterMap.fs
```diff
@@ -0,0 +1,123 @@
...
+            _clipboardRegister.RegisterValue <- value
+        else
+            _caretUnnamedRegister.RegisterValue <- value
+
+    interface IRegisterValueBacking with
+        member x.RegisterValue 
+            with get () = x.RegisterValue
+            and set value = x.SetRegisterValue value
+
+type CaretRegisterMap
```
jaredpar: This should be `type internal` to match the definition in the FSI file. I'm a bit surprised that F# doesn't warn about this case. Not sure which wins on emit but I'm guessing that it's the version in the .fs file.

Src/VimCore/CaretRegisterMap.fsi
```diff
@@ -0,0 +1,7 @@
+﻿namespace Vim
+
+
+type internal CaretRegisterMap = 
+    new: IRegisterMap -> CaretRegisterMap
```
jaredpar: Consider giving the constructor parameter a name here.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -128,6 +127,30 @@ type IFileSystem =
...
+type ISelectionUtil =
+
+    /// Whether multi-selection is supported
+    abstract IsMultiSelectionSupported: bool
+
+    /// Get all the selected spans for the specified text view
+    abstract GetSelectedSpans: unit -> SelectionSpan seq
+
+    /// Set all the selected spans for the specified text view
+    abstract SetSelectedSpans: SelectionSpan seq -> unit
```
jaredpar: Consider giving the parameter here a name.

Src/VimCore/MefComponents.fs
```diff
@@ -533,3 +533,47 @@ type internal VimWordCompletionSessionFactoryService
...
+type internal SelectionUtilService 
+    [<ImportingConstructor>]
+    (
+        _vimSpecificServiceHost: IVimSpecificServiceHost
+    ) =
+
+    member x.GetSelectionUtilFactory () =
+        match _vimSpecificServiceHost.GetService<ISelectionUtilFactory>() with
+        | Some selectionUtilFactory -> selectionUtilFactory
+        | None -> SingleSelectionUtilFactory() :> ISelectionUtilFactory
```
jaredpar: Why create a new instance here for every call versus making this a singleton? The singleton approach is generally what the editor / VsVim does for components.

Src/VimSpecific/Implementation/MultiSelection/MultiSelectionUtilFactory.cs
```diff
@@ -0,0 +1,103 @@
...
+using Vim;
+using Vim.VisualStudio.Specific;
+
+namespace Vim.Specific.Implementation.MultiSelection
+{
+    [Export(typeof(ISelectionUtilFactory))]
+    [Export(typeof(IVimSpecificService))]
+    internal sealed class MultiSelectionUtilFactory : VimSpecificService, ISelectionUtilFactory
+    {
+        private class MultiSelectionUtil : ISelectionUtil
```
jaredpar: `sealed`

Src/VimCore/MefComponents.fs
ricksladkey: Fixed this here to use a singleton and also fixed both the single and multi selection factories themselves to do the same thing.

PR #2654, opened 2019-06-20
Change Completion to vim style.

sqlkoala:


sqlkoala:
Hi,
I wanted to take up that topic again. First, how does it work in vim. After you start a completion with C-n/C-p
image
there are these options:
* You can move up/down the list
* You can exit/discard the completion with C-e.
* You can accept it with C-y
* Any other key (a letter, space, Tab, Enter) also accepts the entry but the key is also inserted.

Right now there is a difference between Vsvim Async and Legacy and both differ from vim (async more: in that the last part does not work). With this PR, completion will behave as in vim.

We could (for users mostly used to VS) make Tab and Enter behave like C-y ( Accept completion and key beeing consumed, not inserted).
One thing I could not get work: discarding in legacy layer, after C-e I still have "cat dog" and not "c dog".

`            SendCommand(IntellisenseKeyboardCommand.Escape);
`
 I tried this in: \WordLegacyCompletionSession.cs ::  IWordCompletionSession.Dismiss()

Test/VimCoreTest/InsertModeIntegrationTest.cs
```diff
@@ -2547,7 +2547,7 @@ public void WordCompletion_Simple_Async()
                 _textView.MoveCaretTo(1);
                 _vimBuffer.ProcessNotation("<C-N>");
                 Dispatcher.DoEvents();
-                _vimBuffer.ProcessNotation("<CR>");
+                _vimBuffer.ProcessNotation("<C-y>");
```
jaredpar: We should still test the `<CR>` behavior here.

Test/VimCoreTest/InsertModeIntegrationTest.cs
```diff
@@ -2599,10 +2623,24 @@ public void WordCompletion_ChooseNext_Async()
                 _textView.MoveCaretTo(1);
                 _vimBuffer.ProcessNotation("<C-N>");
                 Dispatcher.DoEvents();
-                _vimBuffer.ProcessNotation("<C-N><CR>");
+                _vimBuffer.ProcessNotation("<C-N><C-y>");
```
jaredpar: Nit: prefer `<C-Y>` here over `<C-y>` as that is the more common style in the code base.

Test/VimCoreTest/InsertModeIntegrationTest.cs
```diff
@@ -2599,10 +2623,24 @@ public void WordCompletion_ChooseNext_Async()
...
+            /// </summary>
+            [AsyncCompletionWpfFact]
+            public void WordCompletion_Abort_Async()
+            {
+                Create("c dog", "cat copter");
+                _textView.MoveCaretTo(1);
+                _vimBuffer.ProcessNotation("<C-N>");
+                Dispatcher.DoEvents();
+                _vimBuffer.ProcessNotation("<C-e>");
+                Assert.Equal("c dog", _textView.GetLine(0).GetText());
```
jaredpar: Should verify the caret position here too.

Test/VimCoreTest/InsertModeIntegrationTest.cs
jaredpar: Huh. It seems the code and tests have a different style here. Strange. Would prefer the test at least use one of the styles vs. both. 

---
In reply to: [296050137](https://github.com/VsVim/VsVim/pull/2654#discussion_r296050137) [](ancestors = 296050137)

Test/VimCoreTest/InsertModeIntegrationTest.cs
```diff
@@ -2655,7 +2693,7 @@ public void WordCompletion_Escape_Async()
                 Dispatcher.DoEvents();
                 _vimBuffer.Process(KeyNotationUtil.StringToKeyInput("<Esc>"));
                 Assert.Equal(ModeKind.Normal, _vimBuffer.ModeKind);
-                Assert.Equal(0, _textView.GetCaretPoint().Position);
+                Assert.Equal(2, _textView.GetCaretPoint().Position);
```
jaredpar: Why is `2` correct here? I can understand why `1` would be correct here but not `2` as it moves beyond the original word.

jaredpar:
Done with review pass (iteration 14)

Test/VimCoreTest/InsertModeIntegrationTest.cs
ricksladkey: All the mode maps use the lower case convention, and I thought I liked the lowercase convention, but the Vim docs systematically use the uppercase convention (and most of the unit tests). Our own `:map` command (to display mapped key sequences) uses the uppercase convention, as does Vim's `:map` command.

So if ought to be consistent, uppercase makes sense.

sqlkoala:
Do you guys need further input here? (I read that you were moving)

Test/VimCoreTest/InsertModeIntegrationTest.cs
ricksladkey: My testing of Vim completes the whole word `cat` and escape exits insert mode thus moving the caret back to `2`, so `2` is the Vim behavior. This would be clearer if the test also asserted the contents of the line after the operation.

ricksladkey:
> Do you guys need further input here? (I read that you were moving)

I see you addressed all the comments from the first review by @jaredpar except the last one. I manually tested that case and added a comment above explaining how your changes are correct. I was letting Jared be the reviewer for this because he knows more about Vim completion than I do and recently worked on it. I'm pretty sure this will be ready to go once he gets a chance to look at it.

sqlkoala:
I have implemented `C-e` (aborting completion) for legacy now  using a Replace to initial text in Dismiss (no extra work needed in async).
One thing I want to discuss about this code:
The call to Cancel... could call Dismiss() again (including this Replace), after it was just commited. I did not like that, so I set isDismissed to prevent it. But I am not sure this is the best way.

    // Any other key should commit the IWordCompletionSession and we should process
    // the KeyInput as normal
    wordCompletionSession.Commit()
    // Commit will set IsDismissed on wordCompletionSession so CancelWordCompletionSession will not call Dismiss() it
    // Cancel as finish. we need this because otherwise we get a stackoverflow in tests from call to ProcessCore
    x.CancelWordCompletionSession()
    x.ProcessCore keyInput

jaredpar:
@sqlkoala i was mostly waiting on a response to the 0 vs. 2 question on one of the tests. Looks like @ricksladkey got to the bottom of it though and makes sense now.

One last thing: the first two commits have a different commit author (presumably) than the rest of the commits. Notice your icon changes on GitHub. That typically means the commit author info got changed between the commits. 

Makes no difference to me. But wanted to make sure you didn't want to fix the commiter info, rebase before merging. If not then I'm good to merge.

sqlkoala:
Hi @jaredpar , from my side too, we can merge.
About the authors: they are all my commits. I had setup git (for the first time) to work on this and changed my local config later.

PR #2668, opened 2019-07-03
Support end-of-line block mode

### LINKED ISSUE #2667: Pressing 'A' when in Visual Block mode moves caret to wrong position
**Describe the bug**
The caret should move to the right of the block, not the end of the line.

**To Reproduce**

1.  Select a block of text in Visual Block mode
2.  Press 'A'
3.  The caret is in the wrong position.

**Expected behavior**
visualblock

**Screenshots**

 visualblockVS

- Visual Studio version: 15.8.7
- VsVim version: 2.7.0.0
- Programming Language: C# but probably any
    - [ ] ReSharper
    - [ ] Visual Assist

**Additional context**

Note:
The following is written about contributing.
https://github.com/VsVim/VsVim/blob/master/CONTRIBUTING.md

General Keyboard configuration problems are described below.
https://github.com/VsVim/VsVim/issues/2527

ricksladkey:
#### Issues

- Closes #2667

#### Discussion

This change fixes issue #2667, but it also adds support for appending at the end-of-line from visual block mode. The wrinkle is that Vim behaves differently when you have used the `$` (end-of-line) motion since entering visual block mode. If you have done that, the visual mode switches to a sub-mode of visual block mode where there are variable length selections for each line included in the block.

Unfortunately, as far as I can tell, Visual Studio does not support displaying this type of selection properly, but the semantics of block append do now work correctly in that sub-mode.


sqlkoala:
That's strange. I am pretty sure that I already use "block append" at EOL in 2.6 or some 2.6.xx version.
It is as you describe, the visual block to EOL is not show properly but the append works.

ricksladkey:
> That's strange. I am pretty sure that I already use "block append" at EOL in 2.6 or some 2.6.xx version.
> It is as you describe, the visual block to EOL is not show properly but the append works.

I should clarify that VsVim 2.7.0 **always** appends at the end-of-line when using block append, which is not Vim-compatible. What I implemented was to only behave in this way when `$` has been used in visual block mode.

To summarize:

- To append at the end of all included lines from visual block mode, type `$`, then `<S-a>`
- To append at the end of the primary span of the block selection, do not use `$` and then type `<S-a>`

The former does not display the same selection as Vim would, but the behavior is the same.

ricksladkey:
To show that end-of-line block mode affects more than just block append, you can use end-of-line block mode to "chop" the ends off of lines of variable length without using virtual edit. Just type `$d` in visual block mode. This did **not** work in VsVim 2.7.0, but is another consequence of end-of-line block mode.

ricksladkey:
After some experimentation in Vim, I realized that the end-of-line in end-of-line block mode exactly coincides with whether we are maintaining end-of-line for up/down line motions. So, for example, if we have typed `$h` in visual block mode we are no longer in end-of-line block mode because the `h` motion restored maintaining caret spaces instead of end-of-line.

ricksladkey:
Sigh, reading the vim help more carefully, block append differs from block insert (intentionally!) in that block append pads short lines and block insert skips them. I've added that feature with a test.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1802,35 +1810,42 @@ type BlockSpan =
...
+                let lineNumber = x.Start.LineNumber + i
+                match SnapshotUtil.TryGetLine x.Snapshot lineNumber with
+                | None -> ()
+                | Some line -> 
+                    yield makeSpan line x.BeforeSpaces
+        }
         |> NonEmptyCollectionUtil.OfSeq 
         |> Option.get
 
+    member private x.GetLastColumn line = SnapshotColumn(line, line.End)
```
jaredpar: Shouldn't this be called `GetEndColumn`?

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1709,17 +1709,22 @@ type BlockSpan =
...
     val private _height: int
+    val private _endOfLine: bool
 
     new(startPoint: SnapshotPoint, tabStop, spaces, height) = 
         let startColumn = VirtualSnapshotColumn(startPoint)
-        { _startColumn = startColumn; _tabStop = tabStop; _spaces = spaces; _height = height }
+        { _startColumn = startColumn; _tabStop = tabStop; _spaces = spaces; _height = height; _endOfLine = false }
 
-    new(startPoint: VirtualSnapshotPoint, tabStop, spaces, height) =
+    new(startPoint: SnapshotPoint, tabStop, spaces, height, endOfLine) =
```
jaredpar: When is it valid to have `endOfLine == true` and `spaces != 0`? Aren't these properties mutually exclusive?

Src/VimCore/CoreInterfaces.fs
ricksladkey: Ooh, yes, it should.

Src/VimCore/CoreInterfaces.fs
ricksladkey: The spaces are usually non-zero when `endOfLine` is true. It represents the width in spaces of the "proxy block selection", typically the length in spaces from the anchor point to the caret point on whatever line the caret is on.

In other words, because Visual Studio won't allow us to display a ragged box selection, we have to display **some** selection, so we use a regular box selection with the width appropriate for the current line. That selection will be too narrow for some lines and too wide for others, but it will be exactly right for the caret line. So the "spaces" value (even though end-of-line is set) still represents the space width for the line the caret is on.

jaredpar: Gotcha. Can we doc that somewhere? 

---
In reply to: [300842389](https://github.com/VsVim/VsVim/pull/2668#discussion_r300842389) [](ancestors = 300842389)

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1840,15 +1855,23 @@ type BlockSpan =
...
+                    SnapshotOverlapColumn.GetColumnForSpacesOrEnd(line, beforeSpaces + x.SpacesLength, x.TabStop)
             SnapshotOverlapColumnSpan(startColumn, endColumn, x.TabStop))
 
     member x.BlockSpans = x.BlockColumnSpans |> NonEmptyCollectionUtil.Map (fun s -> s.Span)
 
     member x.BlockVirtualSpans = x.BlockVirtualColumnSpans |> NonEmptyCollectionUtil.Map (fun s -> s.VirtualSpan)
 
+    /// Adjust the block span for the specified 'end-of-line' setting
+    member x.AdjustForEndOfLine endOfLine =
+            BlockSpan(x.VirtualStart, x.TabStop, x.SpacesLength, x.Height, endOfLine)
```
jaredpar: Nit: indent is a bit off here.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1840,15 +1855,23 @@ type BlockSpan =
...
+                else
+                    SnapshotOverlapColumn.GetColumnForSpacesOrEnd(line, beforeSpaces + x.SpacesLength, x.TabStop)
             SnapshotOverlapColumnSpan(startColumn, endColumn, x.TabStop))
 
     member x.BlockSpans = x.BlockColumnSpans |> NonEmptyCollectionUtil.Map (fun s -> s.Span)
 
     member x.BlockVirtualSpans = x.BlockVirtualColumnSpans |> NonEmptyCollectionUtil.Map (fun s -> s.VirtualSpan)
 
+    /// Adjust the block span for the specified 'end-of-line' setting
+    member x.AdjustForEndOfLine endOfLine =
```
jaredpar: Think we should call this `WithEndOfLine`. This is essentially just changing a single property without performing any other logic / behavior under the hood. The intent of the code base was to use the naming form `With<property name>` up until now. It's a nod to F# record update expressions `{e with EndOfLine = true }`. 

Few examples (that I looked up to make sure my memory was right here): `IntervalTree` and `SnapshotOverlapColumn.WithTabStop`.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -1991,6 +2014,15 @@ type VisualSpan =
         else
             x
 
+    /// Adjust the visual span for a new 'end-of-line' setting
+    member x.AdjustForEndOfLine endOfLine =
```
jaredpar: Same `With` feedback

Src/VimCore/CoreInterfaces.fs
```diff
@@ -2232,9 +2264,18 @@ type VisualSelection =
                 // The width of a block span decreases by 1 in exclusive.  The minimum though
                 // is still 1
                 let width = max (blockSpan.SpacesLength - 1) 1
-                let blockSpan = BlockSpan(blockSpan.VirtualStart, blockSpan.TabStop, width, blockSpan.Height)
+                let blockSpan = BlockSpan(blockSpan.VirtualStart, blockSpan.TabStop, width, blockSpan.Height, blockSpan.EndOfLine)
                 VisualSelection.Block (blockSpan, blockCaretLocation)
 
+    /// Adjust the visual selection for a new 'end-of-line' setting
+    member x.AdjustForEndOfLine endOfLine =
```
jaredpar: Same `With` feedback

Src/VimCore/CoreInterfaces.fs
```diff
@@ -2452,7 +2493,7 @@ type VisualSelection =
                 | VisualSpan.Line _ -> visualSpan
                 | VisualSpan.Block blockSpan ->
                     let width = blockSpan.SpacesLength + 1
-                    let blockSpan = BlockSpan(blockSpan.VirtualStart, blockSpan.TabStop, width, blockSpan.Height)
+                    let blockSpan = BlockSpan(blockSpan.VirtualStart, blockSpan.TabStop, width, blockSpan.Height, false)
```
jaredpar: Can we use a named argument here to make the meaning of `false` explicit?

Src/VimCore/CoreInterfaces.fs
```diff
@@ -2570,6 +2611,14 @@ type TextObjectKind =
     | AlwaysCharacter
     | AlwaysLine
 
+/// The kind of visual insertion to perform
+[<RequireQualifiedAccess>]
+[<NoComparison>]
+type VisualInsertKind =
+    | Start
+    | End
+    | EndOfLine
```
jaredpar: Can we add a comment that explains the difference between `End` and `EndOfLine`? Not sure it's obvious at a glance.

PR #2673, opened 2019-07-07
Avoid inappropriate scrolling

### LINKED ISSUE #2643: VsVim should not process external caret movements when disabled
VsVim supports ensuring view properties, such as non-zero `scrolloff`, even when the caret movement is performed externally. However, when VsVim is disabled, it should also disable its responses to external caret movements.

### LINKED ISSUE #2664: Highlighting text with the mouse in split view causes the other view to move
**Describe the bug**
In split view, highlighting a section of text in one view causes the other view to scroll to the same position, which defeats the purpose of split view.

**To Reproduce**

In VS2019:

1. Window -> Split
2. Scroll to some location in the code in one view different from the other
3. Select some text with the mouse

**Expected behavior**
The other view stays put.

**Screenshots**
https://imgur.com/j2YE8wQ

- Visual Studio version: VS2019 Community
- VsVim version: 2.7.0.0
- Programming Language: agnostic (F# in example, but also tested in C#, .ini files etc.)
    - [ ] ReSharper
    - [ ] Visual Assist

ricksladkey:
#### Issues

- Closes #2643
- Closes #2664

#### Discussion

This change does not address the problem of multiple text views sharing the same underlying text buffer and therefore the same mode. Nor am I certain that we have exhausted all the places where inappropriate scrolling could occur. However, this change does address the specific issue of allowing the two text views to maintain separate viewports for the situation described.

Since the non-focused text view is mostly passive anyway, I think the worst that can happen is that the non-focused text view is in a strange state when you switch back to it, which typing escape one or more times should clear up.


Src/VimCore/CaretChangeTracker.fsi
```diff
@@ -0,0 +1,6 @@
+﻿#light
+
+namespace Vim
+
+type internal CaretChangeTracker =
+    new: vimBuffer: IVimBuffer * ICommonOperations * IMouseDevice -> CaretChangeTracker
```
jaredpar: Can you add names to all the parameters?

jaredpar:
Curious: know all focus issues were tricky to test before because the unit tests never actually displayed the WPF text editor. Not much to focus if it's not displayed. Now that we have the ability to display the WPF editor during the unit tests (`CreateTextViewDisplay`) that _should_ give us the ability to test focus issues too (not a strong area for me so could be wrong). Is it possible to test this type of change now with that new functionality?

ricksladkey:
>  Is it possible to test this type of change now with that new functionality?

Good idea. Because `IVimHost.IsFocused` is mocked, we can test it without actually displaying either window. I've added a test for that use case and tested that it failed without the fix.

PR #2678, opened 2019-07-09
Support capturing the mouse and auto-scrolling the window

### LINKED ISSUE #2661: Mouse select and drag does not scroll
**Describe the bug**

VsVim blocking scrolling+select

**To Reproduce**

Left-click anywhere in the editor window,drag down to select the text. Editor will stop scrolling when the mouse hits the bottom of the current view (sometimes it scrolls a bit further and then stops)

Disabling VsVim returns the functionality.

**Expected behavior**
I should be able to drag and select like my life depended on it.

**Screenshots**
None

- Visual Studio version: 2019
- VsVim version: 2.7.0.0
- Programming Language: C++
    - [ ] ReSharper
    - [ ] Visual Assist

ricksladkey:
#### Issues

- Closes #2661

### DIFF
--- a/Src/VimWpf/Implementation/Mouse/VimMouseProcessor.cs
+++ b/Src/VimWpf/Implementation/Mouse/VimMouseProcessor.cs
@@ -1,8 +1,5 @@
 ﻿using Microsoft.VisualStudio.Text.Editor;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Windows.Input;
 using System.Windows.Threading;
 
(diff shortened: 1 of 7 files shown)


Src/VimWpf/Implementation/Mouse/VimMouseProcessor.cs
```diff
@@ -1,34 +1,55 @@
...
+using System.Windows.Threading;
 
 namespace Vim.UI.Wpf.Implementation.Mouse
 {
     internal sealed class VimMouseProcessor : MouseProcessorBase
     {
+        /// <summary>
+        /// How many lines per second we should scroll when auto-scrolling
+        /// </summary>
+        private static readonly int s_linesPerSecond = 8;
```
jaredpar: Is there a reason behind 8 here? Or is it just a number that you found works well with auto-scrolling?

ricksladkey: Manually tuned. I started at 20, dropped it to 10 and then dropped it to 5, which was too slow. There is a scroll multiplier the further away from the edge of the viewport the mouse is, so the user can get 8 / 16 / 24 / 32 ... etc. line per second scrolling.

PR #2684, opened 2019-07-10
Support command line editing keys such as left arrow in mappings

### LINKED ISSUE #1103: Support command line cursor movement by mappings
I have a mapping that types text into the command line and then moves the cursor using the arrow keys.

``` vim
" substitute
nnoremap <Leader>s :%s//g<left><left>
" in visual mode, the range '<,'> is typed automatically
vnoremap <Leader>s :s//g<left><left>
```

When I trigger the mapping, the text `%s//g` gets typed properly into the command line, but the cursor does not appear. I am still in the mode where what you type gets added to the end. After the mapping runs, I must press Left three times to get into the intended state.

Fixing #1102 might automatically fix this bug, depending on how you implement it.

### LINKED ISSUE #2683: Mapping <Left> produces invalid characters
**Steps to reproduce the behavior:**

1.  Paste `inoremap foo <Esc>:%s/bar//g<Left><Left>` in _vimrc
1.  Source, if necessary
1.  In insert mode, type `foo`
1.  See `%s/bar//g[][]` in command window, where [] is an invalid character

**Expected behavior**
We expect to see `%s/bar//g` in the command window with the cursor two spaces from the right

- Visual Studio version: 14.0.25420.01
- VsVim version: 2.7.0.1
- Programming Language: X++
    - [ ] ReSharper
    - [ ] Visual Assist

ricksladkey:
#### Issues

- Closes #1103 
- Closes #2683

#### Discussion

This issue has been open since 2013! And it was surprisingly difficult to fix because of the complex interconnection of history, command mode, and the command margin. The main idea is that instead of passing around a string to represent an in-progress command, we need to supplement the text of the command with the current insertion point. As a result, there is a new structure called `EditableCommand` that combines text with a caret position. In addition to the `Left` operation, an editable command also now supports the related Vim keys `Home`, `End`, `Left`, `Right` and `Delete` which, of course, can also can be mapped.

### DIFF
--- a/Src/VimCore/CoreInterfaces.fs
+++ b/Src/VimCore/CoreInterfaces.fs
@@ -4606,6 +4606,85 @@ type HistoryList () =
             let seq = _list :> string seq
             seq.GetEnumerator()
 
+/// Represents a single line of text being edited
+[<StructuralEquality>]
+[<NoComparison>]
+[<Struct>]
+type EditableCommand =
+
+    val private _text: string
+    val private _caretPosition: int
+
+    /// Constructor for initial text with the caret at the end
+    new (text: string) =
+        { _text = text; _caretPosition = text.Length }
+
+    /// Constructor for text and caret position
+    new (text: string, caretPosition: int) =
+        { _text = text; _caretPosition = caretPosition }
+
+    /// The text of the command
+    member x.Text = x._text
+
+    /// The caret position of the command
+    member x.CaretPosition = x._caretPosition
+
+    /// Home key operation
+    member x.Home () =
+        if x._caretPosition <> 0 then
+            EditableCommand(x.Text, 0)
+        else
+            x
+
+    /// End key operation
+    member x.End () =
+        if x.CaretPosition <> x.Text.Length then
+            EditableCommand(x.Text, x.Text.Length)
+        else
+            x
+
+    /// Left arrow key operation
+    member x.LeftArrow () =
+        if x.CaretPosition > 0 then
+            EditableCommand(x.Text, x.CaretPosition - 1)
+        else
+            x
+
+    /// Right arrow key operation
+    member x.RightArrow () =
+        if x.CaretPosition < x.Text.Length then
+            EditableCommand(x.Text, x.CaretPosition + 1)
+        else
+            x
+
+    /// Backspace key operation
+    member x.Backspace () =
+        if x.CaretPosition > 0 then
+            let before = x.Text.Substring(0, x.CaretPosition - 1)
+            let after = x.Text.Substring(x.CaretPosition)
+            EditableCommand(before + after, x.CaretPosition  - 1)
+        else
+            x
+
+    /// Delete key operation
+    member x.Delete () =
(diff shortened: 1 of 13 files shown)


Src/VimCore/CoreInterfaces.fs
```diff
@@ -4606,6 +4606,90 @@ type HistoryList () =
...
+    val private _text: string
+    val private _caretPosition: int
+
+    /// Constructor for initial text with the caret at the end
+    new (text: string) =
+        { _text = text; _caretPosition = text.Length }
+
+    /// Constructor for text and caret position
+    new (text: string, caretPosition: int) =
+        { _text = text; _caretPosition = caretPosition }
```
jaredpar: Did you consider using an optional parameter to make it a single ctor?

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4606,6 +4606,90 @@ type HistoryList () =
...
+
+    /// The text of the command
+    member x.Text = x._text
+
+    /// The caret position of the command
+    member x.CaretPosition = x._caretPosition
+
+    /// Home key operation
+    member x.Home () =
+        if x._caretPosition <> 0 then
```
jaredpar: Nit: `x.CaretPosition` is more consistent with the other members here.

Src/VimCore/CoreInterfaces.fs
```diff
@@ -4606,6 +4606,90 @@ type HistoryList () =
...
+
+    /// Insert text at the caret
+    member x.InsertText text =
+        let before = x.Text.Substring(0, x.CaretPosition)
+        let after = x.Text.Substring(x.CaretPosition)
+        EditableCommand(before + text + after, x.CaretPosition + text.Length)
+
+    /// Display representation of an editable command
+    override x.ToString() =
+        String.Format("Text = '{0}', CaretPosition = {1}", x.Text, x.CaretPosition)
```
jaredpar: `sprintf "Text = %s, CaretPosition = %i" x.Text x.CaretPosition`

Src/VimCore/CoreInterfaces.fs
jaredpar: Hmm ... that may not work well with the C# calling code though.

---
In reply to: [302303313](https://github.com/VsVim/VsVim/pull/2684#discussion_r302303313) [](ancestors = 302303313)

ricksladkey: I also thought about omitting the one argument constructor entirely so that each caller would be forced to at least consider whether the caret position was correct. I also agonized over the name of the type.  These are type names I seriously considered (seriously!):

- `CommandLineState`
- `CommandData` (already means something else)
- `CommandLineData`
- `CommandView` (analogous to `TextView`)
- `InProgressCommand`
- `EditCommand` (is it a command for editing or a a command line related to editing?)
- `CommandInput`
- `TextWithCaret`
- `EditableCommand`

In the end `EditableCommand` was the most descriptive and the least offensive!

Issue #2685, opened 2019-07-10
Development builds inadvertently being published to the Open VSIX Gallery


ricksladkey:
For reasons that I don't fully understand, some development builds, that is builds not merged into the VsVim master branch, are being published to the Open VSIX Gallery. As a result, using the Open VSIX Gallery VsVim feed no longer ensures that you get the build corresponding to the head of the master branch.

Here is an example of a build that was published to the Open VSIX Gallery that was not merged into the master branch:

- https://dev.azure.com/VsVim/VsVim/_build/results?buildId=271&view=logs

In fact, the corresponding unit test jobs failed for that build.

jaredpar:
That definitely shouldn't be happening. Attempted to account for that when I moved us to Azure DevOps but looks like I didn't get it quite right. Looking into it.

ricksladkey:
Closed by #2686

PR #2686, opened 2019-07-10
Publish should only occur on master

jaredpar:
VSIX gallery publishing should only occur on builds in the master
branch.

Before this change publish was occuring whenever the build was
IndividualCI or BatchedCI. The CI term in Azure DevOps means a build
that occurs on merged code (not in PR). This meant every time DevOps was
triggered on merged code the result was published.

The initial intent was that only 'master' was setup to build on merge
but some developmental code got left in and it ended up triggering on
dev branches as well as master. This fixes both problems.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,4 @@
 trigger:
-- dev/jaredpar/*
 - master
 
 pr:
@@ -87,5 +86,5 @@ jobs:
     inputs:
       filePath: Scripts\Build.ps1
       arguments: -ci -uploadVsix -config Release
-    condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI'))
+    condition: and(succeeded(), eq(variables['Build.SourceBranchName'], 'master'))
 


jaredpar:
fixes #2685

azure-pipelines.yml
```diff
@@ -1,5 +1,4 @@
 trigger:
-- dev/jaredpar/*
```
jaredpar: Builds no longer occur when code is merged into `dev/jarepdar`

azure-pipelines.yml
```diff
@@ -87,5 +86,5 @@ jobs:
     inputs:
       filePath: Scripts\Build.ps1
       arguments: -ci -uploadVsix -config Release
-    condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI'))
+    condition: and(succeeded(), eq(variables['Build.SourceBranchName'], 'master'))
```
jaredpar: Publish only occurs when the branch being built is named "master".

jaredpar:
image

PR #2690, opened 2019-07-18
Use VsVim margin for non-relative line numbers

### LINKED ISSUE #2687: Prevent displaying a line number for the phantom line
**Feature category**
- Check(Type 'x') any that Applicable item:
    - [X] Vim
    - [ ] Vim plugin

**Is your feature request related to a problem? Please describe.**

The way that Visual Studio displays line numbers is a constant source of confusion for even the most experienced developers. Traditionally an empty file is considered to have no lines but Visual Studio considers it to have one empty line. Because it is sometimes mandated that source files have no blank lines at the end of the file, developers are often frustrated by the presence of a seemingly blank line they can't get rid of. If they do manage to get rid of it, then tools like git consider the file (which does not end with a linebreak) to be malformed.

**Describe the solution you'd like**

The way to solve this problem is simply not to assign a line number in the line number margin for the phantom line. VsVim already does this when relative line numbers are in effect by displaying a `~` character instead of a line number for the phantom line. In Vim, a `~` in the first column is used to indicate screen space not occupied by the buffer. If VsVim used this convention for the phantom line when `:set number norelativenumber` was in effect, it would go a long way toward clarifying where the real end of the buffer was.

Here is a screenshot of VsVim with relative line numbers in effect showing the last line number agreeing with the reported number of lines and indicating that the phantom line is **past** the end of the buffer:

image

Here is a screenshot of Vim for the same file:

image

And finally, here is a screenshot of VsVim when `:set number norelativenumber` is in effect, showing the mislabelled phantom line:

image

Because as far as Vim (and Git, etc.) are concerned, this file does **not** have a line 1234 since it is only 1233 lines long.

ricksladkey:
#### Issues

- Closes #2687

#### Discussion

This change was harder than expected because of the synchronization between Vim and Visual Studio editor settings. Vim has a more nuanced set of line number features:

- `:set number norelativenumber` - similar to Visual Studio line numbers
- `:set nonumber relativenumber` - line numbers, but all of them relative to the current line
- `:set number relativenumber` - relative line numbers, but absolute for the current line

If `:set vsvim_useeditordefaults` is in effect (and it is by default), then Vim tries to adapt the user's language-specific Visual Studio editor settings. In the case of line numbers, they are either on or off. If they are on, which of the previous Vim settings correspond to that Visual Studio setting?

If a specific buffer should have line numbers (according to their Visual Studio settings), the most compatible and user-friendly solution is to turn on whichever line number settings the user specifically opted into in their rc file. And similarly, if there should be no line numbers (say for example, in a text file), then Vim line numbers should be completely turned off (`:set nonumber norelativenumber`).

So the complexity of this change is in adding:

- The capability to refer to rc file defaults when synchronizing settings
- The capability to synchronize multiple Vim settings for a single Visual Studio setting


ricksladkey:
Oops, this breaks the line number margin on VS2017. Will fix.

```
  <entry>
    <record>882</record>
    <time>2019/07/31 18:39:26.714</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.ArgumentException: An item with the same key has already been added.&#x000D;&#x000A;   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)&#x000D;&#x000A;   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.ContainerMargin.SubscribeToOptionChange(String optionName, Int32 marginIndex)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.ContainerMargin.AddMargins(IList`1 providers, List`1 oldMargins)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.ContainerMargin.Initialize()&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.LeftSelectionMarginProvider.CreateMargin(IWpfTextViewHost textViewHost, IWpfTextViewMargin containerMargin)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.ContainerMargin.&lt;AddMargins&gt;b__25_1(IWpfTextViewMarginProvider mp)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.InstantiateExtension[TExtension,TMetadata,TExtensionInstance](Object errorSource, Lazy`2 provider, Func`2 getter)&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)</description>
  </entry>```

PR #2693, opened 2019-07-20
Move to xunit 2.4.1

jaredpar:
Profiling the unit tests today I noticed we were still on xunit 2.3.1
which is nearly two years old. Decided to move us to a more recent
version.

This change also removes all of our checked in xunit binaries. The code
now relies completely on the binaries from the NuPkg files. This will
make upgrading easier in the future.


ricksladkey:
I swear I was at one time able to requeue a spurious failed check but I can no longer figure out how to do it. You can always push a trivial commit.

ricksladkey:
> I swear I was at one time able to requeue a spurious failed check but I can no longer figure out how to do it. You can always push a trivial commit.

I think I understand: as far as Azure Pipelines is concerned, the job did finish and and it succeeded. The option to requeue the job there is only available if the job failed. What failed was GitHub receiving the notification that it succeeded.

jaredpar:
> What failed was GitHub receiving the notification that it succeeded.

I think this may have been because I changed the VS2015 job to be required as the job was running. GitHub should be able to handle that (in my mind at least). But given the timing it seems likely that's what caused the problem here.

Issue #2698, opened 2019-07-25
Merge test files for same components


jaredpar:
**Describe the bug**
There are a number of components that have multiple test files associated with them:

1. CommandMode has CommandModeIntegrationTests and CommandProcessorLegacyTests
1. CommonOperations has CommonOperationsTests and CommonOperationsIntegrationTests
1. InsertMode has InsertModeIntegrationTests and InsertModeTests

The duplication here is generally because one test file relies heavily on mocking and the other by testing the product. Early on in VsVim mocking was preferred but as the product has evolved, and we've gotten closer to the true complexity of vim, we've strongly preferred testing the product. 

This has been true for some time but lately during major changes I've had to spend non-trivial time maintaining the mocking infrastructure for these components. It spurred me to take a look and see if mocking was truly needed for these files. It appears it is not and they can be merged to the more maintained ones quite easily. 

**Additional context**
This bug is not meant to track removing mocking from our tests, or even to suggest that we do so. Mocking is extremely helpful in testing certain aspects of the product that cannot be easily tested outside of Visual Studio: the shell layer as an example. There are just a few cases where it seems to make sense to test a component one way vs. two ways.

PR #2727, opened 2019-09-09
Clean up duplicate test types

jaredpar:
This largely completes #2698

The motivation here is that many of our components have two test files. For example `CommandMode` has `CommandModeIntegrationTests` and `CommandProcessorLegacyTests`. The motivations for this work are described in #2698. 

This actual change though has largely two parts:

1. Functional changes to VimCore to expose additional data to the testing layer. This was necessary to merge some of the mock based testing into our standard testing.
1. Code to merge the mocking tests into our standard testing.
1. Made a number of types `internal` because I was in the area and just noticed they should be.

Note: this PR doesn't have an urgency attached to it. This is just cleanup work I'd been meaning to do for some time now and finally got around to it.


Src/VimCore/CoreInterfaces.fs
```diff
@@ -5457,10 +5471,10 @@ and IVim =
     abstract RemoveVimBuffer: ITextView -> bool
 
     /// Get the IVimBuffer associated with the given ITextView
-    abstract TryGetVimBuffer: textView: ITextView * [<Out>] vimBuffer: IVimBuffer byref -> bool
+    abstract TryGetVimBuffer: textView: ITextView * vimBuffer: outref<IVimBuffer> -> bool
```
ricksladkey: Learn something new every day!

Test/VimCoreTest/InterpreterTest.cs
```diff
@@ -1587,7 +1587,7 @@ public sealed class WriteTest : InterpreterTest
 
             public WriteTest()
             {
-                VimHost.RunSaveTextAs = (text, filePath) =>
+                VimHost.SaveTextAsFunc = (text, filePath) =>
```
ricksladkey: Attention to detail is what makes VsVim so maintainable!

PR #2731, opened 2019-09-15
Delete a collapsed region correctly

### LINKED ISSUE #1936: Deleting a visual line selection of folded code omits the last fold
Given this code:

```
using System;
class Foo
{
    void Bar()
    {
        Console.WriteLine("Bar");
    }
    void Baz()
    {
        Console.WriteLine("Baz");
    }
}
```

If both methods are collapsed, selecting them using `Vj` and then deleting them (`d`) leaves the last folded method body in the file. Consequently, the only line of the last method that will be pasted later (`p`) will be its signature.

The result of `1GzM2Gza4GVjd`  thus is the following code. Deleting a single folded method is still possible when using `dd` (which behaves properly).

```
using System;
class Foo
{
    {
        Console.WriteLine("Baz");
    }
}
```

### LINKED ISSUE #2713: In 2019 deleting a collapsed region deletes too much
**Description**
Deleting a collapsed region seems to delete the collapsed region's number of lines, plus the collapsed area (minus 1).

**To Reproduce**
Sample code:

/*1*/	void ToDeleteWhenCollapsed()
/*2*/	{
/*3*/
/*4*/	}
/*5*/
/*6*/	void WillAlsoGetDeleted()
/*2*/	{ }
(Sorry about this sample, Markdown formatted my comments, and I don't know how to fix it.  The raw text should actually be valid code)

1.  Collapse lines 2 to 4
2.  Select lines 1-5 with v
3.  Press 'd'
4.  Lines 6 and 7 also get deleted
5. This does not occur when pressing 'c', only 'd'

**Environment**
- Visual Studio version: Professional 2019, version 16.2.3
- VsVim version: 2.7.0.0
- Programming Language: C#
- I noticed this in my environment, but I also could reproduce in an experimental environment (using Visual Studio's plugin SDK) that only had VSVim installed.

### LINKED ISSUE #2729: Deleting collapsed/folded #ifdef/#endif block deletes more lines of code
**Describe the bug**
I have folded blocks of #ifdef/#endif using the outlining in VS2015.
I selected the lines of the starting #ifdef with Shift+v, pressed % to select all lines and pressed d to delete those selected lines.
It seems, that only the visible two lines and source code below the folded block was deleted. The count of the deleted lines seems to be equal to those in the #ifdef/#endif block.

**To Reproduce**

1.  Collapse/fold the #ifdef/#endif block using the VS outlining
2.  Set the caret onto a #ifdef
3.  Press Shift+v to select complete lines
4.  Press % to select the complete #ifdef/#endif block
5.  Press d to delete the selected lines, from which only the first and last of the block are visible

**Expected behavior**
Expected is, that only the selected #ifdef/#endif block gets deleted.

- Visual Studio version: 2015 to 2019
- VsVim version: 2.7.0.0
- Programming Language: C++
    - [ ] ReSharper
    - [ ] Visual Assist

Note:
The following is written about contributing.
https://github.com/VsVim/VsVim/blob/master/CONTRIBUTING.md

General Keyboard configuration problems are described below.
https://github.com/VsVim/VsVim/issues/2527

TKumadaki:
**Issues**

- Fixes #1936
- Fixes #2713
- Fixes #2729


Test/VimCoreTest/CommonOperationsTest.cs
```diff
@@ -141,7 +141,7 @@ public void DeleteLines_OverFold()
             {
                 Create("cat", "dog", "bear", "fish", "tree");
                 _foldManager.CreateFold(_textView.GetLineRange(1, 2));
-                _operations.DeleteLines(_textBuffer.GetLine(0), 3, VimUtil.MissingRegisterName);
+                _operations.DeleteLines(_textBuffer.GetLine(0), 4, VimUtil.MissingRegisterName);
```
TKumadaki: I think that `_operations.DeleteLines` should specify the number of unfolded lines.

jaredpar:
Sorry for losing track of this. The change looks good but it looks like a conflict has come up. Can you resolve that and we'll get the change merged in?

TKumadaki:
I resolved a conflict. Is merging possible?

PR #2742, opened 2019-10-11
Fix local marks info

sqlkoala:
Hi guys,
I have a few more lines to add ;-)
Right now VSvim always shows the filename as info/name for a mark. Whereas vim shows the text at the mark for local marks. It is a small fix.

image


sqlkoala:
It is finally working. I changed/simplified the last steps in the test SpecialMarksAreSet() because it's too painful to  get all the text right after a paste in between. Please look at 
public void SpecialMarksAreSet()
in total, not just the diff.
What do you think?
Should I create an corresponding issue?

jaredpar:
Sorry for losing track of this. Thanks for fixing this up!

PR #2745, opened 2019-10-28
Fix search for Vim docs folder

gibwar:
If there are other folders in the default Vim installation folder that
do not have integers the lookup will fail and proceed to open the topic
online. If a folder returned that didn't have integers, the final
`|> int` expression would throw and eventually return `None`.

This changes the folder search routine to only consider folders that
match the regex `^vim(\d+)` and filter out any other folders that don't
match, like 'vimfiles'.

In my case, after an installation with the default installer left me
with the following directory setup:

```text
C:\Program Files (x86)\Vim
  |- vim81\
  |- vimfiles\
  |- _vimrc.
```

It also seems that `Directory.EnumerateDirectories` also returned
'vimfiles' first, even though natural sort would suggest otherwise. I'm
guessing that it returns it however it was written or however the
underlying file system returns it.

---
Please note this is my first time working in F# in _years_ so if there are better/more efficient ways to do this, I'm definitely open to improvements! Also, due to this not being on a critical path, I didn't see it fit to try and optimize the code or cache the regex. Because this relies on environment variables, static access, resolving and file system access, I didn't see it fit to try and write unit tests to cover these changes.

### DIFF
--- a/Src/VimCore/Interpreter_Interpreter.fs
+++ b/Src/VimCore/Interpreter_Interpreter.fs
@@ -1440,6 +1440,13 @@ type VimInterpreter
     member x.RunVimHelp (subject: string) = 
         let subject = subject.Replace("*", "star")
 
+        let extractFolderVersion = fun pathName ->
+            let folder = System.IO.Path.GetFileName(pathName)
+            let m = System.Text.RegularExpressions.Regex.Match(folder, "^vim(\d+)", System.Text.RegularExpressions.RegexOptions.IgnoreCase)
+            match m with
+            | x when x.Success -> Some(folder, m.Groups.[1].Value |> int)
+            | _ -> None
+
         // Function to find a vim installation folder
         let findVimFolder specialFolder =
             try
@@ -1454,14 +1461,7 @@ type VimInterpreter
                 if System.IO.Directory.Exists(vimFolder) then
                     let latest =
                         System.IO.Directory.EnumerateDirectories vimFolder
-                        |> Seq.map (fun pathName ->
-                            let folder = System.IO.Path.GetFileName(pathName)
-                            let m = System.Text.RegularExpressions.Regex.Match(folder, "^vim(\d+)", System.Text.RegularExpressions.RegexOptions.IgnoreCase)
-                            if m.Success then
-                                (folder, m.Groups.[1].Value |> int)
-                            else
-                                ("", 0))
-                        |> Seq.filter (fun (folder, version) -> folder <> "" && version <> 0)
+                        |> Seq.choose (extractFolderVersion)
                         |> Seq.sortByDescending (fun (_, version) -> version)
                         |> Seq.map (fun (folder, _) -> folder)
                         |> Seq.tryHead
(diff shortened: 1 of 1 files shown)


gibwar:
Also, unrelated, when downloading and compiling the project the first time I had problems with the F# project file and got this error:
```text
FSC : error FS0246: Unrecognized value 'Latest' for --langversion use --langversion:? for complete list
```

Running the recommended command provided the following output:
```text
% fsc --langversion:?
Supported language versions:
preview
default
latest
latestmajor
4.6
4.7 (Default)
```

Adding `<LangVersion>latest</LangVersion>` (lowercase L) to the project allowed it to compile. I did not include the change here as it was outside the scope of the fix and everyone else contributing and CI don't have the problem. I am on latest stable VS2019 16.3.6 Community (VisualStudio.16.Release/16.3.6+29418.71 with
Visual F# Tools 10.4 for F# 4.6   16.3.0-beta.19455.1+0422ff293bb2cc722fe5021b85ef50378a9af823.

Src/VimCore/Interpreter_Interpreter.fs
```diff
@@ -1454,14 +1454,16 @@ type VimInterpreter
...
+                        |> Seq.map (fun pathName ->
+                            let folder = System.IO.Path.GetFileName(pathName)
+                            let m = System.Text.RegularExpressions.Regex.Match(folder, "^vim(\d+)", System.Text.RegularExpressions.RegexOptions.IgnoreCase)
+                            if m.Success then
+                                (folder, m.Groups.[1].Value |> int)
+                            else
+                                ("", 0))
+                        |> Seq.filter (fun (folder, version) -> folder <> "" && version <> 0)
+                        |> Seq.sortByDescending (fun (_, version) -> version)
+                        |> Seq.map (fun (folder, _) -> folder)
```
ricksladkey: This is perfectly fine, but since you asked for a review, two small things come to mind.

1. In imperative languages it is common to use sentinels, i.e. values with special significance. For example, -1. In F# the pattern here might be to use `Some tuple` vs. `None` and then filter to the tuples.

2. For readability at a higher indent level, the lambda function could be factored out into a named function defined above and used once.

ricksladkey:
Thanks for noticing this issue and identifying the problem! It is a big help.

Src/VimCore/Interpreter_Interpreter.fs
gibwar: I think I understand what you mean in 1 and totally understand 2. When I attempted the Some/None approach, I ended up with a nested Option, so later on, `latest` was `string option option` instead of `string option`.

gibwar:
> Thanks for noticing this issue and identifying the problem! It is a big help.

Definitely! I've used VsVim since the 1.4 days and when I saw this come through I was excited and then bummed it ignored my installation. It was fun to jump in and debug!

Src/VimCore/Interpreter_Interpreter.fs
gibwar: Updated it by extracting it and changing it to use `Seq.choose`. Looks quite a bit more F# like and reads a bit easier. Thanks for the feedback and instruction @ricksladkey!

PR #2800, opened 2020-03-13
[VSMac] Allow tab key to complete snippets when in insert mode

### LINKED ISSUE #2786: VSMac: snippet completion is broken
Snippet completion is broken when this extension is enabled on Mac.

**To Reproduce**

1.  Type `ctor`
2.  Press tab to attempt to generate a class constructor

**Expected behavior**
The constructor should be generated

**Actual behavior**
The text `ctor` is left in the editor followed by the tab

**Screenshots**
snippetbroken

- Visual Studio version: Any Mac
- VsVim version: 2.8.0.2
- Programming Language: Any?
    - [ ] ReSharper
    - [ ] Visual Assist

nosami:
Fixes #2786

### DIFF
--- a/Src/VimMac/VimKeyProcessor.cs
+++ b/Src/VimMac/VimKeyProcessor.cs
@@ -28,18 +28,6 @@ namespace Vim.UI.Cocoa
         private readonly ISignatureHelpBroker _signatureHelpBroker;
         private readonly InlineRenameListenerFactory _inlineRenameListenerFactory;
 
-        public ITextBuffer TextBuffer
-        {
-            get { return VimBuffer.TextBuffer; }
-        }
-
-        public ITextView TextView
-        {
-            get { return VimBuffer.TextView; }
-        }
-
-        public bool ModeChanged { get; private set; }
-
         public VimKeyProcessor(
             IVimBuffer vimBuffer,
             IKeyUtil keyUtil,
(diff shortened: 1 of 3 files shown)


Src/VimMac/VimKeyProcessor.cs
```diff
@@ -143,13 +94,61 @@ public override void KeyDown(KeyEventArgs e)
...
+                return false;
+
+            if (_signatureHelpBroker.IsSignatureHelpActive(TextView))
+                return false;
+
+            if (_inlineRenameListenerFactory.InRename)
+                return false;
+
+            if (VimBuffer.Mode.ModeKind == ModeKind.Insert && e.Characters == "\t")
+                // Allow tab key to work for snippet completion
```
jaredpar: Is there a way to check to see if snippet completion is active?

nosami: I supposed that there must be a way of checking for this. The intellisense window doesn't necessarily need to be shown for this though.

I'll try and figure out how snippet completion works.

nosami: There are 2 different cases that we need to support here. 

The first is just before the snippet session is active (when the caret is at the end of `ctor` and tab key is pressed)

The second is during an active snippet session to tab (or back tab) between placeholders.

jaredpar: Does VS4Mac have an API to track when the snippet sessions are active? VS has one but it's pretty ancient. One of the old COM based APIs. Don't think that it will be in the shared layer. 

Actually on second thought I think this is all okay. This doesn't kill the key stroke, it just means it won't be processed by the vim layer. Instead it goes into VS4Mac where it gets handled.

nosami: I figured out how to handle the first case, but there isn't a good API to handle the 2nd case (although it would be possible with a lot of reflection hacking)

I think my original code is OK too. The only time I think it would affect anybody is if they wanted to map the tab key in insert mode to do something else (seems unlikely)

nosami:
Fixed in VsVim 2.8.0.4

PR #2803, opened 2020-03-21
Remove buffer from VimBuffer list when TextView is closed

nosami:
Remove buffer from VimBuffer list when TextView is closed

Fixes https://github.com/VsVim/VsVim/issues/2792https://github.com/VsVim/VsVim/issues/2792https://github.com/VsVim/VsVim/issues/2792

### DIFF
--- a/Src/VimCore/VimBuffer.fs
+++ b/Src/VimCore/VimBuffer.fs
@@ -246,6 +246,11 @@ type internal VimBuffer
         |> Observable.subscribe (fun args -> this.OnBeforeSave args.TextBuffer)
         |> _bag.Add
 
+        // Close the buffer when the TextView closes
+        _textView.Closed
+        |> Observable.subscribe (fun args -> this.Close())
+        |> _bag.Add
+
     member x.IsReadOnly
         with get() = _vim.VimHost.IsReadOnly _vimBufferData.TextBuffer
 
--- a/Src/VimMac/VimKeyProcessor.cs
+++ b/Src/VimMac/VimKeyProcessor.cs
@@ -79,10 +79,6 @@ namespace Vim.UI.Cocoa
                 {
                     handled = TryProcess(keyInput);
                 }
-                else
-                {
-                    handled = false;
-                }
             }
 
             VimTrace.TraceInfo("VimKeyProcessor::KeyDown Handled = {0}", handled);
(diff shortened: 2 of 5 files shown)


Src/VimMac/VimHost.cs
```diff
@@ -702,7 +701,18 @@ public bool SaveTextAs(string text, string filePath)
...
+            var shouldCreateVimBuffer = !textView.TextBuffer.ContentType.IsDebuggerWindow();
+
+            if (shouldCreateVimBuffer)
+                textView.Closed += TextView_Closed;
+
+            return shouldCreateVimBuffer;
+
+            void TextView_Closed(object sender, EventArgs e)
+            {
+                _vim.RemoveVimBuffer(textView);
```
jaredpar: I was puzzled at first why you needed to make this change. Had to dig a bit to figure out that this logic in the Windows version is being handled above the shared layer. 

https://github.com/VsVim/VsVim/blob/master/Src/VsVimShared/HostFactory.cs#L201

That implementation should be moved down to the shared layer at some point certainly. 

Rather than calling `_vim.RemoveBuffer` I think you should do essentially

```cs
if (_vim.TryGetBuffer(textView, out var vimBuffer))
   vimBuffer.Close()
```

Closing the `IVimBuffer` frees up other items in addition to removing from the buffer list. Looking now I don't think I should have made `IVimBuffer.RemoveBuffer` a public API. Or at least it should be making sure that `IVimBuffer` is closed.

nosami: Ok. Thanks. 

It looks like I could move this part `textView.Closed += (x, y) => vimBuffer.Close()` into the VimBuffer constructor

Src/VimCore/VimBuffer.fs
```diff
@@ -246,6 +246,11 @@ type internal VimBuffer
         |> Observable.subscribe (fun args -> this.OnBeforeSave args.TextBuffer)
         |> _bag.Add
 
+        // Close the buffer when the TextView closes
+        _textView.Closed
+        |> Observable.subscribe (fun args -> this.Close())
+        |> _bag.Add
```
jaredpar: I'm struggling to understand I didn't just do this in the first place ...

PR #2814, opened 2020-04-11
Support "update" and show which token failed to parse

wangzq:


jaredpar:
Looks like you have a couple of test failures here.

wangzq:
@jaredpar Looks like transient issue; but I made one more change in the command-utils.ps1 to show the msbuild exit code, please take a look. Thanks

PR #2817, opened 2020-04-13
Close signature help and completion window when escape is pressed

### LINKED ISSUE #2766: VsVim for Mac: exiting insert mode requires multiple presses of Esc
On my machine a reliable repro is this C# function call:

```
using System;
class Foo
{
    public static void Bar()
    {
        var x = "Hello";
        var y = "There";
        Console.WriteLine("{0}, {1}, {2}", Math.PI, x, y);
    }
}
```
Note:  This won't repro on non-working code, so make sure there are no squigglies in the above code before proceeding.

Position the cursor at the comma after `x` and type `i, y` to insert a comma, space, and a `y`.

Now hit Esc to return to normal mode.

Expect: To hit Esc just once.
Actual: Need to hit Esc three times before the cursor turns to a block.

@nosami

nosami:
When signature help was displayed and escape key was pressed, the
key was being handled by VSMac which didn't dismiss display windows.
Changed so that escape key is handled by Vim when signature help
is visible. This ensures that IDisplayWindowBroker.DismissDisplayWindows()
gets called.

Fixes #2766

### DIFF
--- a/Src/VimMac/Properties/AddinInfo.cs
+++ b/Src/VimMac/Properties/AddinInfo.cs
@@ -5,7 +5,7 @@ using Mono.Addins.Description;
 [assembly: Addin(
     "VsVim",
     Namespace = "Vim.Mac",
-    Version = "2.8.0.7"
+    Version = "2.8.0.8"
 )]
 
 [assembly: AddinName("VsVim")]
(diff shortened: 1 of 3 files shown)


nosami:
@jaredpar Did you get a chance to look at this yet? I would like to get it included in the next point release for Mac. Thanks!

PR #2831, opened 2020-05-26
Change `:make` to build project only

sankhesh:
This change supports native visual studio behavior of building
the current project only (and not the whole solution)
when `:make` or `:mak` is invoked.

When a shebang (`!`) is appended to the command i.e. `:make!`,
a full solution build is triggered.

### DIFF
--- a/Src/VsVimShared/VsVimHost.cs
+++ b/Src/VsVimShared/VsVimHost.cs
@@ -960,9 +960,16 @@ namespace Vim.VisualStudio
             return null;
         }
 
-        public override void Make(bool jumpToFirstError, string arguments)
+        public override void Make(bool buildSolution, string arguments)
         {
-            SafeExecuteCommand(null, "Build.BuildSolution");
+            if (buildSolution)
+            {
+                SafeExecuteCommand(null, "Build.BuildSolution");
+            }
+            else
+            {
+                SafeExecuteCommand(null, "Build.BuildOnlyProject");
+            }
         }
 
         public override bool TryGetFocusedTextView(out ITextView textView)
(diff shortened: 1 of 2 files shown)


Src/VsVimShared/VsVimHost.cs
```diff
@@ -960,9 +960,16 @@ private FSharpOption<ListItem> NavigateToLocation(NavigationKind navigationKind,
...
+        public override void Make(bool buildSolution, string arguments)
         {
-            SafeExecuteCommand(null, "Build.BuildSolution");
+            if (buildSolution)
+            {
+                SafeExecuteCommand(null, "Build.BuildSolution");
+            }
+            else
+            {
+                SafeExecuteCommand(null, "Build.BuildOnlyProject");
```
jaredpar, line 971: Why `Build.BuildOnlyProject` instead of `Build.BuildSelection`? The latter is generally what I use for "current project".

sankhesh: `BuildSelection` builds a project currently selected in the solution explorer. It does not necessarily have to be related to the file being edited currently. `BuildOnlyProject` builds the *current project* based on the buffer being edited.

IMHO, it might work for you because VS actively changes the selected project in the **SolutionExplorer** to match the current project. 

Note that it is also possible to select multiple projects in the solution explorer and call `BuildSelection` which is not something I'd want `:make` to do, given that vim is buffer-centric.

jaredpar: Thanks. I wasn't aware of this command before this.

PR #2867, opened 2020-12-27
Automate VSMac extension GitHub release

### LINKED ISSUE #2866: Automate the Visual Studio for Mac extension publish workflow using GitHub releases
Visual Studio for Mac is changing the way that extensions will be published [starting from January](https://docs.microsoft.com/en-us/visualstudio/mac/migrate-extensions?view=vsmac-2019)

The mechanism for publishing the extension will use GitHub releases. Should we use this repo for doing that?

We can probably take advantage of this by automating the publishing of the extension. This is something that is not currently possible.

What does the process look like for publishing the Windows extension? I guess it would make sense to try and make it as similar as possible.

nosami:
Pushing a tag with the following format will create a release containing the VSMac
extension file.

`v{Extension.Number}-vsm{Minimum Supported VSMac version}`

Any tag name that does not contain the string `vsm` will not create a release.

Fixes #2866

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,8 @@
 trigger:
-- master
+  branches:
+    include:
+    - master
+    - refs/tags/*
 
 pr:
 - dev/jaredpar/*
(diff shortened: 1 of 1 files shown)


nosami:
@jaredpar I think this can be merged now. What do you think?

jaredpar (approved):

PR #2889, opened 2021-05-05
[Mac] Fixes some issues with VimKeyProcessor

### LINKED ISSUE #2888: VsMac auto-format randomly switches off with VsVim 2.8.0.11
Hi, since updating to VsVim 2.8.0.11 Visual Studio's auto-formatting stops working randomly. 
When this happens e.g. typing a close brace will not format the code inside. Opening braces on a new line appear at the wrong tab stop, and so on.

I'm trying to roll back to an earlier VsVim to test whether the plugin is the cause.
Unfortunately the VS folks have removed the .mpack archive at https://addins.monodevelop.com/Project/Index/403

Any idea where I might find an earlier .mpack?
Thanks!

EDIT: Managed to fish out 2.8.0.7 from my Time Machine backup and things work again. So the problem is very likely in the current version of the plugin.

nosami:
The last fix broke auto-indentation and signature help.

This code is now simpler with less hacks and seems to work much better.

Tested the following and didn't find any issues.

- inoremap
- auto indentation when pressing } or ;
- Insert mode insertion followed by '.'
- Signature help
- Snippet completion
- Intellisense
- Inline rename
- Macro recording involving insert mode

Fixes #2888

### DIFF
--- a/Src/VimMac/KeyMappingTimeoutHandler.cs
+++ b/Src/VimMac/KeyMappingTimeoutHandler.cs
@@ -1,0 +1,139 @@
+﻿using System;
+using System.ComponentModel.Composition;
+using System.Windows.Threading;
+
+namespace Vim.UI.Wpf.Implementation.Misc
+{
+    /// <summary>
+    /// This class is responsible for handling the timeout of key mappings for a given
+    /// IVimBuffer.  If the timeout occurs before the key mapping is completed then the
+    /// keys should just be played as normal
+    /// </summary>
+    [Export(typeof(IVimBufferCreationListener))]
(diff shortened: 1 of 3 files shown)


nosami:
I just noticed another issue with inoremap that I need to fix somehow. `inoremap fd <esc>` works when `fd` is typed, but `f` followed by any other character swallows the first `f`

nosami:
I think I might have finally nailed all the KeyProcessor issues.

The only limitation I see now is that Intellisense is disabled while a macro is being recorded. This seems like it might be a fairly sensible limitation though as I can imagine this would be pretty tricky to implement correctly. @jaredpar does VsVim on Windows handle this?

jaredpar (approved):

Issue #2886, opened 2021-04-24
[VSMac] Caret width doesn't match text width when zooming in or out


nosami:
**Describe the bug**
In VSMac, The caret width doesn't match the text width after zooming. The caret width retains the width of the text before the zoom.

**To Reproduce**

1.  Zoom text with `command =` or shrink with `command -`
2.  Notice caret width

**Expected behavior**
The caret width should grow or shrink when zooming in or out

**Screenshots**
<img width="556" alt="image" src="https://user-images.githubusercontent.com/667194/115953741-16af8600-a4e5-11eb-90c8-a79d48f1d08a.png">

**Environment (please complete the following information):**

- Visual Studio version: VSMac

PR #2890, opened 2021-05-06
[Mac] Resize caret when zoom level changes

nosami:
Also cleans up eventhandlers

Fixes https://github.com/VsVim/VsVim/issues/2886

### DIFF
--- a/Src/VimMac/CaretUtil.cs
+++ b/Src/VimMac/CaretUtil.cs
@@ -10,6 +10,7 @@ namespace Vim.Mac
     internal class CaretUtil : IVimBufferCreationListener
     {
         private readonly InlineRenameListenerFactory _inlineRenameListenerFactory;
+        private IVimBuffer _vimBuffer;
 
         [ImportingConstructor]
         public CaretUtil(InlineRenameListenerFactory inlineRenameListenerFactory)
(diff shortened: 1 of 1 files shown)


Src/VimMac/CaretUtil.cs
```diff
@@ -46,9 +47,39 @@ private void SetCaret(IVimBuffer vimBuffer)
 
         void IVimBufferCreationListener.VimBufferCreated(IVimBuffer vimBuffer)
         {
-            SetCaret(vimBuffer);
-            vimBuffer.SwitchedMode += (_,__) => SetCaret(vimBuffer);
-            _inlineRenameListenerFactory.RenameUtil.IsRenameActiveChanged += (_, __) => SetCaret(vimBuffer);
+            _vimBuffer = vimBuffer;
```
jaredpar, line 50: If multiple IVimBuffer are created they will overwrite this variable. Believe that causes the later event handlers to potentially operate on the wrong instance. Basically is this okay if two IVimBuffer are created and then the first one is closed. Won’t that cause the event handlers to make changes to second one?

nosami: Ah, yes. Good spot!

jaredpar (approved):

PR #2906, opened 2021-07-04
Move to shared projects 

### LINKED ISSUE #2900: Moving to a shared project approach
The goal of VsVim is to support as many different Visual Studio versions as possible. Up until this point VsVim has done this by creating a single VSIX that will work in a large range of Visual Studio versions. In order to support behaviors that are specific to a given Visual Studio version, or range of versions, it has a simple plugin system (the VsSpecific projects). 

This has worked really well over the years. Unfortunately beginning with Visual Studio 2022 this model is no longer going to be possible. The move to Visual Studio 2022 both changes the target architecture of the binaries (x86 to x64) as well as the references needed to access Visual Studio APIS. The latter is the most problematic because the new interop references in 2022 represent a breaking change from all previous versions of Visual Studio. Hence a single binary which uses these APIs can no longer load in both VS 2022 and VS 2019. 

Keeping the existing model would mean taking the VsSpecific plugin system to the extreme. Essentially moving all of the code out of VsVimShared and into VsSpecific. That gets really complicated because MEF registration in VsSpecific layer isn't possible. Hence it would require rather substantial changes to how much of the code acquires components. Essentially it is fighting against the current here. 

After thinking about this for a while I believe the right approach is to embrace the way Visual Studio is recommending developers author their extensions. Essentially move to a model where you have one VSIX per Visual Studio version. They recommend using a shared project for the core logic of the extension and having a VSIX per Visual Studio version which has a source.extension.manifest file + a reference to the shared project. 

https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022#use-shared-projects-for-multi-targeting

I did some experiments with this over in the [dev/jaredpar/shared](https://github.com/VsVim/VsVim/tree/dev/jaredpar%2Fshared) branch. Was able to move to this model with minimal disruption to the code base. There is still some work to be done there but overall feel pretty good about the development experience. It also means that it should make it easier to target newer versions of assemblies and get better version specific behavior without having to be so indirect with plugins.

The biggest downside about this approach is having to manage 2-3 VSIX for upload verses a single one. My rough plan is that I will structure the VSIX as follows:

- VsVim: the existing entry in the Visual Studio gallery will always target the latest RTM version of Visual Studio 
- VsVim for Visual Studio 2019: this will be the structure of VSIX for the earlier releases of Visual Studio (replace the year for each version). 

This plan should not impact the Vs4Mac work as that exists in a layer parallel to the Vs4Windows work. 

Interested in hearing any thoughts, pushback, etc ... on this. Hoping to have a draft PR for the changes in the next week or so. 

Likely at the same time we move to the shared model I will drop support for some older Visual Studio versions. This is just standard when a new version of Visual Studio comes out. Hence it's not specific to this change, it's just more convenient to do it as a part of this change.

jaredpar:
Want to apologize up front as this is a big change and likely not reviewable as a single item. The individual commits should be reviewable should one want to go that route. I tried a couple of approaches to keep this diff down and tried to think through a few ways to break this up better into more reviewable chunks but didn't find any good paths here. Moving to shared projects is essentially undoing 10 years of how VsVim was built and tested and hence the change is big. 

Rather than reviewing may be better to look at the result of the PR here. Those can be summarized as follows:

- Moved all of the projects above `VimCore` into shared source projects (this is the recommended method for [developing VS extensions going forward](https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022#use-shared-projects-for-multi-targeting)). 
- Moved from a single `VsVim` project to one per supported Visual Studio version: `VsVim2017` and`VsVim2019`
- Moved from a three different test projects for testing `VimWpf` and `VsVim` to a single one per supported Visual Studio Version: `VsVim2017Test` and `VsVim2019Test`
- Removed all of the projects that were meant to support dynamic loading in Visual Studio (`VsSpecific`, `VsInterfaces`, `Vs2017` and `Vs2019`) and merged them into `VsVimShared`. 

The result of this PR gives us functional extensions for VS2017 and VS2019. It mostly removes support for VS2015. I will send a follow up PR that completely removes that support (doing it in this PR added a lot more churn). After that I'm going to come back and add the support for VS2022 in as well. 

closes #2900


nosami (approved):
I'm not the best person to review this, but I'm happy if the macOS build passes

PR #2908, opened 2021-07-06
Remove VS 2015 support

jaredpar:
This removes the remaining pieces of Visual Studio 2015 support from the code base.


jaredpar:
This change is 52 files but it's essentially:

- Deleting the `References\Vs2015` folder which had a lot of files in it 
- Removing all of the `#if VS_SPECIFIC_2015` which is effectively deleting dead code

nosami (approved):

PR #2917, opened 2021-07-20
VS 2022 Support

jaredpar:
This commit adds the initial support for VS 2022 into the main tree.
This is pretty straight forward with the shared project approach now.
There was a single test failure in folds that I need to track down but
that's about the only issue I hit.

A lot of credit goes to @vivlimmsft in commit
17b98f2. That is the work that got
VsVim into VS2022 Preview 1. The work involved convinced me that the
shared project approach was necessary going forward. This PR is
effectively doing that and completing the Preview 1 work.


Kethku:
Super excited for this as not having a working extension in 2022 has been defacto blocking my ability to work in vs.

Some things I noticed:
- the cursor feels thinner when in insert mode. I'm guessing something changed...
- the options window for setting keybindings is weird now. It kinda flickers and its hard to click buttons or w/e

If you'd like I can raise separate issues for these, just figured I'd let you know here.

Kethku:
Also oddly immediately after installing, things worked fine, but then after a restart, I can't open text documents with the extension enabled. VS freezes and doesn't recover

Kethku:
Weirdly reinstalling seems to cause it to work again until I restart

jaredpar:
@Kethku 

Sorry, been out for the last week. Just getting back to things 

> the cursor feels thinner when in insert mode. I'm guessing something changed...

It's possible this has changed but it wouldn't be due to VsVim. For insert mode VsVim actually just lets Visual Studio draw the caret. They did revamp the UI a bit though so it's possible they made changes here. 

> the options window for setting keybindings is weird now. It kinda flickers and its hard to click buttons or w/e

That one I hadn't noticed. It's likely this is still a VS issue vs. a VsVim one but please file a tracking issue so I don't lose track of it.

PR #2925, opened 2021-08-17
disable mode change (Visual Studio -> VIM) after GoToDefinition().

whpark:


nosami:
Is the issue here that VsVim is going into Normal mode when you GoToDefinition to a new file? If so, I think that is expected.

What happens if you GoToDefinition into the same file when in Insert mode?

whpark:
oh, I'm sorry to bother you, I just thought it would stay disabled.

whpark:
close

PR #2938, opened 2021-11-05
Work around IntelliCode handling Escape

jaredpar:
IntelliCode appears to to be handling Escape in more places now (also
possible I just didn't fully catch all of them initially). The most
problematic case is that it's handling Escape even when there is no full
line completion displayed. That is breaking our Insert mode experience
as Escape is dismissing IntelliSense but not returning to Normal mode.

After some digging around I found a better condition to determine when
IntelliCode is handling IntelliSense keystrokes. This seems to catch all
of the big cases I could find.

### DIFF
--- a/Src/VsVimShared/Implementation/IntelliCode/IntelliCodeCommandTarget.cs
+++ b/Src/VsVimShared/Implementation/IntelliCode/IntelliCodeCommandTarget.cs
@@ -13,16 +13,18 @@ using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Text.Editor;
 using System.Reflection;
 using System.Diagnostics;
+using Microsoft.VisualStudio.Text;
 
 namespace Vim.VisualStudio.Implementation.IntelliCode
 {
     internal sealed class IntelliCodeCommandTarget : ICommandTarget
     {
-        private object? _cascadingCompletionSource;
-        private bool _lookedForCascadingCompletionSource;
+        private (object CascadingCompletionSource, FieldInfo CurrentPrediction, FieldInfo CompletionState)? _data;
+        private bool _lookedForData;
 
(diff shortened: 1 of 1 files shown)


Src/VsVimShared/Implementation/IntelliCode/IntelliCodeCommandTarget.cs
```diff
@@ -13,44 +13,93 @@
...
+        private bool IsIntelliCodeHandlingEscape()
+        {
+            EnsureLookedForData();
+            if (_data is { } data)
+            {
+                try
+                {
+                    var completionState = data.CompletionState.GetValue(data.CascadingCompletionSource);
+                    var currentPrediction = data.CurrentPrediction.GetValue(data.CascadingCompletionSource);
+                    return completionState is object || currentPrediction is object;
```
333fred, line 93: Do you want to trace the result here so that if this condition stops being successful without exceptions, you can still trace the failure?

333fred, line 93:
```suggestion
                    return completionState is not null || currentPrediction is not null;

```
Someone should write a blog post titled "`is object` considered harmful"

jaredpar: The expectation is this returns `false` far more often than it returns `true`. Consider that for the lifetime of a buffer every `QueryStatus` around `Escape` will end up here (even if it's for normal mode, key mapping, etc ...). 

The traces already will show us a miss here. Essentially when we see the KeyUp message for escape but not KeyDown then we know it was eaten.

jaredpar: Nah, someone already tweeted that it's the best option 😄

PR #2963, opened 2022-02-28
[VSMac] Switch macOS build to net6

nosami:
Since some common types got moved out of the `System` namespace in `net6` (`nfloat` etc), there are now some exceptions that causes the extension to not function correctly in VSMac in some cases.

The only way to resolve these issues are to build the extension against the same set of assemblies as the preview version of VSMac. We've been building against an old version (8.4) for quite some time now. This also implies that this extension will need to use `net6` as the TargetFramework.

### DIFF
--- a/Src/VimCore/FSharpUtil.fs
+++ b/Src/VimCore/FSharpUtil.fs
@@ -584,7 +584,7 @@ module CharUtil =
     let IsTagNameChar x = System.Char.IsLetterOrDigit(x) || x = ':' || x = '.' || x = '_' || x = '-'
     let IsFileNameChar x = IsTagNameChar x
     let IsPrintable x =
-        let category = CharUnicodeInfo.GetUnicodeCategory(x)
+        let category = CharUnicodeInfo.GetUnicodeCategory(x: char)
         PrintableCategories.Contains category
     let ToLower x = System.Char.ToLower(x)
     let ToUpper x = System.Char.ToUpper(x)
@@ -970,7 +970,7 @@ module internal SystemUtil =
         else
             Some text
 
-    let EnsureRooted currentDirectory text = 
+    let EnsureRooted (currentDirectory: string) (text: string) = 
         if Path.IsPathRooted text || not (Path.IsPathRooted currentDirectory) then
             text
         else
@@ -980,7 +980,7 @@ module internal SystemUtil =
     /// isn't rooted it will be rooted inside of 'currentDirectory'
     ///
     /// This method can throw when provided paths with invalid path characters.
-    let ResolveVimPath currentDirectory text = 
+    let ResolveVimPath (currentDirectory: string) text = 
         match text with
         | "." -> currentDirectory
         | ".." -> Path.GetPathRoot currentDirectory
--- a/Src/VimCore/VimCore.fsproj
+++ b/Src/VimCore/VimCore.fsproj
@@ -3,7 +3,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Vim.Core</RootNamespace>
     <AssemblyName>Vim.Core</AssemblyName>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <OtherFlags>--standalone</OtherFlags>
     <NoWarn>2011</NoWarn>
     <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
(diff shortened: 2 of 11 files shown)


Src/VimCore/VimCore.fsproj
```diff
@@ -3,7 +3,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Vim.Core</RootNamespace>
     <AssemblyName>Vim.Core</AssemblyName>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
nosami, line 6: Will need to ifdef this change

nuget.config
```diff
@@ -4,5 +4,6 @@
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="vssdk" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vssdk/nuget/v3/index.json" />
+    <add key="vsmacsdk" value="https://devdiv.pkgs.visualstudio.com/DevDiv/_packaging/VSMac/nuget/v3/index.json" />
```
nosami, line 7: This NuGet source is not yet public but `Microsoft.VisualStudioMac.Sdk` should be available from `nuget.org` soon.

The package contains some of the core assemblies used in VSMac.

Src/VimCore/VimCore.fsproj
jaredpar: Should be able to just multi-target here between `net472` and `net6`

Src/VimMac/RelativeLineNumbers/CocoaExtensions.cs
```diff
@@ -11,6 +11,7 @@
 using CoreGraphics;
 using CoreText;
 using Foundation;
+using ObjCRuntime;
```
jaredpar, line 14: Why is this necessary? Did types move around to a new namespace?

nosami: Yes. System.nfloat got moved to ObjCRuntime.nfloat

VsVim.sln
```diff
@@ -30,7 +30,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Documentation", "Documentat
 		Documentation\Supported Features.md = Documentation\Supported Features.md
 	EndProjectSection
 EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "VimCore", "Src\VimCore\VimCore.fsproj", "{333D15E0-96F8-4B87-8B03-467220EED275}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "VimCore", "Src\VimCore\VimCore.fsproj", "{333D15E0-96F8-4B87-8B03-467220EED275}"
```
nosami, line 33: Not sure why VSMac changed the project type guid here. I probably should just revert this change.

jaredpar: Yeah unsure why it did this. SLN files are mostly a mystery to me though.

nosami: They are a mystery to everyone I think. Only the machines can understand them :)

Src/VimCore/FSharpUtil.fs
```diff
@@ -584,7 +584,7 @@ module CharUtil =
     let IsTagNameChar x = System.Char.IsLetterOrDigit(x) || x = ':' || x = '.' || x = '_' || x = '-'
     let IsFileNameChar x = IsTagNameChar x
     let IsPrintable x =
-        let category = CharUnicodeInfo.GetUnicodeCategory(x)
+        let category = CharUnicodeInfo.GetUnicodeCategory(x: char)
```
jaredpar, line 587: Were the type's necessary here and forced by an F# change in inference?

nosami: although now I just looked at that code again, it would probably be neater to add the type annotation to `IsPrintable x` instead.

jaredpar: Either way is fine. I'm just always curious if it's a bug / broken behavior the code was relying on in F# or if there was a diff reason. Found a few times where I had relied on F# bugs.

nosami:
I think this is pretty much ready now. As far as I can tell/guess, this build will pass as soon as we have a public URL for VSMac preview 7 to build against.

jaredpar (approved):

nosami:
There are tons of warnings that look like this

```
/Users/jasonimison/src/vsvim/Src/VimMac/VimHost.cs(622,33): warning CA1416: This call site is reachable on all platforms. 'Process.StandardInput' is only supported on: 'macOS/OSX'. [/Users/jasonimison/src/vsvim/Src/VimMac/VimMac.csproj]
```

I can make the warnings go away by targeting `net6.0-macos` instead of `net6` but then it is no longer buildable on Windows. I think I just need to ignore these warnings for now.

PR #2973, opened 2022-06-21
Push MS.VS.Threading into VimCore

jaredpar:
This moves the reference to Microsoft.VisualStudio.Threading down into VimCore. This assembly is now very central to how the editor works and having it not present in VimCore was causing issues reconciling some warnings / errors I was running into. 

I asked around this should not present a problem in Vs4Mac. This assembly is used there and `JoinableTaskContext` can be used in a MEF composition. But wanted @nosami to see this change going through to double check.


nosami:
This looks fine to me. VSMac already uses both `Microsoft.VisualStudio.Threading` and `JoinableTaskContext`

nosami (approved):

PR #2979, opened 2022-06-22
Mac OS Publishing

jaredpar:


jaredpar:
@nosami this PR sets up the MacOS build, publish and release steps for GitHub Actions. I've verified with a previous run that we're uploading the mpack file successfully. Can see that [here](https://github.com/VsVim/VsVim/runs/7007789384?check_suite_focus=true). 

I'm unsure how to verify the release steps though. Is there a way to create a dummy release where I can verify the changes and then just delete it after?

nosami:
> @nosami this PR sets up the MacOS build, publish and release steps for GitHub Actions. I've verified with a previous run that we're uploading the mpack file successfully. Can see that [here](https://github.com/VsVim/VsVim/runs/7007789384?check_suite_focus=true).
> 
> I'm unsure how to verify the release steps though. Is there a way to create a dummy release where I can verify the changes and then just delete it after?

The way that I tested originally was to mark the release as "Draft". Once I checked that everything else was working I switched the script to stop marking it as draft.

nosami:
Looks about right though :)

nosami:
One more thing that I forgot. You need to name a tag in a specific format before it will publish the artefacts to the Release page.

The last tag was - `v2.8.0.15-vsm17.0`. Just need to change the version number (manually) in the tag name to match the version number in AddinInfo.cs

jaredpar:
image

jaredpar pushed the wrong branch. Git tried to save him but he used his Jedi powers and `--force` a bad decision.

jaredpar:
@nosami okay tested out publishing and believe I have it working now. Here is the draft release that I published. 

https://github.com/VsVim/VsVim/releases/tag/untagged-fcc4471be4ed62ddc9fa

Does that have everything you need in it? 

Note: The title and tag have "test" in them just because i didn't want to accidentally overwrite any existing releases. Will remove before merge.

nosami:
Looks good. Thanks for doing that 👍

I think it should be fine now. If it needs anything changing, I can pick it up.

PR #3025, opened 2022-11-01
Use next bracket text object if cursor is not currently within a bracket text object

JochenBaier:
Emulate the behavior of Vim (compared with gvim 9.0.57 on Windows) to use the next bracket text object when the cursor is not inside a bracket text object. The next bracket text object is searched using the opening bracket character.

Example (cursor is italic, selected text is bold):
_c_at (fish) dog -> vib -> cat (**fish**) dog
_c_at (fish) dog -> vab -> cat **(fish)** dog

Please double check PR, this is my first contact with VsVim source code and my first contact with F#.

### DIFF
--- a/Src/VimCore/CommandUtil.fs
+++ b/Src/VimCore/CommandUtil.fs
@@ -1954,6 +1954,18 @@ type internal CommandUtil
             else
                 CommandResult.Completed ModeSwitch.NoSwitch
 
+    /// Search for next open bracket after current caret position and return bracket text object if found
+    member x.GetNextBracketTextObject(motion: Motion) =
+        match motion with
+        | (Motion.AllBlock blockKind | Motion.InnerBlock blockKind) -> 
+            let startChar, _ = blockKind.Characters
+            let searchdata = SearchData(startChar.ToString(), SearchPath.Forward, false)
+            let searchResult = _vimTextBuffer.Vim.SearchService.FindNextPattern x.CaretPoint searchdata _wordUtil.SnapshotWordNavigator 1 
+            match searchResult with
+            | SearchResult.Found  (_, span, _, _)-> _motionUtil.GetTextObject motion span.Start
+            |_ -> None
+        |_ -> None
+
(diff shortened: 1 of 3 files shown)


lawrence-laz:
As I understand this currently supports brackets only, right? Could this be improved to cover `vi"`, `vi>`, `vi'`, etc. without too much effort?  Related to #3046 .

JochenBaier:
> As I understand this currently supports brackets only, right? Could this be improved to cover `vi"`, `vi>`, `vi'`, etc. without too much effort? Related to #3046 .

" and ' is already supported without the patch
<> is only supported within

nosami:
This PR seems very useful. Did you look to see how `vi"` works today?

I'm not totally familiar with how this code works but it seems like it is using `VisualCommand.MoveCaretToTextObject` to move the caret to the first `"`.
It seems like it would be a smaller change to make the existing code work for `(` and `<`

pawelminkina:
Is that going to be merged? This is very useful feature to have to be able to use "vi)".

issmith962:
> This PR seems very useful. Did you look to see how `vi"` works today?
> 
> I'm not totally familiar with how this code works but it seems like it is using `VisualCommand.MoveCaretToTextObject` to move the caret to the first `"`. It seems like it would be a smaller change to make the existing code work for `(` and `<`

You may be right about this, but figured I would just chime in to point out an important difference. In the modern versions of vim, commands to move to a '"' pair are current-line-only for both the starting and ending '"'. The bracket selections can both start and end on lines that are not the current line. Therefore, they do need a bit more handling than the '"' selections. 

(as an aside, I've built the extension from this branch + installed on my machine, will be testing quite a bit for the next few days since I use VsVim at work. Will add comments to this PR if I find any issues etc)

issmith962:
In case it's helpful: I've been test-driving this branch at my day job for 1.5 months now, and it's been working perfectly. The new bracket functionality works exactly as expected, all unit tests for the project are passing, and I haven't noticed any bugs.

PR #3077, opened 2023-11-08
Remove SeqUtil.filterToSome

nosami:
Slight refactoring.

Seq.choose f is equivalent to Seq.map f |> SeqUtil.filterToSome


nosami:
Given that the build used to work but now doesn't, I'm going to go out on a limb and say that this build error is an MSBuild bug.

jaredpar (approved):

PR #3078, opened 2023-11-08
Remove VSSDK targets

nosami:
It seems these targets are not required and are currently causing build issues.
(See build log for https://github.com/VsVim/VsVim/pull/3077)


nosami:
So, the targets _are_ needed!

PR #3130, opened 2025-01-21
Add IntelliCodeExtensionAdapter class implementation to avoid vim to show on difference controls for edits.

osobeso:
In VS Copilot Chat features that insert code generated by the LLM we have an embedded view similar to a `Peek View` that allows us to display the differences that will be inserted, and allows the user to either refine on those using the embedded chat input or directly modifying the diffs, or accept/reject them when they are part of copilot edits experience.

Either way these difference views are only meant to display the differences but should not display a Vim buffer.

This change allows us to exclude these specific views from construct a Vim buffer.

Note: I added this under the IntelliCode folder because there was already a folder and namespace for IntelliCode, and we are the same team behind copilot for VS.

Before:
image
(Note the vim buffer at the bottom of the element covering the last line of the difference element)

After:
image
(All lines are now visible)

### DIFF
--- a/Src/VsVimShared/Implementation/IntelliCode/IntelliCodeExtensionAdapter.cs
+++ b/Src/VsVimShared/Implementation/IntelliCode/IntelliCodeExtensionAdapter.cs
@@ -1,0 +1,29 @@
+using Microsoft.VisualStudio.Text.Editor;
+using System.ComponentModel.Composition;
+
+namespace Vim.VisualStudio.Implementation.IntelliCode;
+
+[Export(typeof(IExtensionAdapter))]
+internal sealed class IntelliCodeExtensionAdapter : VimExtensionAdapter
+{
+    private const string DifferenceViewerWithAdaptersRole = nameof(DifferenceViewerWithAdaptersRole);
+
+    [ImportingConstructor]
+    internal IntelliCodeExtensionAdapter()
+    {
+    }
+
+    // Suppress VsVim in the Embedded Difference Control elements.
+    protected override bool ShouldCreateVimBuffer(ITextView textView) =>
+        !IsEmbeddedDifferenceControlWindow(textView);
+
+    private static bool IsEmbeddedDifferenceControlWindow(ITextView textView)
+    {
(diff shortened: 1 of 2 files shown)


Src/VsVimShared/VsVimShared.projitems
```diff
@@ -40,6 +40,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\InlineRename\InlineRenameListenerFactory.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\InlineRename\InlineRenameUtil.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\IntelliCode\IntelliCodeCommandTarget.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)implementation\intellicode\IntelliCodeExtensionAdapter.cs" />
```
vivlimmsft, line 43: nit: the capitalization of this line is different rom the others

osobeso: odd behavior, this is auto-generated code 🤷
I'll manually fix.

Src/VsVimShared/Implementation/IntelliCode/IntelliCodeExtensionAdapter.cs
```diff
@@ -0,0 +1,29 @@
+using Microsoft.VisualStudio.Text.Editor;
+using System.ComponentModel.Composition;
+
+namespace Vim.VisualStudio.Implementation.IntelliCode;
+
+[Export(typeof(IExtensionAdapter))]
+internal sealed class IntelliCodeExtensionAdapter : VimExtensionAdapter
+{
+    internal const string DifferenceViewerWithAdaptersRole = nameof(DifferenceViewerWithAdaptersRole);
```
vivlimmsft, line 9: if this PR merges, please add a comment to our implementation that links to this PR & says vsvim is depending on this name.

jaredpar (approved):

osobeso:
@jaredpar Hi Jared. would appreciate your help with merging this PR.
Thanks!
If there is anything I can help with, please let me know.

PR #3152, opened 2025-08-02
Fix CI: Update from deprecated upload-artifact action & switch to macOS-latest

vivlim:
The GitHub action runs were failing for two reasons:
- `actions/upload-artifact@v3` was deprecated. I upgraded to `v4`, there were no other changes needed for that migration
- `macOS-11` runners don't seem to be available anymore. I changed it to `macOS-latest`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,7 @@ jobs:
 
   macos-build:
     name: MacOS Build and Publish
-    runs-on: macOS-11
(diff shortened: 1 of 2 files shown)


vivlim:
@jaredpar, can this be merged?

Seems like it might result in a new version being published, I don't have a sense of whether there's anything risky that needs specific testing.

jaredpar (approved):

PR #3156, opened 2025-09-10
Update vsixmanifest to support Visual Studio 2026

vivlim:
Updating the manifest so that VsVim can be installed on [Visual Studio 2026](https://devblogs.microsoft.com/visualstudio/visual-studio-2026-insiders-is-here/).

Fortunately this was *much* simpler than 2019 -> 2022 was. (#2917)

### DIFF
--- a/Src/VsVim2022/source.extension.vsixmanifest
+++ b/Src/VsVim2022/source.extension.vsixmanifest
@@ -12,28 +12,28 @@
     <Tags>vsvim</Tags>
   </Metadata>
   <Installation>
-    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[17.0,18.0)">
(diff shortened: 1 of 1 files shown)


vivlim:
Seems like this isn't actually necessary, install isn't being blocked like it was when I last tried

vivlim:
Hmm. I don’t remember for sure, but the fact you’re hitting that seems like a good reason to update the manifest after all.

vivlim:
Confirmed this is necessary, I saw the same message when trying to install the vsix attached to https://github.com/VsVim/VsVim/actions/runs/17704228699.

I'm not sure what I did before to install it. Might have gone through the extension manager somehow, not sure.

micrb:
Hi, is there somewhere I could get the latest build? Looks like the artifacts have expired for this PR.

Issue #3162, opened 2025-12-07
VsVim builds are not being published


vivlim:
When a build completes it is not actually published to the vsix marketplace *or* the Open Vsix Gallery (as the readme currently says).

As a manual workaround you can go to this repo's [Actions](https://github.com/VsVim/VsVim/actions) page, click the most recent VsVim CI run (currently [this](https://github.com/VsVim/VsVim/actions/runs/19258677705)), then click on the 'Vsix' artifact towards the bottom of the page to download a zip containing the vsix. You can then extract the zip and double-click the vsix to install.

devcayed:
Could someone re-run the action? The artifacts produced in the referenced run are expired by now and can no longer be downloaded.
Thanks!

BPowell76:
Artifacts still expired. Why are there some many builds for Mac, but nothing for Windows? In the Windows build CI pipeline, it looks like there is bad code.

sgrassie:
Any movement on this at all?

vivlim:
The build break was fixed in #3173 and you can now grab a build from this run: https://github.com/VsVim/VsVim/actions/runs/30423654076

PR #3163, opened 2025-12-07
Overhaul build pipeline, introduce git-derived version numbering, and publish a private gallery directly from GH artifacts

vivlim:
This PR:

- Publishes CI builds as artifacts without wrapping them in a zip ([example from my fork](https://github.com/vivlim/VsVim/releases/tag/ci-20251017-2.11.3-vsix))
- Adds Nerdbank.GitVersioning and uses it to derive version numbers from git history, resulting in each commit's build having a different version number - no more having to manually bump version numbers
- Removes the Mac build from the pipeline to reduce complexity and maintenance overhead, since VS for Mac is EOL.
- Generates and publishes a feed.xml file that can be [used in VS as a private gallery to subscribe to CI builds](https://learn.microsoft.com/en-us/visualstudio/extensibility/private-galleries?view=visualstudio).
  - If you would like to try this and install builds from my fork, you can try adding this feed url:
```
https://github.com/vivlim/VsVim/releases/latest/download/feed.xml
```

Regarding #3162, having this private gallery is a significant usability improvement for installing builds directly from the master branch, but it doesn't replace the need to publish to the marketplace. I see it as a complementary addition (... and also, I wanted to see if it would work)

I've worked on this PR intermittently over a few months so I'll also need to review it myself to make sure I didn't leave anything half-baked, but I think it's in a reviewable state.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,101 +4,110 @@ name: VsVim CI
 
 on:
   push:
-    branches: [ "master", "main" ]
+    branches: [ "master", "main" ] # Branch builds with names starting with 'ma' will be published to this repo's releases feed
     tags:
       - '*'
-  pull_request:
-    branches: [ "master", "main", "dev/gha" ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 jobs:
-  windows-build:
-    name: Windows Build and Test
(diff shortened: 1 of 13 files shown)


vivlim:
This is currently failing to build, need to investigate further. :( https://github.com/vivlim/VsVim/issues/1

PR #3167, opened 2026-02-14
Fix Ctrl-W,V and :vs vertical split in VS 2026

### LINKED ISSUE #3161: Ctrl-W, V doesn't split window in VS 2026
I get this error when I try:

VsVim Error Error Start
Command "Window.NewWindow" is not valid.
VsVim Error Error End

Nick-Sohacki:
In VS 2026, the Window.NewWindow command was removed, causing Ctrl-W,V and :vs to fail with "Command 'Window.NewWindow' is not valid." This adds a fallback that uses the "Duplicate Tab" menu command (via CommandBar reflection) followed by
Window.NewVerticalTabGroup to achieve the same vertical split behavior.

Fixes #3161


Nick-Sohacki:
it looks like Visual Studio recently added a new command "Window.NewTabRight" that does exactly what we want. the PR right now does a more complicated work-around, so i'm gonna close it

Issue #3116, opened 2024-07-29
"Yank" hangs Visual Studio for 3 seconds before continuing with clipboard=unnamed


roel-v:
I'm sorry to bring up this years (decade by now?) old issue. I've seen https://github.com/VsVim/VsVim/issues/2067 and all the previous ones. When I use any yank command (with clipboard=unnamed set), say yy, the VS interface hangs for 3 seconds before continuing. The line is copied to the system clipboard correctly. I have read the wiki on this topic, I actually first encountered this problem many years ago, although I seem to remember it was only every second or third yank back then. I just retrained my muscle memory to use Ctrl-C/Ctrl-X which work fine, but I'm upgrading machines and now I was re-evaluating my .vsvimrc and thought I'd try again. If I enable the 'report clipboard errors' in the prefs and log to the output window, I get the same trace as in all the old bug reports. My question is - isn't there a way to map yy to the VS-native ctrl-c, and maybe d and D to ctrl-X?

Bonus question, feel free to ignore: despite reading all the old bug reports and discussion on them, I still don't understand what the real problem is. How can the clipboard get into a "faulty" state? What is 'faulty' about it? All other programs can call OpenClipboard() just fine, is there something about VsVim's way of using it?

(all using Windows 11, latest version of Visual Studio 2022, latest version of VsVim).

CretaceousConstructor:
Same problem has hunted me for over two years.

nevarrik:
I'm also looking into working around this, have either of you tried Diego Vieira's workaround from this issue? https://github.com/VsVim/VsVim/issues/2035

Inlining here for convenience:

> @divieira Wow thanks a lot ! I experienced a lot of sluggishness using the `x` key repeatedly in VS 2019 but I still wanted to use the system clipboard. I didn't know we could bypass the issue like that. As a matter of fact, I use `x` a lot (probably a bad vim habit but, you know...) and I don't ever want to put the deleted chars in any register so I just added a rule for putting those in the "black hole register" :
> 
> ```
> " Sometimes slow, see https://github.com/VsVim/VsVim/issues/2035
> "set clipboard=unnamed
> noremap y "*y
> noremap yy "*yy
> noremap Y "*y$
> noremap d "*d
> noremap dd "*dd
> noremap D "*D
> noremap p "*p
> noremap P "*P
> " black hole register for x because chars do not need to go in any accessible register anyway
> noremap x "_x
> ```
> 
> And now using x is crisp and fast ! Thanks again !
[...]

Issue #3040, opened 2023-02-27
System clipboard operations unusable slow after update to VS2022 17.5


NewUser2020:
**Describe the bug**
After updating to the most recent version of visual studio (17.5), every **second** clipboard operations take around 4 seconds.

**To Reproduce**
`set clipboard=unnamed`
Copy text to clipboard (for example with 'yy') -> The UI is frozen for about 4 seconds
Copy text to clipboard (for example with 'yy') -> No delay
Copy text to clipboard (for example with 'yy') -> The UI is frozen for about 4 seconds
Copy text to clipboard (for example with 'yy') -> No delay
...

Note that the delay does not happen if the system clipboard is not used (`set clipboard=`).

**Expected behavior**
Well, it should not freeze. It totally destroys the workflow...

**Environment (please complete the following information):**

- Visual Studio version: VS2022 17.5.0
- VsVim version: VsVim_2.10.0.6.vsix
- Programming Language: C#
- Check(Type 'x') any that are installed:
    - [ ] ReSharper
    - [ ] Visual Assist

jamesfAnet:
I'm getting the same thing

anctilma:
I'm getting the same thing in VS2019 16.11.27 with VsVim v2.8.0.0

BlueInt32:
This kind of clipboard issues has been bugging me for a long time and often have had to do with some other app keeping a handle on the system clipboard. Since I switched to Windows 11 OR the latest VS version, vsvim regularly reports it cannot access the system clipboard now. Even stopping all other applications on the system does not solve the problem, so the culprit could be Windows 11 or VS itself. 

Anyway, I think it's not worth the wait for someone to actually find a fix because it seems it's not even clear who is really responsible. 

FWIW and if you're like me, copying from VS to the system clipboard is not *that* common, so it may be relevant to have a specific mapping to get used to, to copy to the system clipboard only if visual mode is active: 

```
set clipboard= 
vnoremap <leader>y "*y
```

frozax:
I have the same issue as well. 
It might be a conflict with another app, but there is no issue with the VIM extension of VSCode for example.

frozax:
It's me again. I've spent some time debugging the issue, and I have a local patch that fixes it, at least for me. I've never wrote F# or WPF code before, so there might be better ways to do this. For those interested, here's how I fixed it (basically, using lower level code to access the Windows clipboard):

```
[...]

Issue #3038, opened 2023-02-02
System clipboard copy operation causes lag on a KVM guest.


exceedsystem:
**Describe the bug**
Copying(or cutting) text from VSVIM to the system clipboard on a Windows KVM guest causes a lag of several seconds.

**To Reproduce**

1. Add the following entry to the `_vimrc`.
```
set clipboard=unnamed,autoselect
```

2. Launch Visual Studio and open any document.
3. Select any text in visual mode and  press `y` or `x`.

**Expected behavior**
The copy to the clipboard should be completed immediately.

**Screenshots**
In line 1, 'demonstration' is copied using 'y'. In line 2, the entire line is selected and deleted using 'x'.

![GIF 2023-02-02 19-18-33](https://user-images.githubusercontent.com/70489172/216298164-042ec673-ff02-42aa-b79a-f057c81a54e2.gif)

**Environment (please complete the following information):**
- OS version:
Windows 10 Pro(21H2)
- Visual Studio version:
16.11.23
- VsVim version:
2.8.0.0
- Programming Language:
Any
- Check(Type 'x') any that are installed:
    - [ ] ReSharper
    - [ ] Visual Assist

**Additional context**
It is likely that this problem is related to the SPICE Agent service, which sharing the clipboard between the guest OS and host OS(when the SPICE Agent service is stopped, the problem does not occur). However, other programs do not have this problem.

Note:
The following is written about contributing.
https://github.com/VsVim/VsVim/blob/master/CONTRIBUTING.md

General Keyboard configuration problems are described below.
https://github.com/VsVim/VsVim/issues/2527

Best regards.

frozax:
See https://github.com/VsVim/VsVim/issues/3040 for a potential fix

Issue #2948, opened 2021-12-25
Visual Studio lags for ~4 seconds when yanking or deleting to clipboard (set clipboard = unnamed)


se-omarayman:
**Describe the bug**
I am a windows user, and I wanted  vsvim to copy code to the clipboard to make pasting outside visual studio easier,
when I was on windows 10 vsvim worked perfectly, but since upgrading to windows 11, whenever I yank or delete to the system clipboard the UI freezes for about 4 seconds.

**To Reproduce**
1- set clipboard = unnamed
2- try to yank or delete with yy, dd, x, etc..
3- UI freezes for about 4 seconds.

**Environment (please complete the following information):**

- Visual Studio version: 17.0.4 (2022)
- VsVim version: 2.10.0.5
- Programming Language: C#
- Windows 11
- Check(Type 'x') any that are installed:
    - [ ] ReSharper
    - [ ] Visual Assist

zbjxb:
I had the same problem

frozax:
I have this problem as well, but it's new. I've disabled all the other extensions and it's still happening. Anyone has a workaround?
I'm in C++, Windows 11, latest Visual Studio Community 2022 and latest VsVim.

frozax:
Might be similar to https://github.com/VsVim/VsVim/issues/3040, where I posted a potential fix.

PR #3170, opened 2026-05-09
Changing clipboard implementation to prevent freeze when attempting to yank

knetworx:
I'm a game dev who works primarily in C++. C# isn't my forte, I haven't touched it since college (about 2 decades ago), and I've never made an extension for VS before. But I've been using VsVim for years, and I've tried all kinds of workarounds for yank either freezing, failing, or not using the unnamed clipboard, and it's been the bane of my existence forever.

So I forked the project, and spent some time with Claude working through it, and have what appears to be a solution. I managed to run the extension in a VS 2022 build (17.14.30), and the yank worked as expected, with no delay/freezing, and I was able to paste the text across different apps. Feel free to tear apart the PR, but please take a look, test it out in your own environment, and see if it can help, because I've now got a release vsix running on my local install, and I feel so much better about life now.

Related tickets:
[Visual Studio lags for ~4 seconds when yanking or deleting to clipboard (set clipboard = unnamed)](https://github.com/VsVim/VsVim/issues/2948)
["Yank" hangs Visual Studio for 3 seconds before continuing with clipboard=unnamed](https://github.com/VsVim/VsVim/issues/3116)
[System clipboard operations unusable slow after update to VS2022 17.5](https://github.com/VsVim/VsVim/issues/3040)
[System clipboard copy operation causes lag on a KVM guest.](https://github.com/VsVim/VsVim/issues/3038)

### DIFF
--- a/Src/VimWpf/Implementation/Misc/ClipboardDevice.cs
+++ b/Src/VimWpf/Implementation/Misc/ClipboardDevice.cs
@@ -1,73 +1,171 @@
-﻿using System;
+using System;
 using System.ComponentModel.Composition;
+using System.Runtime.InteropServices;
 using System.Threading;
-using System.Windows;
 
 namespace Vim.UI.Wpf.Implementation.Misc
 {
     [Export(typeof(IClipboardDevice))]
     internal sealed class ClipboardDevice : IClipboardDevice
     {
-        private readonly IProtectedOperations _protectedOperations;
+        private static class NativeMethods
+        {
+            public const uint CF_UNICODETEXT = 13;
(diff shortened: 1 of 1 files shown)


knetworx:
It's also working for me over Parsec, as mentioned by this comment: https://github.com/VsVim/VsVim/issues/3040#issuecomment-1755020197
Note that @frozax made a follow-up comment with a patch that seems to similarly fix the issue, but it seems that never got merged: https://github.com/VsVim/VsVim/issues/3040#issuecomment-1756118012

knetworx:
Need to resolve issues with other keybindings that put text on the clipboard.

PR #3175, opened 2026-07-29
VS 2026: Fix vertical split, and experimental instance deployment. Stop building for 2019

### LINKED ISSUE #3161: Ctrl-W, V doesn't split window in VS 2026

vivlim:
Fixes #3161.

On VS 2026, uses `Windows.NewTabRight`. On prior versions, that will throw the first time and we'll fall back to `Window.NewWindow` + `Window.NewVerticalTabGroup`. I went with this instead of a version check because I don't know exactly which version that change happened in.

I also got the VS "experimental instance" dev loop working on VS 2026 again, though that probably can't be built on VS 2022 anymore due to some substantial changes in the build tools. VS 2026 is able to deploy *to* VS 2022 though (that's how I tested the vertical split change on 2022)

The newer VSSDK build tools aren't able to handle 2019's assemblies (error below). I'd like to just remove those rather than continue building for 2019 out of `master`. In a future PR I can clean up more of the mess I made back in #2917 (adding support for 2022, which required bifurcating a bunch of stuff)

```
error : CreatePkgDef : error : BadImageFormatException: Could not load file or assembly 'file:///D:\a\VsVim\VsVim\Binaries\Release\VsVim2019\net472\VsVim.dll' or one of its dependencies. An attempt was made to load a program with an incorrect format. [D:\a\VsVim\VsVim\Src\VsVim2019\VsVim2019.csproj]
```


VsVim.sln
```diff
@@ -284,6 +284,7 @@ Global
 		{E6552DEF-32C8-407C-9940-52A2CC54E521}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{E6552DEF-32C8-407C-9940-52A2CC54E521}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
 		{E6552DEF-32C8-407C-9940-52A2CC54E521}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{E6552DEF-32C8-407C-9940-52A2CC54E521}.Debug|Mixed Platforms.Deploy.0 = Debug|Any CPU
```
vivlim, line 209: VSSDK build tools no longer install during build, that happens during deploy now, so this is necessary

Src/VsVim2022/VsVim2022.csproj
```diff
@@ -13,8 +13,9 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <VsVimProjectType>Vsix</VsVimProjectType>
     <VsVimVisualStudioTargetVersion>17.0</VsVimVisualStudioTargetVersion>
+    <UseWPF>true</UseWPF> <!-- needed with newer VSSDK build tools, avoids trying to pack during WPF build -->
```
vivlim, line 16: It's a good thing I've already encountered this and knew to try it 🫠

Src/VsVimShared/LatchingFallback.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+
+namespace Vim.VisualStudio;
+
+/// <summary>
+/// Stateful container for running one of two actions depending on whether the first one throws or not.
+/// </summary>
+internal class LatchingFallback(Action action, Action fallbackAction)
```
vivlim, line 8: Not strictly necessary but I wanted to keep this stuff contained along with its state

